PR #1, opened 2017-06-19
builder

kashike:


src/main/java/net/kyori/text/BaseComponent.java
```diff
@@ -46,61 +46,62 @@
...
    * The string to insert when this component is shift-clicked in chat.
    */
-  @Nullable private String insertion;
+  @Nullable protected final String insertion;
 
-  @Nonnull
-  @Override
-  public List<Component> children() {
-    return Collections.unmodifiableList(this.children);
+  protected BaseComponent(@Nonnull final List<Component> children, @Nullable final TextColor color, @Nonnull final TextDecoration.State obfuscated, @Nonnull final TextDecoration.State bold, @Nonnull final TextDecoration.State strikethrough, @Nonnull final TextDecoration.State underlined, @Nonnull final TextDecoration.State italic, @Nullable final ClickEvent clickEvent, @Nullable final HoverEvent hoverEvent, @Nullable final String insertion) {
```
SupaHam: A constructor accepting just a Builder instance might come in handy.

src/main/java/net/kyori/text/BaseComponent.java
```diff
@@ -242,4 +201,107 @@ public String toString() {
...
+     */
+    @Nonnull protected List<Component> children = EMPTY_COMPONENT_LIST;
+    /**
+     * The color of this component.
+     */
+    @Nullable protected TextColor color;
+    /**
+     * If this component should have the {@link TextDecoration#OBFUSCATED obfuscated} decoration.
+     */
+    @Nonnull protected TextDecoration.State obfuscated = TextDecoration.State.NOT_SET;
```
SupaHam: An `EnumMap<TextDecoration, TextDecoration.State>` might be more convenient for storing styles in case more get added in the future.

src/main/java/net/kyori/text/KeybindComponent.java
```diff
@@ -51,15 +80,104 @@ public String keybind() {
...
+
+  @Nonnull
+  @Override
+  public KeybindComponent color(@Nullable TextColor color) {
+    return new KeybindComponent(this.children, color, this.obfuscated, this.bold, this.strikethrough, this.underlined, this.italic, this.clickEvent, this.hoverEvent, this.insertion, this.keybind);
+  }
+
+  @Nonnull
+  @Override
+  public KeybindComponent decoration(@Nonnull TextDecoration decoration, @Nonnull TextDecoration.State state) {
```
SupaHam: This method would be simplified by both a Builder-arg constructor and usage of EnumMap.

_All in the name of scalability!_

src/main/java/net/kyori/text/KeybindComponent.java
```diff
@@ -51,15 +80,104 @@ public String keybind() {
...
+
+  @Nonnull
+  @Override
+  public KeybindComponent resetStyle() {
+    return new KeybindComponent(this.children, null, TextDecoration.State.NOT_SET, TextDecoration.State.NOT_SET, TextDecoration.State.NOT_SET, TextDecoration.State.NOT_SET, TextDecoration.State.NOT_SET, null, null, null, this.keybind);
+  }
+
+  @Nonnull
+  @Override
+  public Component copy() {
```
SupaHam: Not sure if it might be worth Specifying the current `Component` subclass as a return type of `copy`. I don't believe there is much to gain currently, but it could make a difference in the future.

SupaHam:
Great work, my friend!

This PR looks like it'll make this library very capable of all of the Minecraft text features with the addition of immutability. 👍

kashike:
x

PR #4, opened 2017-08-15
Add compatibility with older Guava versions

lucko:
The public plugins I use text in have to be compatible with guava 17.0, since that's what was used in some of the older versions of MC. (sadly, a significant majority still don't run 1.12 😢 )

I've been using a forked version (https://github.com/lucko/text/tree/luck) until now, but would kinda like to stop doing that. 

Not sure if you wanted this in upstream, but I would guess it would be useful for other people looking to use the lib.

### DIFF
--- a/src/main/java/net/kyori/text/AbstractComponent.java
+++ b/src/main/java/net/kyori/text/AbstractComponent.java
@@ -23,7 +23,6 @@
  */
 package net.kyori.text;
 
-import com.google.common.base.MoreObjects;
 import com.google.common.collect.ImmutableList;
 import net.kyori.text.event.ClickEvent;
 import net.kyori.text.event.HoverEvent;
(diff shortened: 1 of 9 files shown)


kashike:
This branch targets Minecraft 1.12, which requires Guava 21.0. I have no plans to pull this currently, but I can look into resetting the 1.11 branch, which uses Guava 17.0, off of master (excluding 1.12 features, of course) after #5 is pulled.

Thoughts?

lucko:
Yeah, up to you.

The only thing you really lose is the (supposed?) convenience of the ToString utility -- which in my opinion doesn't really account for much. Personally, I would prefer that over maintaining multiple branches. That's just me tho. :wink:

I don't really know what your aims are with this, but for my own use cases, compatibility with older MC versions is essential. (Bukkit/BungeeCord plugins)

kashike:
https://github.com/KyoriPowered/text/commit/958843d52e0850415150347223ad5d0d98d500fd

PR #5, opened 2017-08-15
Allow the legacy character to be specified in LegacyComponent

lucko:
Allows the use of both `§` and `&`.

### DIFF
--- a/src/main/java/net/kyori/text/LegacyComponent.java
+++ b/src/main/java/net/kyori/text/LegacyComponent.java
@@ -52,7 +52,13 @@ public final class LegacyComponent {
   @Deprecated
   @Nonnull
   public static TextComponent from(@Nonnull final String input) {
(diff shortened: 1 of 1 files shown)


src/main/java/net/kyori/text/LegacyComponent.java
```diff
@@ -52,7 +52,13 @@ private LegacyComponent() {
   @Deprecated
   @Nonnull
   public static TextComponent from(@Nonnull final String input) {
-    int next = input.lastIndexOf(CHARACTER, input.length() - 2);
+    return from(input, CHARACTER);
+  }
+
+  @Deprecated
+  @Nonnull
+  public static TextComponent from(@Nonnull final String input, final char legacyChar) {
```
kashike: Name the second parameter `character`.

lucko: Changed.

kashike:
Thanks, @lucko.

PR #7, opened 2017-08-16
Fix deserialization of legacy text

lucko:
Fixes all components inheriting color/style from the first component in the chain.

For example, currently:
```
&3> &rpms &7- &rfalse
```

is parsed to

### DIFF
--- a/src/main/java/net/kyori/text/LegacyComponent.java
+++ b/src/main/java/net/kyori/text/LegacyComponent.java
@@ -63,7 +63,7 @@ public final class LegacyComponent {
       return TextComponent.of(input);
     }
 
-    final List<TextComponent.Builder> parts = new ArrayList<>();
+    final List<TextComponent> parts = new ArrayList<>();
 
     TextComponent.Builder current = null;
     boolean reset = false;
@@ -76,7 +76,7 @@ public final class LegacyComponent {
         if(from != pos) {
           if(current != null) {
             if(reset) {
-              parts.add(current);
+              parts.add(current.build());
               reset = false;
               current = TextComponent.builder("");
             } else {
@@ -99,21 +99,11 @@ public final class LegacyComponent {
     } while(next != -1);
 
     if(current != null) {
-      parts.add(current);
+      parts.add(current.build());
     }
 
     Collections.reverse(parts);
-    switch(parts.size()) {
-      case 0:
-        return TextComponent.of(pos > 0 ? input.substring(0, pos) : "");
-      case 1:
-        return parts.get(0).build();
-      case 2:
-      default:
-        return parts.get(0)
-          .append(parts.subList(1, parts.size()).stream().map(TextComponent.Builder::build).collect(Collectors.toList()))
-          .build();
-    }
+    return TextComponent.builder(pos > 0 ? input.substring(0, pos) : "").append(parts).build();
   }
 
   private static boolean applyFormat(@Nonnull final TextComponent.Builder builder, @Nonnull final TextFormat format) {
--- a/src/test/java/net/kyori/text/LegacyComponentTest.java
+++ b/src/test/java/net/kyori/text/LegacyComponentTest.java
@@ -24,6 +24,8 @@
 package net.kyori.text;
 
 import net.kyori.text.format.TextColor;
+import net.kyori.text.format.TextDecoration;
+
 import org.junit.Test;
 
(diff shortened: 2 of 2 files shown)


Minecrell:
That switch block tries to optimize things where optimization isn't necessary, because there is only a a single case where we can leave out the extra root component: that is when `pos == 0 && parts == 1`.

In all other cases the additional root component is required either because it holds the unformatted text at the beginning or to prevent wrong inheritance (like you've noticed).

The old code completely ignores the unformatted text at the beginning. E.g. let's say we have `Hello &ckashike`. The current code will parse this to `kashike` (red) and ignore the `Hello ` entirely.

You correctly fix this for `parts >= 2` by adding back `pos > 0 ? input.substring(0, pos) : ""` but the same needs to be done for `parts == 1`. At this point you might as well remove the switch block because all branches do the same. It can be just written as:

```java
return TextComponent.builder(pos > 0 ? input.substring(0, pos) : "").append(parts).build();
```

src/main/java/net/kyori/text/LegacyComponent.java
```diff
@@ -103,17 +103,9 @@ public static TextComponent from(@Nonnull final String input, final char charact
...
-      case 1:
-        return parts.get(0).build();
-      case 2:
-      default:
-        return parts.get(0)
-          .append(parts.subList(1, parts.size()).stream().map(TextComponent.Builder::build).collect(Collectors.toList()))
-          .build();
-    }
+    return TextComponent.builder(pos > 0 ? input.substring(0, pos) : "")
+      .append(parts.stream().map(TextComponent.Builder::build).collect(Collectors.toList()))
```
Minecrell: You can avoid creating a new list here if you build the components before adding them to `parts`.

lucko:
Thanks @Minecrell. I've removed the switch block and replaced it with the default for 2+ parts.

src/main/java/net/kyori/text/LegacyComponent.java
lucko: Ah yeah. Thanks

src/test/java/net/kyori/text/LegacyComponentTest.java
```diff
@@ -38,6 +38,6 @@ public void testFrom() {
   @Test
   public void testFromColor() {
     assertEquals(TextComponent.of("foo").color(TextColor.GREEN), LegacyComponent.from(LegacyComponent.CHARACTER + "afoo"));
```
Minecrell: To be fair, this test relies exactly on the one case where you could make the optimization, so you either need to add it back for `pos == 0 && parts == 1` or modify what's expected.

lucko: I just removed the extra test.

lucko:
(Kinda) related to this, but with the following component, what would be the correct legacy version?

```java
TextComponent component = TextComponent.builder("hi").decoration(TextDecoration.BOLD, TextDecoration.State.TRUE)
  .append(TextComponent.of("foo").color(TextColor.GREEN).decoration(TextDecoration.BOLD, TextDecoration.State.FALSE))
  .append(TextComponent.of("bar").color(TextColor.BLUE))
  .append(TextComponent.of("baz"))
  .build();
```

Currently it spits out: `&lhi&afoo&9barbaz`
But should it not be: `&lhi&afoo&9&lbar&lbaz` (where the child components inherit style & color from their parent?)

Minecrell:
You're right, the `to` code is broken:
```
[Aug 13 10:32:06] <minecrell> Anyway, you need to fix the "to" code :)
[Aug 13 10:34:18] <kashike> meh, I'll fix it later
```

These two ways would be correct:
- `&lhi&afoo&9&lbar&r&lbaz` (The additionally `&r` is necessary before `baz` because text style (bold etc) does not reset the color)
- `&lhi&r&afoo&r&9&lbar&r&lbaz&r` (That's Mojang's implementation, it resets after each component and applies the formatting again)

The first variant is of course much nicer because it avoids unnecessary codes but it is not that easy to implement: You need to track the current style and reset as-needed when iterating over the texts. [The Sponge implementation](https://github.com/SpongePowered/SpongeCommon/blob/bleeding/src/main/java/org/spongepowered/common/mixin/core/text/MixinTextComponentBase.java#L105-L149) is kind of messy, there are probably better ways...

kashike:
Thanks, @lucko.

PR #8, opened 2017-08-16
Introduce component serializers

kashike:


src/main/java/net/kyori/text/serializer/ComponentSerializers.java
```diff
@@ -0,0 +1,42 @@
...
+/**
+ * A collection of component serializers.
+ */
+public final class ComponentSerializers {
+
+  public static final ComponentSerializer<Component, String> GSON = GsonComponentSerializer.INSTANCE;
+  /**
+   * @deprecated legacy
+   */
+  @Deprecated
```
jamierocks: no need for this to be deprecated

sure the spec underlying the code is deprecated, but this is just going to make peoples ides moan at them, although perhaps it should if they use legac.....

carry on

src/main/java/net/kyori/text/serializer/LegacyComponentSerializer.java
```diff
@@ -0,0 +1,100 @@
...
+import javax.annotation.Nonnull;
+
+/**
+ * A legacy component serializer.
+ *
+ * @param <C> the component type
+ * @deprecated legacy
+ */
+@Deprecated
+public interface LegacyComponentSerializer<C extends Component> extends ComponentSerializer<C, String> {
```
jamierocks: probably just dperecating the class will do

src/main/java/net/kyori/text/serializer/ComponentSerializers.java
```diff
@@ -0,0 +1,42 @@
...
+package net.kyori.text.serializer;
+
+import net.kyori.text.Component;
+
+/**
+ * A collection of component serializers.
+ */
+public final class ComponentSerializers {
+
+  public static final ComponentSerializer<Component, String> GSON = GsonComponentSerializer.INSTANCE;
```
kashike: `GSON` or `JSON`?

connorhartley: Difficult question... I would say... keep it how it is.

lucko: I would take the approach Sponge has. 

LEGACY_CHAR and JSON. 

Users of text shouldn't need to worry about how the method is implemented.

jamierocks: JSON

mikroskeem: Yes JSON would make more sense

src/main/java/net/kyori/text/serializer/ComponentSerializers.java
lucko: Idk, isn't just being called "LEGACY" enough? 

Deprecation warnings are annoying :(

src/main/java/net/kyori/text/serializer/GsonComponentSerializer.java
```diff
@@ -0,0 +1,239 @@
...
+import net.kyori.text.format.TextDecoration;
+
+import java.lang.reflect.Type;
+import java.util.ArrayList;
+import java.util.List;
+
+import javax.annotation.Nonnull;
+import javax.annotation.Nullable;
+
+public enum GsonComponentSerializer implements ComponentSerializer<Component, String>, JsonDeserializer<Component>, JsonSerializer<Component> {
```
lucko: Any reason why this is public?

Shouldn't the instance be obtained from the ComponentSerializers class?

kashike: To allow registration on custom `Gson` instances.

PR #34, opened 2019-05-13
Add support for detecting and linking to URLs in the legacy serializer

### LINKED ISSUE #33: Legacy serializer breaks MC detection of links in chat
When serializing a message from legacy that includes a link and sending it to a player. The link is not clickable.

The same link sent in base MC chat is clickable, but legacy seems to break this function

astei:
Fixes #33


astei:
This might be moved into `text-extras` or will use a different serializer field most likely. I prefer the latter especially because Velocity users expect an API close to BungeeCord's and we absolutely need this behavior somewhere.

serializer-legacy/src/main/java/net/kyori/text/serializer/legacy/LegacyComponentSerializerImpl.java
```diff
@@ -25,19 +25,19 @@
...
-import java.util.Collections;
-import java.util.EnumSet;
-import java.util.List;
-import java.util.Set;
+import java.util.*;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
 
 class LegacyComponentSerializerImpl implements LegacyComponentSerializer {
+  private static final Pattern URL_PATTERN = Pattern.compile("(?:(https?)://)?([-\\w_.]{2,}\\.[a-z]{2,4})(/\\S*)?");
```
hugmanrique: One char subdomains are pretty common (especially for image servers, e.g. `i.imgur.com`. I believe the current regex only allows for 2+ char subdomains.

Same thing for TLDs. gTLDs longer than 4 chars are becoming pretty common (.shopping, .tech, etc.).

Maybe not supporting punnycode domains is a good idea (I don't even know how the MC font renders these chars).

This PR looks good to me except the regex (which can be changed). URL parsing is a hard problem 😛.

hugmanrique: Extreme minor nit: mind changing the a-z to \w (the same way the subdomains are matched).

This regex also doesn't match on numeric domains (I believe the spec allows the full ASCII charset except control chars maybe, don't quote me on it)

mikroskeem: [This handles really short urls fine](https://github.com/mikroskeem/Adminchatter/blob/master/common/src/main/kotlin/eu/mikroskeem/adminchatter/common/utils/Chat.kt#L61)

lucko: Aaaand another pattern to throw into the mix ;)

https://hub.spigotmc.org/stash/projects/SPIGOT/repos/craftbukkit/browse/src/main/java/org/bukkit/craftbukkit/util/CraftChatMessage.java#21

kashike:
Thanks, @astei.

Issue #39, opened 2019-07-23
Footer is always longer than header with Pagination


501warhead:
This may be a mistake on my end, attempting to integrate this API with my recent works and having some difficulty with pagination. No matter what I attempt to do, the footer is always slightly longer than the header. I've attempted at this point any number of fixes and tried digging for some potential solutions.

I do know this is still a very new feature, so I figured I'd at least provide a bug report for my findings to see if it helps

Version: `3.0.2`

Using `text-serializer-gson`,`text-feature-pagination`, and `text-api` with `text-bukkit-adapter`

Platform is Paper

Width: 53
![image](https://user-images.githubusercontent.com/11152237/61743550-d2360a00-ad95-11e9-857e-5679877e219a.png)

Width: 50
![image](https://user-images.githubusercontent.com/11152237/61743709-217c3a80-ad96-11e9-96d2-c76d44a1d5c0.png)

With longer title, width 50
![image](https://user-images.githubusercontent.com/11152237/61743992-8e8fd000-ad96-11e9-9516-0deab3a9721d.png)

Width 50 with dummy data to cause pagination
![image](https://user-images.githubusercontent.com/11152237/61744659-293cde80-ad98-11e9-873b-e89252f969f0.png)

smol title
![image](https://user-images.githubusercontent.com/11152237/61746801-c863d500-ad9c-11e9-83e5-3d036ea04e7c.png)

Code (In Kotlin) ((An abomination of formatting, I apologize in advance))
```Kotlin
return Pagination.builder().line { cs ->
            cs.character('=')
        }.width(53).build(TextComponent.of("Charters").color(TextColor.GOLD), Pagination.Renderer.RowRenderer { value, index ->
            return@RowRenderer listOf(if (value == null) TextComponent.of("${index + 1}. ").color(TextColor.YELLOW).content("ERR?").color(TextColor.RED) else
                TextComponent.of("${index + 1}. ")
                        .color(TextColor.YELLOW)
                        .append(TextComponent.of(value.name).color(TextColor.GRAY))
                        .hoverEvent(HoverEvent.of(HoverEvent.Action.SHOW_TEXT, TextComponent.of("Click to view this charter").color(TextColor.GRAY)))
                        .clickEvent(ClickEvent.of(ClickEvent.Action.RUN_COMMAND, "/charter info ${value.name}")))
        }, Pagination.PageCommandFunction { page -> "/charter list -page $page" })
```

```Kotlin
[...]

PR #40, opened 2019-07-27
Custom line rendering support

501warhead:
A follow up from #39 for Pagination

I've run into an issue that the maintainers here likely have already grappled with from @lucko 's input, where the variable width Minecraft font characters leave the header and footers unaligned. This is likely a headache to solve, but the thing that I felt was that at the very least there should be more customization as to how the lines are rendered to give customization to those who implement to handle it individually.

This PR migrates line rendering logic into a new method on `Pagination.Renderer`. It's a very basic implementation because I feel that this is something that probably will have some discussion around it and I'd rather not go overboard without seeing what the maintainers here would prefer.

There aren't any contribution guidelines that I've seen so this is mostly just a shot in the dark based on the current code base.

Side note, the `repeat` method is kind of an odd floating method, uncertain really where to put it after this migration.

### DIFF
--- a/feature-pagination/src/main/java/net/kyori/text/feature/pagination/Pagination.java
+++ b/feature-pagination/src/main/java/net/kyori/text/feature/pagination/Pagination.java
@@ -23,6 +23,10 @@
  */
 package net.kyori.text.feature.pagination;
 
+import java.util.Collection;
+import java.util.Collections;
+import java.util.List;
+import java.util.function.Consumer;
 import net.kyori.text.Component;
 import net.kyori.text.TextComponent;
 import net.kyori.text.event.ClickEvent;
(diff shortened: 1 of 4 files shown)


feature-pagination/src/main/java/net/kyori/text/feature/pagination/Pagination.java
```diff
@@ -202,6 +203,14 @@ public String toString() {
         .build();
     }
 
+    default @NonNull Component renderLine(final char character, final @NonNull Style style, final int length) {
+      return TextComponent.of(repeat(String.valueOf(character), length), style);
+    }
+
+    static @NonNull String repeat(final @NonNull String character, final int count) {
+      return String.join("", Collections.nCopies(count, character));
```
liach: I'd prefer a for loop and stringbuilder instead of a short line of code that adds extra overhead

501warhead: Should be resolved now with latest commit

kashike:
Hey @501warhead, sorry for not getting to this pull request sooner. We have refactored a large amount of the codebase, and so I will be closing this pull request. Please feel free to update your branch and I will re-open.

PR #49, opened 2019-09-30
Added storage nbt component

liach:
Look at 19w39a

Signed-off-by: liach <liach@users.noreply.github.com>


liach:
Updated to target 1.15.

PR #58, opened 2020-03-29
Add support for different style merge strategies

kashike:


kashike:
> besides `IF_ABSENT_ON_TARGET` strategy, how about rename `ALL` to `IF_PRESENT_ON_SOURCE` and add a separate `OVERWRITE` (i.e. force setting to null even when insertion etc is in an undefined state)

Well, we currently exclude anything from the `source` that is `null` completely... that would be a change to existing functionality.

PR #62, opened 2020-04-17
Let's go on an adventure.

kashike:


api/src/main/java/net/kyori/adventure/audience/Audience.java
```diff
@@ -21,23 +21,27 @@
...
+/**
+ * An audience.
+ */
+public interface Audience {
+  /**
+   * Sends a message.
+   *
+   * @param message the message
+   */
+  void message(final @NonNull Component message);
```
Electroid, line 39: `sendMessage` and `playSound`? 

I know the full lowercase thing is a style, but it just sounds weird rn. It would also line up nicely with platform APIs.

kashike: `sound` changed to `playSound`, but I think `message` alone is fine.

api/src/main/java/net/kyori/adventure/bossbar/BossBar.java
```diff
@@ -0,0 +1,170 @@
...
+   * Gets the percent.
+   *
+   * @return the percent
+   */
+  float percent();
+
+  /**
+   * Sets the percent.
+   *
+   * @param percent the percent
```
Electroid, line 59: Should the api call out this is from [0, 1]. Notchian clients render > 1 weirdly.

api/src/main/java/net/kyori/adventure/bossbar/AbstractBossBar.java
```diff
@@ -0,0 +1,173 @@
...
+  }
+
+  @Override
+  public float percent() {
+    return this.percent;
+  }
+
+  @Override
+  public @NonNull BossBar percent(final float percent) {
+    if(percent != this.percent) {
```
Electroid, line 71: An optimization hint here: often boss bars are used for long countdowns (say 30 minutes or 2 hours). If `percent` changes from `0.0100` to `0.01001` there is no *effective* difference any client could ever see. Yet, an updated packet would be sent anyway.

To reduce packet verbosity: I would recommend defining an `static final int PERCENT_EPSILON = 0.01`, if the change in percent is less than that, the class updates the field, but `changed` is not called.

I would put this in the abstract implementation, because I believe this would benefit and be useful by all implementations.

api/src/main/java/net/kyori/adventure/sound/SoundStopImpl.java
```diff
@@ -0,0 +1,48 @@
...
+ */
+package net.kyori.adventure.sound;
+
+import net.kyori.adventure.key.Key;
+import org.checkerframework.checker.nullness.qual.Nullable;
+
+final class SoundStopImpl implements SoundStop {
+  static SoundStop ALL = new SoundStopImpl(null, null);
+  private final @Nullable Key sound;
+  private final Sound.@Nullable Source source;
```
liach, line 32: Can just do 
```suggestion
  private final @Nullable Sound.Source source;
```
because `Sound$Source` is a static inner class (enum), and the type path is always empty (than going into inner class)

api/src/main/java/net/kyori/adventure/sound/Sound.java
```diff
@@ -0,0 +1,97 @@
...
+   * @return the name
+   */
+  @NonNull Key name();
+
+  /**
+   * Gets the source.
+   *
+   * @return the source
+   */
+  @NonNull Source source();
```
liach, line 59: is this category or source?

api/src/main/java/net/kyori/adventure/bossbar/BossBar.java
```diff
@@ -0,0 +1,176 @@
...
+  @NonNull BossBar name(final @NonNull Component name);
+
+  /**
+   * Gets the percent.
+   *
+   * <p>The percent is a value between 0 and 1.</p>
+   *
+   * @return the percent
+   */
+  float percent();
```
liach, line 57: Is there a type annotation indicating the range of a float? if there is consider using one for this value and the setter below.

api/src/main/java/net/kyori/adventure/sound/SoundStopImpl.java
kashike: Not valid.

api/src/main/java/net/kyori/adventure/bossbar/BossBar.java
kashike: Sadly not.

api/src/main/java/net/kyori/adventure/audience/Audience.java
```diff
@@ -0,0 +1,79 @@
...
+import net.kyori.adventure.bossbar.BossBar;
+import net.kyori.adventure.sound.Sound;
+import net.kyori.adventure.sound.SoundStop;
+import net.kyori.adventure.text.Component;
+import org.checkerframework.checker.nullness.qual.NonNull;
+
+/**
+ * An audience.
+ */
+public interface Audience {
```
lucko, line 35: Few questions:

What is `Audience`?

* A single "player"? 
* A single "command sender" type object - something that can receive messages, sounds, etc?
* A group of either of the above?

I'm assuming it's no 3? 

If so, what happens when you try to play a sound to the console for example? No-op, exception, ..., ?

None of this is really very clear from the interface spec - I wonder whether a different name might be more suitable, depending on the answers to the above questions.

kashike: `Audience` is number three.

If you try to do something unsupported, it will be swallowed.

kashike:
> In general, the move from just `text` to `text` plus related things seems like a good move - certainly titles, bossbar, etc.

:tada: 
> I think that `Sound` might be a step too far for inclusion in the main library - and could potentially be better suited in an add-on, kinda like how we have the serializers separated out of the main project.

The only issue there is that `Audience` would not contain these methods, which is the goal.
> Not entirely sold on the introduction of `ExaminableProperty` - but I don't feel super strongly about it either way.
>
> I just wonder whether the same data could be exposed using `Map<String, Object>` in the same, if not nicer way - without the need to pollute the library with another dependency.

Definitely not in a nicer way.

api/src/main/java/net/kyori/adventure/bossbar/AbstractBossBar.java
```diff
@@ -0,0 +1,178 @@
...
+      if(enoughForClientToNotice) {
+        this.changed(Change.PERCENT);
+      }
+    }
+    return this;
+  }
+
+  // https://github.com/KyoriPowered/text/pull/62#discussion_r410790072
+  private static boolean enoughForClientToNotice(final float oldValue, final float newValue) {
+    return (newValue - oldValue) >= MINIMUM_PERCENT_CHANGE;
```
Electroid, line 84:
```suggestion
    return Math.abs(newValue - oldValue) >= MINIMUM_PERCENT_CHANGE;
```

api/src/main/java/net/kyori/adventure/text/event/HoverEvent.java
```diff
@@ -85,13 +85,13 @@
...
-  private final Action action;
+  private final Action<?> action;
   /**
    * The hover event value.
    */
-  private final Component value;
+  private final Object value;
 
-  private HoverEvent(final @NonNull Action action, final @NonNull Component value) {
+  private HoverEvent(final @NonNull Action<?> action, final @NonNull Object value) {
```
liach, line 98: can move the generic on hover event itself imo, no need to abide to mojang code

api/src/main/java/net/kyori/adventure/bossbar/BossBar.java
```diff
@@ -0,0 +1,199 @@
...
+   * @param flags the flags
+   * @return the bossbar
+   */
+  @NonNull BossBar removeFlags(final @NonNull Flag@NonNull... flags);
+
+  @NonNull BossBar addListener(final @NonNull Listener listener);
+
+  @NonNull BossBar removeListener(final @NonNull Listener listener);
+
+  interface Listener {
```
Electroid, line 138: `@FunctionalInterface` :)

kashike: https://github.com/KyoriPowered/text/commit/6cbf1f02e9acb07f037e86e074def80d77525f32

api/src/main/java/net/kyori/adventure/audience/Audience.java
Electroid: I would still strongly consider having a `sendMessage` to allow for easier drop-in replacement. Maybe `message` is an alias?

api/src/main/java/net/kyori/adventure/audience/Audience.java
```diff
@@ -0,0 +1,95 @@
...
+
+  /**
+   * Shows a title.
+   *
+   * @param title the title
+   */
+  void showTitle(final @NonNull Title title);
+
+  /**
+   * Clears the title.
```
Electroid, line 87: Would be nice to have a brief blurb on the difference between clear and reset. TBH, I don't even know the diff.

api/src/main/java/net/kyori/adventure/audience/Audience.java
```diff
@@ -0,0 +1,95 @@
...
+   * @param bar the bossbar
+   */
+  void hideBossBar(final @NonNull BossBar bar);
+
+  /**
+   * Shows a message on the action bar.
+   *
+   * @param message the message
+   */
+  void showActionBar(final @NonNull Component message);
```
Electroid, line 63: There seems to be a pattern emerging: `send` for ephemeral stuff that can't be "taken back" and `show`/`hide` for things that will persist unless hidden later. Maybe `sendActionBar` would be better here?

api/src/main/java/net/kyori/adventure/audience/ForwardingAudience.java
```diff
@@ -0,0 +1,65 @@
...
+/**
+ * An audience that forwards all methods to another audience.
+ */
+public interface ForwardingAudience extends Audience {
+  /**
+   * Gets the forwarded audience.
+   *
+   * @return the forwarded audience
+   */
+  @NonNull Audience audience();
```
Electroid, line 19: It might be useful for this to be `@Nullable`, then silently fail if `null`. Or maybe define an `Audience.empty()` to do the same thing?

api/src/main/java/net/kyori/adventure/bossbar/BossBar.java
```diff
@@ -0,0 +1,200 @@
...
+
+  /**
+   * Sets the flags.
+   *
+   * @param flags the flags
+   * @return the bossbar
+   */
+  @NonNull BossBar removeFlags(final @NonNull Flag@NonNull... flags);
+
+  @NonNull BossBar addListener(final @NonNull Listener listener);
```
Electroid, line 134: Now that I think about this again, is a `Listener` interface needed? You could define a `BiConsumer<BossBar, Change>`?

api/src/main/java/net/kyori/adventure/sound/SoundImpl.java
```diff
@@ -21,48 +21,54 @@
...
-
-  /**
-   * The name map.
-   */
-  public static final NameMap<TextColor> NAMES = NameMap.create(TextColor.class, constant -> constant.name);
-  /**
-   * The name of this color.
-   */
-  private final String name;
+final class SoundImpl implements Examinable, Sound {
```
Electroid, line 32: Should these `Impl` classes have a `hashCode` and `equals` implementation? Might make sense, for instance, to put these in a `Map`.

api/src/main/java/net/kyori/adventure/bossbar/BossBar.java
kashike: I want the method to be named `bossBarChanged` so that it makes more sense if someone decides to implement `Listener` in their own class.

api/src/main/java/net/kyori/adventure/audience/Audience.java
kashike: https://github.com/KyoriPowered/text/pull/62/commits/fd7ce1a567b36ac5cda35207c90354e398777f5d

api/src/main/java/net/kyori/adventure/audience/Audience.java
kashike: https://github.com/KyoriPowered/text/pull/62/commits/fd7ce1a567b36ac5cda35207c90354e398777f5d

api/src/main/java/net/kyori/adventure/audience/ForwardingAudience.java
kashike: https://github.com/KyoriPowered/text/pull/62/commits/fd7ce1a567b36ac5cda35207c90354e398777f5d

api/src/main/java/net/kyori/adventure/bossbar/BossBar.java
kashike: https://github.com/KyoriPowered/text/pull/62/commits/fd7ce1a567b36ac5cda35207c90354e398777f5d

api/src/main/java/net/kyori/adventure/sound/SoundImpl.java
kashike: https://github.com/KyoriPowered/text/pull/62/commits/fd7ce1a567b36ac5cda35207c90354e398777f5d

api/src/main/java/net/kyori/adventure/title/Title.java
```diff
@@ -0,0 +1,82 @@
...
+   * @return the subtitle
+   */
+  @NonNull Component subtitle();
+
+  /**
+   * Gets the time (in ticks) the title will fade-in.
+   *
+   * @return the time (in ticks) the title will fade-in
+   */
+  @NonNull Duration fadeInTime();
```
zml2008, line 67: I'm not sure about using Duration here -- title times are in ticks, not wall time, and Duration measures wall time. There was a discussion about this on a Sponge PR for the same sort of design iirc.

api/src/main/java/net/kyori/adventure/title/Title.java
```diff
@@ -0,0 +1,82 @@
...
+   * Creates a title.
+   *
+   * @param title the title
+   * @param subtitle the subtitle
+   * @param fadeInTime the fade-in duration
+   * @param stayTime the stay duration
+   * @param fadeOutTime the fade-out duration
+   * @return the title
+   */
+  static @NonNull Title of(final @NonNull Component title, final @NonNull Component subtitle, final @NonNull Duration fadeInTime, final @NonNull Duration stayTime, final @NonNull Duration fadeOutTime) {
```
zml2008, line 44: Should probably split this up into at least two initializer methods -- since it's valid to send a title without specifically setting times.

Additionally, any single time can be set to a protocol `-1` (representation in-api may be different - while negative Durations are valid, null may be a better way to mark non-set durations) to preserve whatever times were last sent to the client -- it should be clear from JD how those are represented.

api/src/main/java/net/kyori/adventure/title/Title.java
```diff
@@ -0,0 +1,82 @@
...
+    return new TitleImpl(title, subtitle, fadeInTime, stayTime, fadeOutTime);
+  }
+
+  /**
+   * Gets the title.
+   *
+   * @return the title
+   */
+  @NonNull Component title();
```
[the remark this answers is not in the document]
kashike: Idea was to use `TextComponent.empty()`.

api/src/main/java/net/kyori/adventure/title/Title.java
kashike: @Electroid thoughts here?

Electroid: I think a good tradeoff would be using ticks in the interface, but having some `static` constructors that accept either `int ticks` or `Duration time`

kashike: We will stick with `Duration` after some discussion.

PR #66, opened 2020-04-25
Flesh out font handling

zml2008:
Fonts need to default to null to be able to be merged properly.

Add in a merge strategy for fonts as well.

tested working ingame :)


api/src/test/java/net/kyori/adventure/text/format/StyleTest.java
```diff
@@ -168,12 +170,25 @@ void testMerge_insertion() {
     assertEquals(s0, s1.insertion(null));
   }
 
+  private static final Key TEST_FONT = Key.of("kyori", "comicsans");
```
kashike, line 173: oh hell no

PR #75, opened 2020-06-27
serializer-legacy: Add option to downsample text colors when using legacy format

zml2008:
otherwise we start trying to send RGB color codes to old versions, and that looks pretty awkward :)


zml2008:
thing to resolve before this is merged: I used `downsampleColors`, but the gson serializer uses `downsampleColor`. Preference?

PR #76, opened 2020-06-27
Add support for serializing to bungeecord-chat's legacy hex format

lucko:
Pretty gross but it works I suppose.

Is needed in order to support setting things like entity / itemstack custom names to `Component`s on Bukkit.

### DIFF
--- a/text-serializer-legacy/src/main/java/net/kyori/adventure/text/serializer/legacy/LegacyComponentSerializer.java
+++ b/text-serializer-legacy/src/main/java/net/kyori/adventure/text/serializer/legacy/LegacyComponentSerializer.java
@@ -159,13 +159,21 @@ public interface LegacyComponentSerializer extends ComponentSerializer<Component
     @NonNull Builder hexColors();
 
     /**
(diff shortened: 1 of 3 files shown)


text-serializer-legacy/src/main/java/net/kyori/adventure/text/serializer/legacy/LegacyComponentSerializer.java
```diff
@@ -158,6 +158,15 @@ static Builder builder() {
...
     @NonNull Builder hexColors();
 
+    /**
+     * Sets that the serializer should use the '&amp;x' repeated code format when serializing hex colors.
+     *
+     * <p>This is the format adopted by the BungeeCord (and by usage, Spigot) text API.</p>
+     *
+     * @return this builder
+     */
+    @NonNull Builder useXRepeatedCodeHexFormat();
```
Electroid, line 168: `hexColorsWithXs` ?

lucko: Yep that works, except it's singular `x`- only one at the start of the format...

PR #79, opened 2020-06-29
Fix nbt array reading with prefixed indexes

KennyTV:
This makes the old `[0:"hello",1:"hello2"]` list format readable. Since it's always written from 0 upwards (and any other sane legacy lib should do the same, and and it would require changes to the builder otherwise), I simply ignored the actual index.


zml2008:
Thanks Kenny -- would you be able to toss in a test case for this new reading behaviour as well?

KennyTV:
Yup, added a test case!

hugmanrique:
The current code doesn't work if a string contains the `:` character. I would also add an extra test with such case.

zml2008:
`:` characters are not allowed in unquoted strings. Vanilla uses Brigadier for parsing SNBT, so the allowed characters are [these](https://github.com/Mojang/brigadier/blob/master/src/main/java/com/mojang/brigadier/StringReader.java#L169)

hugmanrique:
You're right, my bad :/

PR #80, opened 2020-06-29
audience: Add canSend* methods

zml2008:
These give a rough way of starting to test which types of data an audience can receive.

### DIFF
--- a/api/src/main/java/net/kyori/adventure/audience/Audience.java
+++ b/api/src/main/java/net/kyori/adventure/audience/Audience.java
@@ -87,6 +87,16 @@ public interface Audience {
   // ------------------
 
   /**
+   * Get if there is any way this audience can receive a message.
+   *
+   * <p>Sending a message when this method returns {@code false} will be a no-op.
+   * This method returning true does not guarantee that a message send will succeed.</p>
+   *
+   * @return false if messages cannot be sent to this source
+   */
+  boolean canSendMessage();
+
+  /**
    * Sends a message.
    *
    * @param message the message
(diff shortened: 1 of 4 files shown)


Electroid:
Since `Audience` will no-op, it seems the only use-case for checking if an `Audience` can see something is to try something else.

```java
final Audience steve;

if (steve.canShowBossBar()) {
   steve.showBossBar(BossBar.of(...));
} else {
   steve.showActionBar(TextComponent.of(...));
}
```

The problem is this isn't helpful with groups of `Audience`s, which will be quite common. 

```java
final Audience redTeam;

if (redTeam.canShowBossBar()) {
  // This doesn't work, because don't want to fallback on action bars for the whole team
}
```

Instead, maybe `Audience` methods should return an immutable `Result`?

```java
public interface Result implements Audience {
   // Some other methods?
   Audience toReceived();
   Audience toNotReceived();
}
```

That way you could do something more elegant like this.

```java
final Audience blueTeam;

blueTeam.showBossBar(BossBar.of(...))
    .toNotReceived().showActionBar(TextComponent.of(...))
    .toNotReceived().sendMessage(TextComponent.of(...));
```

zml2008:
I think that would be a good design for a fully-featured way to determine the success/failure of an operation, but I don't know if we want the overhead of building a player collection every message send -- especially since most cases will probably ignore the result. What I'm going for here is a more minimal "this operation will do nothing" check -- for example, if you're on a 1.7.10 server you can know that sending action bars and titles won't do anything without having to try to parse the server version directly.

Thoughts on which design direction makes more sense?

Electroid:
Maybe this is a concern of platform then? I'm not sure we want to encourage this in the api.

zml2008:
I'm going to close this in favour of #87 -- since the approach here is not what we are going to end up going with.

PR #89, opened 2020-06-30
Separate BossBar update logic from the API

astei:
This PR is more meant as a way to spark a discussion about how to best handle boss bars in platforms without resorting to coarse locks to ensure operations are safe in a multithreaded environment. In addition, this abstraction provides a more "natural" abstraction over the underlying boss bar objects in each platform whilst still hiding platform-specific details. I am happy to implement this solution in Velocity.

Specifically, this PR:

* Turns `BossBar` into an immutable class, consistent with the rest of the API. All "mutating" methods now return new `BossBar`s and the `flags()` method now returns an unmodifiable view of the flags
* Adds a new `BossBarContainer` class that `adventure-platform` and other platform implementations are expected to implement, which wraps a `BossBar` and trusts the container to have its own logic for handling boss bars on the client
* `Audience` boss bar methods now expect an `BossBarContainer`

I want to tag @zml2008 as he discussed this topic with me and doesn't like the current solution in `adventure-platform` - deferring the logic of updating and showing boss bars to the platform (without any listeners) might be a better solution than what we have now. (As a nice consequence, if managed correctly, this could used to conveniently control the order in which boss bars appear on the client and allow reordering very easily.)

### DIFF
--- a/api/src/main/java/net/kyori/adventure/audience/Audience.java
+++ b/api/src/main/java/net/kyori/adventure/audience/Audience.java
@@ -24,7 +24,7 @@
 package net.kyori.adventure.audience;
 
 import java.util.Arrays;
-import net.kyori.adventure.bossbar.BossBar;
+import net.kyori.adventure.bossbar.BossBarContainer;
 import net.kyori.adventure.inventory.Book;
 import net.kyori.adventure.sound.Sound;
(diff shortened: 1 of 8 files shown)


astei:
Closing, I found an alternative solution that requires no API changes.

PR #90, opened 2020-07-01
Finish Audience documentation, solidify the interface

Electroid:
- Removed `WeakAudience`, it didn't have much use
- Documented the specification and motivation for `Audience`
- Merged `MultiAudience` into `ForwardingAudience`
- All methods except `Audience#sendMessage` are stubbed


api/src/main/java/net/kyori/adventure/audience/Audience.java
```diff
@@ -24,163 +24,199 @@
...
+   * @param audiences an array of audiences, can be empty
    * @return an audience
    */
   static @NonNull Audience of(final @NonNull Audience@NonNull... audiences) {
-    final int length = audiences.length;
-    if(length == 0) {
-      return empty();
-    } else if(length == 1) {
-      return audiences[0];
+    switch(audiences.length) {
```
kashike, line 99: Undo conversion to `switch` here.

api/src/main/java/net/kyori/adventure/audience/Audience.java
```diff
@@ -24,163 +24,199 @@
...
-
-  // ------------------
-  // ---- Messages ----
-  // ------------------
-
-  /**
-   * Sends a message.
-   *
-   * @param message the message
+   * @since 1.0
```
kashike, line 124: Do you mean `4.0.0` (applies to all cases throughout file)?

api/src/main/java/net/kyori/adventure/audience/Audience.java
```diff
@@ -24,163 +24,199 @@
...
+   * @see Component
+   * @param message a message
    */
-  void sendActionBar(final @NonNull Component message);
-
-  // ----------------
-  // ---- Titles ----
-  // ----------------
+  default void sendActionBar(final @NonNull Component message) {}
```
kashike, line 137: Why becoming `default`?

api/src/main/java/net/kyori/adventure/audience/EmptyAudience.java
```diff
@@ -23,58 +23,29 @@
...
 import net.kyori.adventure.text.Component;
-import net.kyori.adventure.title.Title;
 import org.checkerframework.checker.nullness.qual.NonNull;
 
 /* package */ final class EmptyAudience implements Audience {
   /* package */ static final EmptyAudience INSTANCE = new EmptyAudience();
 
   @Override
-  public void sendMessage(final @NonNull Component message) {
+  public void sendMessage(@NonNull Component message) {
```
kashike, line 33: Should stay `final`.

api/src/main/java/net/kyori/adventure/audience/Audience.java
Electroid: The contract is `sendMessage` is required, but everything else is stub by default. This important since lots of implementations will be extending `Audience`. We don't want push back that we're going to break everyone because we add some new method.

api/src/main/java/net/kyori/adventure/audience/Audience.java
```diff
@@ -58,129 +101,125 @@
       return empty();
     } else if(length == 1) {
       return audiences[0];
+    } else {
+      return of(Arrays.asList(audiences));
```
liach, line 105: To keep code clear, you shouldn't write `else` after a `return`.

api/src/main/java/net/kyori/adventure/audience/Audience.java
```diff
@@ -58,129 +101,125 @@
...
+   * made will be reflected in <tt>Audience</tt>.</p>
+   *
+   * @since 4.0.0
+   * @see ForwardingAudience
+   * @param audiences an iterable of audiences, can be empty
    * @return an audience
    */
   static @NonNull Audience of(final @NonNull Iterable<? extends Audience> audiences) {
-    return (MultiAudience) () -> audiences;
+    return (ForwardingAudience) audiences;
```
liach, line 121:
```suggestion
    return (ForwardingAudience) () -> audiences;
```

api/src/main/java/net/kyori/adventure/audience/Audience.java
kashike: Why make `sendMessage` required and everything else not?

Electroid: If an `Audience` doesn't implement the basic message, it's probably not the right abstraction? We could also make this not require too.

api/src/main/java/net/kyori/adventure/audience/Audience.java
Electroid: Curious what the difference is here.

zml2008: one is a cast, one creates an instance of the functional interface with an explicit type hint

api/src/main/java/net/kyori/adventure/audience/Audience.java
zml2008: We already go against a required `sendMessage` in other parts of our API -- everywhere we return Audience.empty() instead of null, for example. It would only create confusion to claim that sendMessage is always required.

I'm also against Audience itself having default implementations -- maybe a subinterface or abstract audience could have the defaults?

Electroid: Defaulting on the man interface is the easiest way to future proof it, while also allowing implementors to extend Audience directly and not though a stub. Given the flexibility, I can also see plugins implementing this in their own `MyPluginPlayer` interfaces and what not. It will be a lot easier for us and them when we add the new `playX` feature, it doesn't break any code.

api/src/main/java/net/kyori/adventure/audience/Audience.java
liach: yes, the `() ->` is to create a method handle that returns the local var on call

PR #91, opened 2020-07-02
Introduce View

Electroid:
`View` is an interface with a `UUID id()` that represents some persistent, unique view. This is helpful for both users and implementations to have a unique id that can be keyed for a `BossBar`, or in the future, a potential `TabList`.

### DIFF
--- a/api/src/main/java/net/kyori/adventure/audience/Audience.java
+++ b/api/src/main/java/net/kyori/adventure/audience/Audience.java
@@ -24,6 +24,8 @@
 package net.kyori.adventure.audience;
 
 import java.util.Arrays;
+import java.util.UUID;
+
 import net.kyori.adventure.bossbar.BossBar;
 import net.kyori.adventure.inventory.Book;
 import net.kyori.adventure.sound.Sound;
@@ -139,9 +141,9 @@ public interface Audience {
   /**
    * Hides a bossbar.
    *
-   * @param bar the bossbar
+   * @param barId the bossbar id
    */
-  void hideBossBar(final @NonNull BossBar bar);
+  void hideBossBar(final @NonNull UUID barId);
 
   // ----------------
   // ---- Sounds ----
--- a/api/src/main/java/net/kyori/adventure/bossbar/BossBar.java
+++ b/api/src/main/java/net/kyori/adventure/bossbar/BossBar.java
@@ -26,12 +26,13 @@ package net.kyori.adventure.bossbar;
 import java.util.Set;
 import net.kyori.adventure.text.Component;
 import net.kyori.adventure.util.Index;
+import net.kyori.adventure.util.View;
 import org.checkerframework.checker.nullness.qual.NonNull;
 
 /**
  * A bossbar.
  */
-public interface BossBar {
+public interface BossBar extends View {
   /**
    * The minimum value the percent can be.
    */
@@ -52,7 +53,6 @@ public interface BossBar {
    * @throws IllegalArgumentException if percent is less than 0 or greater than 1
    */
   static @NonNull BossBar of(final @NonNull Component name, final float percent, final @NonNull Color color, final @NonNull Overlay overlay) {
-    BossBarImpl.checkPercent(percent);
     return new BossBarImpl(name, percent, color, overlay);
   }
 
@@ -68,7 +68,6 @@ public interface BossBar {
    * @throws IllegalArgumentException if percent is less than 0 or greater than 1
    */
   static @NonNull BossBar of(final @NonNull Component name, final float percent, final @NonNull Color color, final @NonNull Overlay overlay, final @NonNull Set<Flag> flags) {
-    BossBarImpl.checkPercent(percent);
     return new BossBarImpl(name, percent, color, overlay, flags);
   }
 
--- a/api/src/main/java/net/kyori/adventure/audience/EmptyAudience.java
+++ b/api/src/main/java/net/kyori/adventure/audience/EmptyAudience.java
@@ -31,6 +31,8 @@ import net.kyori.adventure.text.Component;
 import net.kyori.adventure.title.Title;
 import org.checkerframework.checker.nullness.qual.NonNull;
 
+import java.util.UUID;
+
 /* package */ final class EmptyAudience implements Audience {
   /* package */ static final EmptyAudience INSTANCE = new EmptyAudience();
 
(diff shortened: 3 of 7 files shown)


liach:
What's the advantage of an uuid over a minecraft key like vanilla minecraft uses?

kashike:
> 
> 
> What's the advantage of an uuid over a minecraft key like vanilla minecraft uses?

Vanilla uses a `UUID` for all bossbars. Only bossbars that are serialised/created by commands use a `Key`.

api/src/main/java/net/kyori/adventure/bossbar/BossBar.java
```diff
@@ -52,7 +53,6 @@
    * @throws IllegalArgumentException if percent is less than 0 or greater than 1
    */
   static @NonNull BossBar of(final @NonNull Component name, final float percent, final @NonNull Color color, final @NonNull Overlay overlay) {
-    BossBarImpl.checkPercent(percent);
```
kashike, line 55: This is intentional, checks before attempting to create it this way.

api/src/main/java/net/kyori/adventure/bossbar/BossBar.java
```diff
@@ -68,7 +68,6 @@
    * @throws IllegalArgumentException if percent is less than 0 or greater than 1
    */
   static @NonNull BossBar of(final @NonNull Component name, final float percent, final @NonNull Color color, final @NonNull Overlay overlay, final @NonNull Set<Flag> flags) {
-    BossBarImpl.checkPercent(percent);
```
kashike, line 71: This is intentional, checks before attempting to create it this way.

api/src/main/java/net/kyori/adventure/bossbar/BossBar.java
liach: meaning other ways of bossbar creation may have out-of-range percentage?

Electroid: Having a check inside the constructor seems reasonable? Also what liach said.

liach: Check inside constructor is bad if out-of-range percentages may exist.

api/src/main/java/net/kyori/adventure/audience/Audience.java
```diff
@@ -139,9 +141,9 @@
   /**
    * Hides a bossbar.
    *
-   * @param bar the bossbar
+   * @param barId the bossbar id
    */
-  void hideBossBar(final @NonNull BossBar bar);
+  void hideBossBar(final @NonNull UUID barId);
```
zml2008, line 146: Why the change to a UUID? and why only for the `hide` operation?

We only track + control Adventure `BossBar` instances, so I don't see how it makes sense to pass an arbitrary uuid -- this feels like it will only be used as `viewer.hideBossBar(bar.id())`.

liach: well this allows us to hide bossbars without a reference to an instance, which would be safer. boss bars are identified by unique id as described by kashike above, so i think this is an improvement. for show operation, the contents of the bar is necessary since the bar can be new.

zml2008: Where will you get a UUID except from an instance?

liach: The point is to keep track of a uuid than an instance to prevent resource leaks.

zml2008: You still need to hold on to an instance to be able to change the contents of the boss bar, and boss bar listeners need to hold on to copies of the instance to ensure they remove themselves from its list of listeners. Where do you see these resource leaks coming from that would be resolved by using a UUID?

api/src/main/java/net/kyori/adventure/bossbar/BossBar.java
Electroid: I don't follow

api/src/main/java/net/kyori/adventure/audience/Audience.java
Electroid: Not all cases will make changes to a boss bar once shown, it also seems silly to enforce users to hold the entire object. Every implementation is only going to care about the `UUID`.

Electroid: An alternative idea is doing something like this:

```java
interface Id {
  UUID id();
}

interface Audience {
  hideBossBar(Id barId);
}

Audience a;
a.hideBossBar(Id.of(...));
a.hideBossBar(BossBar.of(...));
```

theminecoder: As an api user I would rather not have to extract and possibly store the uuid just to remove the boss bar from the player.  I might as well just store the whole bossbar where I get the added options of modifying the content. It's making the api complicated just for a potential tiny memory saving (which I doubt will happen anyway as the implementation still has the active bossbar until all players are deregistered).

Electroid: Thought on above? ^ That way you can use either `BossBar` or `Id`

zml2008: I don't think that's worth the added complexity. Boss bars cannot be safely implemented without having the implementation hold a reference to the `BossBar` instance, and at that point a pointer is a pointer and you might as well hold on to the more useful one.

liach: Well you can always call the uuid version when you have the bossbar instance. like for a command node, you have a `remove(String)` while you have a `register(CommandNode)`. The fact that you don't have a `remove(CommandNode)` doesn't affect your usage at all. Same logic for uuid here.

theminecoder: Still not great, what if players make their own id instance and pass it to the method, we are having to introduce additional logic whereas the type system already has this solved for us.

liach: > 
> 
> Still not great, what if players make their own id instance and pass it to the method, we are having to introduce additional logic whereas the type system already has this solved for us.

Equivalent scenario:
We have modified brigadier so that you can call `remove(String)` on a command node to remove children.

"Still not great, what if players make their own string instance and pass it to the method, we are having to introduce additional logic whereas the type system already has this solved for us."

kashike: Definitely not a fan of changing these methods to accept `UUID` alone. What is the upside to allowing not a full `BossBar` to be passed?

Electroid:
Closing in favor of #94

PR #98, opened 2020-07-12
Improve performance of BinaryTagIO by buffering and clarify stream closing

astei:
While looking at `adventure-nbt` I spotted two issues:

* Lack of stream buffering. This can hurt performance, especially for compressed NBT and slow disks.
* `BinaryTagIO` will indiscriminately close streams given to it. For single-shot cases or reading from a in-memory buffer this isn't an issue, but if (say) you were parsing a set of NBT streams this could be an issue.

To fix this:

* `Path`-taking methods and `*Compressed*` methods now buffer the relevant streams (I assume if you're passing in an `InputStream` or a `DataInput` directly that you know what you're doing)
* Added code so that `BinaryTagIO` won't close streams it is given

Still a bit of a WIP (need to write tests), but wanted to get this up here.

### DIFF
--- a/nbt/src/main/java/net/kyori/adventure/nbt/IOStreamUtil.java
+++ b/nbt/src/main/java/net/kyori/adventure/nbt/IOStreamUtil.java
@@ -1,0 +1,72 @@
+/*
+ * This file is part of adventure, licensed under the MIT License.
+ *
+ * Copyright (c) 2017-2020 KyoriPowered
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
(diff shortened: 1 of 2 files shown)


nbt/src/main/java/net/kyori/adventure/nbt/IOStreamUtil.java
```diff
@@ -0,0 +1,72 @@
...
+ */
+package net.kyori.adventure.nbt;
+
+import java.io.IOException;
+import java.io.InputStream;
+import java.io.OutputStream;
+
+/* package */ final class IOStreamUtil {
+  private IOStreamUtil() {
+    throw new AssertionError();
```
kashike, line 32: ?

astei: This class just has decorators for blocking `close()` on `InputStream`s and `OutputStream`s. It's a utility class and thus should not be instantiated. The exception throwing is just an extra precaution.

kashike: Can I trouble you to remove the `throw`? It is inconsistent with every other class.

PR #100, opened 2020-07-13
minor additions and changes

kashike:


api/src/main/java/net/kyori/adventure/audience/Audience.java
```diff
@@ -127,6 +127,26 @@
...
+   * @see Component
+   * @since 4.0.0
+   */
+  default void sendMessage(final @NonNull Component message, final @NonNull MessageType type) {
+  }
+
+  /**
+   * Message types.
+   */
+  enum MessageType {
```
Electroid, line 147: Can we move this to be an independent class, outside of Audience?

kashike: Sure.

PR #101, opened 2020-07-13
Update readme.md for more 4.0-ness

ItsTehBrian:
Changed instances of `TextColor` to `NamedTextColor` where applicable, and renamed one instance of `text` to `adventure`.

The serialization examples don't seem to work, but I have no idea what they're *supposed* to be lol


zml2008:
Thanks for the contribution! I'm afraid we have started moving towards a new documentation system so we won't really be using the README anymore. However, I've converted your changes into the new format -- check the referenced commit to see those, and feel free to toss a comment if you feel anything didn't get converted appropriately.

PR #102, opened 2020-07-14
Introduce translation key framework

Electroid:
Both plugins and platforms often require translation keys. It's a pain for each platform to try and do their own thing when the use-case is almost universally the same: "I have a resource bundle, please automatically translate them for me based on a viewer's locale."

The idea is to introduce a `feature-translation`, that is highly recommended to be implemented by platforms, but not explicitly required. So when a user sends a `TranslatableComponent` to an `Audience`, it is automatically translated by checking the `TranslationRegistry`.

This is a first draft, so welcoming ideas.


text-feature-translation/src/main/java/net/kyori/adventure/text/feature/translation/ResourceBundleTranslator.java
```diff
@@ -0,0 +1,168 @@
...
+   * @param locale a locale
+   * @return {@code true} if the locale has translations
+   */
+  public boolean registerLocale(final @NonNull Locale locale) {
+    if (this.resources.containsKey(locale)) return false;
+
+    final String path = MessageFormat.format(this.resourceFormat, locale.toLanguageTag());
+    final ResourceBundle resource;
+    try {
+      resource = ResourceBundle.getBundle(path, locale, UTF8);
```
zml2008, line 77: Don't ResourceBundles already have their own cache, and locale resolution? It might make more sense to override the `ResourceBundle`'s own resolution rather than introducing another layer

text-feature-translation/src/main/java/net/kyori/adventure/text/feature/translation/ResourceBundleTranslator.java
```diff
@@ -0,0 +1,168 @@
...
+  }
+
+  /**
+   * Registers a new locale.
+   *
+   * @param locale a locale
+   * @return {@code true} if the locale has translations
+   */
+  public boolean registerLocale(final @NonNull Locale locale) {
+    if (this.resources.containsKey(locale)) return false;
```
zml2008, line 72: is this supposed to return `true`?

text-feature-translation/src/main/java/net/kyori/adventure/text/feature/translation/TranslationRegistryImpl.java
```diff
@@ -125,12 +137,15 @@
 
     // no arguments makes this render very simple
     if(args.isEmpty()) {
-      return builder.content(format.format(null, new StringBuffer(), null).toString()).build();
+      return builder.content(format).build();
     }
 
+    // TODO: MessageFormat has unnecessary overhead over String, rewrite this code to use String
+    final MessageFormat messageFormat = new MessageFormat(format);
```
zml2008, line 144: Why not have the translator return a MessageFormat? It should be possible to have a ResourceBundle instance that caches MessageFormat instances (probably within a ThreadLocal since the formats are not thread-safe).

text-feature-translation/src/main/java/net/kyori/adventure/text/feature/translation/TranslationRegistry.java
```diff
@@ -0,0 +1,52 @@
...
+/**
+ * A registry of translators for rendering components.
+ */
+public interface TranslationRegistry extends Translator, ComponentRenderer<Locale> {
+  /**
+   * Gets the translation registry.
+   *
+   * @return the translation registry
+   */
+  static TranslationRegistry get() {
```
zml2008, line 40: I don't know if I like the idea of global stored state -- this seems like an easy way to accidentally end up with message conflicts.

api/src/main/java/net/kyori/adventure/key/DuplicateKeyException.java
```diff
@@ -0,0 +1,55 @@
...
+ */
+package net.kyori.adventure.key;
+
+import org.checkerframework.checker.nullness.qual.NonNull;
+import org.checkerframework.checker.nullness.qual.Nullable;
+
+import static java.util.Objects.requireNonNull;
+
+/**
+ * When a duplicate key is created.
```
astei, line 32: `This exception is thrown when a duplicate key has been created. It includes the key that was created in its place.`

api/src/main/java/net/kyori/adventure/key/InvalidKeyException.java
```diff
@@ -0,0 +1,55 @@
...
+ */
+package net.kyori.adventure.key;
+
+import org.checkerframework.checker.nullness.qual.NonNull;
+import org.checkerframework.checker.nullness.qual.Nullable;
+
+import static java.util.Objects.requireNonNull;
+
+/**
+ * When an invalid key is created.
```
astei, line 32: `This exception is thrown when an invalid key has been created.`

text-feature-translation/src/main/java/net/kyori/adventure/text/feature/translation/TranslationRegistry.java
astei: Agreed. I think it should be the platform's job to provide the translation registry for plugins.

text-feature-translation/src/main/java/net/kyori/adventure/text/feature/translation/TranslationRegistryImpl.java
```diff
@@ -0,0 +1,125 @@
...
+
+  private static final class Translation implements Key, Examinable {
+    private final String owner;
+    private final String key;
+    private final Map<Locale, MessageFormat> formats;
+
+    private Translation(final @NonNull Key key) {
+      this.owner = key.namespace();
+      this.key = key.value();
+      this.formats = new ConcurrentSkipListMap<>(Comparator.comparing(Locale::toLanguageTag));
```
astei, line 51: Does this have to be a `ConcurrentSkipListMap`? This could just as easily been a `ConcurrentHashMap` (I assume you meant to make this class thread-safe).

text-feature-translation/src/main/java/net/kyori/adventure/text/feature/translation/TranslationRegistryImpl.java
```diff
@@ -0,0 +1,125 @@
...
+    public @NonNull Stream<? extends ExaminableProperty> examinableProperties() {
+      return Stream.of(
+        ExaminableProperty.of("owner", this.owner),
+        ExaminableProperty.of("key", this.key),
+        ExaminableProperty.of("formats", this.formats)
+      );
+    }
+  }
+
+  private final Map<String, Translation> translations = new ConcurrentSkipListMap<>();
```
astei, line 92: Does this have to be a `ConcurrentSkipListMap`? This could just as easily been a `ConcurrentHashMap` (I assume you meant to make this class thread-safe).

text-feature-translation/src/main/java/net/kyori/adventure/text/feature/translation/TranslationRegistryImpl.java
```diff
@@ -0,0 +1,125 @@
...
+  }
+
+  @Override
+  public void unregisterAll(@NonNull String namespace) {
+    this.translations.values().removeIf(translation -> translation.namespace().equals(namespace));
+  }
+
+  @Override
+  public @Nullable MessageFormat translate(@NonNull String key, @NonNull Locale locale) {
+    final Translation recentTranslation = this.recentTranslation.get();
```
astei, line 111: This smells like premature optimization.

text-feature-translation/src/main/java/net/kyori/adventure/text/feature/translation/TranslationRegistry.java
Electroid: I would agree platforms should handle the translation rendering layer, but I think having a platform-neutral entrypoint for keys is the right way to go. Keys are typically neutral of platform.

api/src/main/java/net/kyori/adventure/translate/TranslationRegistry.java
```diff
@@ -0,0 +1,149 @@
...
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+ * SOFTWARE.
+ */
+package net.kyori.adventure.translate;
```
kashike, line 24: Could you move these to `net.kyori.adventure.translation`?

api/src/main/java/net/kyori/adventure/translate/Translator.java
```diff
@@ -0,0 +1,44 @@
...
+import java.text.MessageFormat;
+import java.util.Locale;
+import org.checkerframework.checker.nullness.qual.NonNull;
+import org.checkerframework.checker.nullness.qual.Nullable;
+
+/**
+ * A message format translator.
+ */
+@FunctionalInterface
+public interface Translator {
```
kashike, line 35: `TranslationSource`, maybe?

api/src/main/java/net/kyori/adventure/translate/TranslationRegistry.java
```diff
@@ -0,0 +1,149 @@
...
+import java.util.MissingResourceException;
+import java.util.ResourceBundle;
+import java.util.regex.Pattern;
+import net.kyori.adventure.util.UTF8ResourceBundleControl;
+import org.checkerframework.checker.nullness.qual.NonNull;
+
+/**
+ * A registry of translations.
+ */
+public interface TranslationRegistry extends Translator {
```
Cybermaxke, line 40: How will this support translations that are translated using `String.format`? This is what vanilla uses.

Maybe we need a common format interface which can both represent `MessageFormat` and `String.format`?

api/src/main/java/net/kyori/adventure/translate/TranslationRegistry.java
```diff
@@ -0,0 +1,149 @@
...
+  /**
+   * Registers a resource bundle of translations.
+   *
+   * @param locale a locale
+   * @param resourceBundlePath a resource bundle path
+   * @param escapeSingleQuotes whether to escape single quotes
+   * @throws IllegalArgumentException if a translation key is already exists
+   * @see #registerAll(Locale, ResourceBundle, boolean)
+   */
+  default void registerAll(final @NonNull Locale locale, final @NonNull String resourceBundlePath, final boolean escapeSingleQuotes) {
```
Cybermaxke, line 133: Vanilla uses json files for translations, it would be interesting to have a resource bundle loader for this as well.

api/src/main/java/net/kyori/adventure/translate/TranslationRegistry.java
kashike: This is not designed to handle vanilla translations, as those are done on the client.

Cybermaxke: Still, it would be nice to have support for `String.format`?

Cybermaxke: Maybe using `Format` (the superclass) of `MessageFormat` is an option? This will still allow custom formats without explicitly supporting them

PR #111, opened 2020-07-20
Don't use streams when checking validity of namespace and value

astei:


api/src/main/java/net/kyori/adventure/key/KeyImpl.java
```diff
@@ -45,11 +45,21 @@
   }
 
   /* package */ static boolean namespaceValid(final @NonNull String namespace) {
-    return namespace.chars().allMatch(NAMESPACE_PREDICATE);
+    for(int i = 0; i < namespace.length(); i++) {
```
kashike, line 48:
```suggestion
    for(int i = 0, length = namespace.length(); i < length; i++) {
```

api/src/main/java/net/kyori/adventure/key/KeyImpl.java
```diff
@@ -45,11 +45,21 @@
...
+      if(!NAMESPACE_PREDICATE.test(namespace.charAt(i))) {
+        return false;
+      }
+    }
+    return true;
   }
 
   /* package */ static boolean valueValid(final @NonNull String value) {
-    return value.chars().allMatch(VALUE_PREDICATE);
+    for(int i = 0; i < value.length(); i++) {
```
kashike, line 57:
```suggestion
    for(int i = 0, length = value.length(); i < length; i++) {
```

PR #112, opened 2020-07-20
Add default string to TextComponent#make()

theminecoder:
Allows use of this method by only appending other components without having to remember to give it a string


kashike:
Thank you for the pull request, but `content` [already defaults to empty](https://github.com/KyoriPowered/adventure/blob/master/api/src/main/java/net/kyori/adventure/text/TextComponentImpl.java#L166).

theminecoder:
Look at you already solving the one annoyance of text I had to fight with.

PR #128, opened 2020-08-08
Update build tool configurations

PiggyPiglet:
adventure isn't hosted on central, it's hosted on sonatype. the compile keyword is also deprecated in gradle, for this specific use case, implementation should be used instead. Not sure if maven repository declarations require an id or name; I don't use maven.

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -15,6 +15,10 @@ There are various artifacts:
 
 * Maven
 ```xml
+<repository>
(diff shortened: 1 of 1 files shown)


zml2008:
I'm afraid the READMEs are just outdated -- we're moving our documentation over to https://docs.adventure.kyori.net/, which already contains this information.

PR #131, opened 2020-08-13
Fix empty compound parsing

KennyTV:
Previously, parsing would fail for compounds like `{}` or `{a: \"hello\", empty: {}}`

### DIFF
--- a/nbt/src/main/java/net/kyori/adventure/nbt/TagStringReader.java
+++ b/nbt/src/main/java/net/kyori/adventure/nbt/TagStringReader.java
@@ -24,7 +24,6 @@
 package net.kyori.adventure.nbt;
 
 import java.util.ArrayList;
-import java.util.HashMap;
 import java.util.List;
 import java.util.stream.IntStream;
 import java.util.stream.LongStream;
--- a/nbt/src/main/java/net/kyori/adventure/nbt/TagStringReader.java
+++ b/nbt/src/main/java/net/kyori/adventure/nbt/TagStringReader.java
@@ -24,6 +24,7 @@
 package net.kyori.adventure.nbt;
 
 import java.util.ArrayList;
+import java.util.HashMap;
 import java.util.List;
 import java.util.stream.IntStream;
 import java.util.stream.LongStream;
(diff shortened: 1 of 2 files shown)


nbt/src/test/java/net/kyori/adventure/nbt/StringIOTest.java
```diff
@@ -242,6 +242,12 @@ public void testLongArrayTag() throws IOException {
     assertEquals(LongArrayBinaryTag.of(2, 4, 6, -8, 10, 12), this.stringToTag("[L; 2l, 4l, 6l, -8l, 10l, 12l]"));
   }
 
+  @Test
+  public void testEmptyCompoundTag() throws StringTagParseException {
+    final TagStringReader read = new TagStringReader(new CharBuffer("{}"));
+    assert read.compound().keySet().isEmpty();
```
lucko, line 248: Use `assertTrue(...)` here instead of `assert` :)

nbt/src/main/java/net/kyori/adventure/nbt/TagStringReader.java
```diff
@@ -38,6 +39,11 @@
 
   public CompoundBinaryTag compound() throws StringTagParseException {
     this.buffer.expect(Tokens.COMPOUND_BEGIN);
+    if(this.buffer.peek() == Tokens.COMPOUND_END) {
+      this.buffer.take();
+      return new CompoundBinaryTagImpl(new HashMap<>());
```
kashike, line 44:
```suggestion
      return CompoundBinaryTag.empty();
```

KennyTV: Oh yeah I didn't see that the EMPTY map is wrapped into a new one when the tag is edited, I assumed it would make it immutable

nbt/src/main/java/net/kyori/adventure/nbt/TagStringReader.java
```diff
@@ -24,6 +24,7 @@
 package net.kyori.adventure.nbt;
 
 import java.util.ArrayList;
+import java.util.HashMap;
```
zml2008, line 27:
```suggestion

```
unused import (thanks checkstyle!)

apart from that, seems good. if you feel motivated, it'd be good to do similar tests for empty lists + arrays, but i don't mind doing those.

KennyTV: Not sure how empty lists/arrays are handled by MC itself, but in adventure, `[]` becomes a list with one entry being an empty string (so `[""]`) and parsing `[L;]` fails - but I'd rather have you look at these :p

zml2008: ah -- i think both of those are incorrect behaviours :p

PR #133, opened 2020-08-14
api: titles can be in ticks, not just a Duration

kashike:


Cybermaxke:
I preferred the approach with `DurationOrTicks`, for a couple of reasons.
- `Int` can be of any unit, seconds, hours, ticks, millis? Using `OfTicks` makes that clear that you want to use ticks, `OfDuration` makes it clear that you want clock based duration.
- Implementations that don't use "ticks" have greater control. If you provide a clock based duration, then it would take the actual time, and not an estimation by converting ticks to time.
  As an example in sponge, if you would create a scheduled task with a delay of "10 ticks", it would be 10 minecraft ticks. If you provided a Duration, it will use the system nano time. `OfTicks` != `OfDuration`
- Why do you want to specify time with ticks, if you can use seconds/minutes anyway? I know this is library is minecraft related, but I would prefer to say "show the title for 1 second" over "show the title for 20 ticks".
- Related to https://github.com/SpongePowered/SpongeAPI/pull/2107, using `Int` goes against the goal I am trying to achieve in sponge

Maybe it would be better to rename `DurationOrTicks` to `TimeOrTicks` though (or something else), since `OfTicks` and `OfDuration` are both "durations".

lucko:
> Int can be of any unit, seconds, hours, ticks, millis? Using OfTicks makes that clear that you want to use ticks, OfDuration makes it clear that you want clock based duration.

Agreed - the constructor can always be changed to clarify that.

> Implementations that don't use "ticks" have greater control. If you provide a clock based duration, then it would take the actual time, and not an estimation by converting ticks to time.
As an example in sponge, if you would create a scheduled task with a delay of "10 ticks", it would be 10 minecraft ticks. If you provided a Duration, it will use the system nano time. OfTicks != OfDuration

Indeed, but this is not a scheduling library. We're talking about titles, which when rendered by a Minecraft client are timed using ticks, and ticks only.

> Why do you want to specify time with ticks, if you can use seconds/minutes anyway? I know this is library is minecraft related, but I would prefer to say "show the title for 1 second" over "show the title for 20 ticks".

What's the harm in supporting both?!

kashike:
Another thing to consider is, right now the only time-related thing we have is titles, but that might change in the future. Being consistent is key, and is why I went with just `Duration` originally.

api/src/main/java/net/kyori/adventure/title/TitleImpl.java
```diff
@@ -99,14 +80,71 @@ public int hashCode() {
...
+      ExaminableProperty.of("times", this.times)
     );
   }
 
   @Override
   public String toString() {
     return this.examine(StringExaminer.simpleEscaping());
   }
+
+  static class TimesImpl implements Examinable, Times {
```
Electroid, line 92: I feel this is only going to increase boilerplate. `Title#fadeIn` becomes `Title#times#fadeIn`. Is there a pragmatic reason we need a `Times` class?

kashike:
After discussing more with @Cybermaxke and others, I am going to close this pull request and we will stick with `Duration`.

PR #136, opened 2020-08-15
api: add ComponentMessageThrowable

kashike:


api/src/main/java/net/kyori/adventure/util/ComponentMessageThrowable.java
```diff
@@ -0,0 +1,73 @@
...
+ * An extension interface for {@link Throwable}s to provide a {@link Component}-based message.
+ */
+public interface ComponentMessageThrowable {
+  /**
+   * Gets the {@link Component}-based message from a {@link Throwable}, if available.
+   *
+   * @param throwable the throwable
+   * @return the message
+   */
+  static @Nullable Component get(final @Nullable Throwable throwable) {
```
octylFractal, line 40: Copy-pasta from Discord:
I don't think I like the sound of `ComponentMessageThrowable.get(Throwable)` -> `Component`?
static "get" usually implies it returns the type of the containing class
would prefer `getMessage(Throwable)`
and `getOrConvertMessage(Throwable)`

PR #137, opened 2020-08-16
text-serializer-gson: Ensure GsonComponentSerializer#toBuilder() carries over legacy hover event serializer configuration

astei:
Just a bug I noticed while working on something unrelated.


text-serializer-gson/src/main/java/net/kyori/adventure/text/serializer/gson/GsonComponentSerializer.java
```diff
@@ -101,8 +101,13 @@ static Builder builder() {
     @NonNull Builder legacyHoverEventSerializer(final @NonNull LegacyHoverEventSerializer serializer);
 
     /**
-     * Output a legacy hover event {@code value} in addition to the modern {@code contents}
-     *
+     * Clears the serializer used to interpret and emit legacy hover event {@code value} payloads.
+     * @return this builder
+     */
+    @NonNull Builder clearLegacyHoverEventSerializer();
```
kashike, line 107: I think it might be better to just make `legacyHoverEventSerializer` accept a `@Nullable` parameter instead?

astei: That's reasonable. I can make this change.

PR #144, opened 2020-08-28
Revise text component replacement

zml2008:
This now properly handles non-`TextComponent` children, plus traverses `TranslatableComponent`s and hover events.

At some point the logic in this and in `TranslatableComponentRenderer` should be abstracted out into a common component visitor. There's also potential to expand the exposed replacement api -- this PR doesn't expose any new functionality.

### DIFF
--- a/api/src/main/java/net/kyori/adventure/text/AbstractComponent.java
+++ b/api/src/main/java/net/kyori/adventure/text/AbstractComponent.java
@@ -85,7 +85,7 @@ public abstract class AbstractComponent implements Component, Examinable {
 
   @Override
   public @NonNull Component replaceText(final @NonNull Pattern pattern, final @NonNull UnaryOperator<TextComponent.Builder> replacement, final @NonNull IntFunction2<PatternReplacementResult> fn) {
-    return TextReplacementRenderer.INSTANCE.render(this, new TextReplacementRenderer.State(pattern, (result, build) -> replacement.apply(build), fn));
+    return TextReplacementRenderer.INSTANCE.render(this, new TextReplacementRenderer.State(pattern, (result, builder) -> replacement.apply(builder), fn));
   }
 
   @Override
--- a/api/src/main/java/net/kyori/adventure/text/AbstractComponent.java
+++ b/api/src/main/java/net/kyori/adventure/text/AbstractComponent.java
@@ -27,8 +27,11 @@ import java.util.ArrayList;
 import java.util.Collections;
 import java.util.List;
 import java.util.Objects;
+import java.util.function.UnaryOperator;
+import java.util.regex.Pattern;
 import java.util.stream.Stream;
 import net.kyori.adventure.text.format.Style;
+import net.kyori.adventure.util.IntFunction2;
 import net.kyori.examination.Examinable;
 import net.kyori.examination.ExaminableProperty;
 import net.kyori.examination.string.StringExaminer;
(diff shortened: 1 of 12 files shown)


api/src/main/java/net/kyori/adventure/text/TextComponentImpl.java
```diff
@@ -79,67 +80,145 @@
...
-            if(!prefix.isEmpty()) {
-              produced.add(withoutChildren.content(prefix));
-            }
+  public @NonNull Component replace(final @NonNull Pattern pattern, final @NonNull UnaryOperator<Builder> replacement,
+          final @NonNull IntFunction2<PatternReplacementResult> fn) {
+    return new Replacer(pattern, (result, build) -> replacement.apply(build), fn).replace(this);
+  }
+
+  static class Replacer {
+    private final @NonNull Pattern pattern;
```
kashike, line 89: `@NonNull ` is not required on fields.

api/src/main/java/net/kyori/adventure/text/event/HoverEvent.java
```diff
@@ -244,6 +247,9 @@ public ShowItem(final @NonNull Key item, final @NonNegative int count, final @Nu
      * @return a {@code ShowItem}
      */
     public @NonNull ShowItem count(final @NonNegative int count) {
+      if(count == this.count) {
+        return this;
+      }
```
kashike, line 252:
```suggestion
      if(count == this.count) return this;
```
applies to all cases like this.

api/src/main/java/net/kyori/adventure/text/renderer/ComponentRenderer.java
```diff
@@ -40,4 +41,15 @@
...
   @NonNull Component render(final @NonNull Component component, final @NonNull C context);
+
+  /**
+   * Return a {@link ComponentRenderer} that takes a different context type
+   *
+   * @param transformer context type transformer
+   * @param <T> transformation function
+   * @return mapping renderer
+   */
+  default <T> ComponentRenderer<T> map(final Function<T, C> transformer) {
```
kashike, line 52: `mapContext`?

api/src/main/java/net/kyori/adventure/text/AbstractComponent.java
```diff
@@ -80,6 +83,11 @@ protected AbstractComponent(final @NonNull List<? extends ComponentLike> childre
     return this.style;
   }
 
+  @Override
+  public @NonNull Component replaceText(final @NonNull Pattern pattern, final @NonNull UnaryOperator<TextComponent.Builder> replacement, final @NonNull IntFunction2<PatternReplacementResult> fn) {
+    return TextReplacementRenderer.INSTANCE.render(this, new TextReplacementRenderer.State(pattern, (result, build) -> replacement.apply(build), fn));
```
kashike, line 88:
```suggestion
    return TextReplacementRenderer.INSTANCE.render(this, new TextReplacementRenderer.State(pattern, (result, builder) -> replacement.apply(builder), fn));
```

api/src/main/java/net/kyori/adventure/text/event/HoverEvent.java
```diff
@@ -225,6 +225,7 @@ public ShowItem(final @NonNull Key item, final @NonNegative int count, final @Nu
      * @return a {@code ShowItem}
      */
     public @NonNull ShowItem item(final @NonNull Key item) {
+      if(requireNonNull(item, "time").equals(this.item)) return this;
```
kashike, line 228:
```suggestion
      if(requireNonNull(item).equals(this.item)) return this;
```

PR #145, opened 2020-08-31
Use TextComponent instead of Component in LegacyComponentSerializer

alexstaeding:
e120cd9f4e4f675b8880a8cabb996ce67b7de96e changed the output component type of `LegacyComponentSerializer` from `TextComponent` to `Component` because of restrictions caused by the addition of the replace methods in Component.

However, as the `LegacyComponentSerializer` exclusively deals with components that are definitely text, it makes sense to use `TextComponent` instead of `Component` (as it used to be).

This PR adds the recently edited replace methods (in `Component`) to `ScopedComponent` to alleviate the restriction and allow `LegacyComponentSerializer` to deal with TextComponents instead of Components again.

### DIFF
--- a/api/src/main/java/net/kyori/adventure/text/ScopedComponent.java
+++ b/api/src/main/java/net/kyori/adventure/text/ScopedComponent.java
@@ -27,6 +27,8 @@ import java.util.ArrayList;
 import java.util.List;
 import java.util.Set;
 import java.util.function.Consumer;
+import java.util.function.UnaryOperator;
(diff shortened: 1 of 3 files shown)


zml2008:
1. those casts aren't valid
2. again, why

alexstaeding:
Are those casts not just as valid as the other casts in `ScopedComponent`?

zml2008:
replace doesn't return self, as I said in the velocity discord it returns `self | TextComponent`

PR #151, opened 2020-09-10
api: add audience that wraps a collection

### LINKED ISSUE #116: Mutable form of ForwardingAudience
Something along the lines of sponge api 7's [MutableMessageChannel](https://github.com/SpongePowered/SpongeAPI/blob/stable-7/src/main/java/org/spongepowered/api/text/channel/MutableMessageChannel.java), which would simply lets the user add/remove audiences at will.

kashike:
fixes #116

### DIFF
--- a/api/src/main/java/net/kyori/adventure/audience/MutableForwardingAudience.java
+++ b/api/src/main/java/net/kyori/adventure/audience/MutableForwardingAudience.java
@@ -1,0 +1,73 @@
+/*
+ * This file is part of adventure, licensed under the MIT License.
+ *
+ * Copyright (c) 2017-2020 KyoriPowered
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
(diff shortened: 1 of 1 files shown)


api/src/main/java/net/kyori/adventure/audience/MutableForwardingAudience.java
```diff
@@ -0,0 +1,73 @@
...
+  }
+
+  /**
+   * Creates.
+   *
+   * @param audiences the collection
+   * @since 4.0.0
+   */
+  public MutableForwardingAudience(final @NonNull Collection<Audience> audiences) {
+    this.audiences = audiences;
```
Cybermaxke, line 66: `null` check?

api/src/main/java/net/kyori/adventure/audience/MutableForwardingAudience.java
```diff
@@ -0,0 +1,73 @@
...
+  private final Collection<Audience> audiences;
+
+  /**
+   * Creates an audience, using an {@link ArrayList} as the backing collection.
+   *
+   * @return an audience
+   * @since 4.0.0
+   */
+  public static @NonNull MutableForwardingAudience arrayList() {
+    return new MutableForwardingAudience(new ArrayList<>());
```
Electroid, line 46: Should default `Collection`s be thread-safe?

kashike:
I am going to close this for now. If anybody would still like this, please comment on #116.

PR #162, opened 2020-09-28
Make fallback locale configurable

lucko:


api/src/main/java/net/kyori/adventure/translation/TranslationRegistryImpl.java
```diff
@@ -36,8 +37,23 @@
 import static java.util.Objects.requireNonNull;
 
 final class TranslationRegistryImpl implements Examinable, TranslationRegistry {
+  private static final Supplier<Locale> SYSTEM_DEFAULT_LOCALE;
+  static {
+    final String property = System.getProperty("net.kyori.adventure.defaultLocale");
```
kashike, line 42:
```suggestion
    final String property = System.getProperty("net.kyo".concat("ri.adventure.defaultLocale");
```

kashike (changes requested):

Electroid:
I don't think this will work with a global translator -- if 1 plugin changes the default, it's likely to break another plugin that depends on another default.

lucko:
@Electroid Yep, I agree - I made another PR #167 to resolve this.

kashike:
#179

PR #165, opened 2020-09-30
Add Non-Regex, Non-Lambda overloads for replace methods

nathanfranke:
Example usage:
---

Before:
```java
Component message = Component.text("Your nickname is now {}.", NamedTextColor.YELLOW)
        .replaceFirstText(Pattern.compile("{}", Pattern.LITERAL), old -> Component.text(nickname, NamedTextColor.GREEN).toBuilder());
```

After:
```java
Component message = Component.text("Your nickname is now {}.", NamedTextColor.YELLOW)
        .replaceFirstText("{}", Component.text(nickname, NamedTextColor.GREEN));
```

### DIFF
--- a/api/src/main/java/net/kyori/adventure/text/Component.java
+++ b/api/src/main/java/net/kyori/adventure/text/Component.java
@@ -1469,6 +1469,18 @@ public interface Component extends ComponentBuilderApplicable, ComponentLike, Ho
   }
 
   /**
+   * Finds and replaces text within any {@link Component}s using a string literal.
+   *
+   * @param search a string literal
+   * @param replacement a {@link ComponentLike} to replace each match
+   * @return a modified copy of this component
+   * @since 4.0.0
+   */
+  default @NonNull Component replaceText(final @NonNull String search, final @Nullable ComponentLike replacement) {
+    return this.replaceText(Pattern.compile(search, Pattern.LITERAL), old -> replacement);
+  }
+
+  /**
    * Finds and replaces text within any {@link TextComponent}s using a regex pattern.
    *
    * @param pattern a regex pattern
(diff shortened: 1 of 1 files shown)


zml2008:
This looks good -- but looks like you accidentally got kashike's commit into your history. Try doing an interactive rebase onto master to strip that out.

nathanfranke:
Fixed. I actually did that on purpose because this PR depended on kashike's

api/src/main/java/net/kyori/adventure/text/Component.java
```diff
@@ -1468,6 +1468,18 @@ default boolean hasStyling() {
     return !this.style().isEmpty();
   }
 
+  /**
+   * Finds and replaces text within any {@link Component}s using a string literal.
+   *
+   * @param old a string literal
```
kashike, line 1474: Rename this parameter from `old` to `search`.

api/src/main/java/net/kyori/adventure/text/Component.java
```diff
@@ -1480,6 +1492,18 @@ default boolean hasStyling() {
     return this.replaceText(pattern, replacement, (index, replaced) -> PatternReplacementResult.REPLACE);
   }
 
+  /**
+   * Finds and replaces the first occurrence of text within any {@link Component}s using a string literal.
+   *
+   * @param old a string literal
```
kashike, line 1498: Rename this parameter from `old` to `search`.

kashike (changes requested):

nathanfranke:
I added a new overload for every existing `Pattern` overload.

Only problem is this code is repeated twice. `(index, replaced) -> replaced < numberOfReplacements ? PatternReplacementResult.REPLACE : PatternReplacementResult.STOP`

Tried to make a method but private methods aren't supported in interfaces until Java 9.

kashike (approved):

api/src/main/java/net/kyori/adventure/text/Component.java
```diff
@@ -1505,6 +1542,19 @@ default boolean hasStyling() {
     return this.replaceText(pattern, replacement, (index, replaced) -> replaced < numberOfReplacements ? PatternReplacementResult.REPLACE : PatternReplacementResult.STOP);
   }
 
+  /**
+   * Finds and replaces {@code n} instances of text within any {@link TextComponent}s using a string literal.
```
zml2008, line 1546: this should be updated to mention `fn`

kashike (approved):

PR #215, opened 2020-11-27
api: TextColor.lerp

kashike:


MiniDigger (approved):

api/src/main/java/net/kyori/adventure/text/format/TextColor.java
```diff
@@ -188,6 +188,35 @@
...
 
+  /**
+   * Linearly interpolates between {@code this} and {@code that} by {@code t}.
+   *
+   * @param t interpolation value
+   * @param that upper bound
+   * @return the interpolated value
+   * @since 4.3.0
+   */
+  default @NonNull TextColor lerp(final float t, final @NonNull TextColor that) {
```
kashike, line 199: Unsure about this non-`static` method personally.

api/src/main/java/net/kyori/adventure/text/format/TextColor.java
```diff
@@ -188,6 +188,35 @@
     return this.value() & 0xff;
   }
 
+  /**
+   * Linearly interpolates between {@code this} and {@code that} by {@code t}.
```
Electroid, line 192: Can there be a `<p>` that explains what this does in layman's terms? I don't think the average user of adventure has a solid grasp of linear interpolation with respect to colors.

e.g. https://p5js.org/reference/#/p5/lerpColor

api/src/main/java/net/kyori/adventure/text/format/TextColor.java
```diff
@@ -188,6 +188,35 @@
...
+   *
+   * @param t interpolation value
```
Electroid, line 194: Mention this value is [0, 1]

Electroid: Also, maybe another name than `t` ?

api/src/main/java/net/kyori/adventure/text/format/TextColor.java
```diff
@@ -188,6 +188,35 @@
...
+   * @param a lower bound
+   * @param b upper bound
+   * @return the interpolated color
+   * @since 4.3.0
+   */
+  static @NonNull TextColor lerp(final float t, final @NonNull TextColor a, final @NonNull TextColor b) {
+    return color(
+      Math.round(a.red() + t * (b.red() - a.red())),
```
Electroid, line 214: Should we clamp `t` from [0,1] or throw an error if out of bounds?

Electroid (changes requested):

api/src/main/java/net/kyori/adventure/text/format/TextColor.java
zml2008: `step`? `fraction`?

api/src/main/java/net/kyori/adventure/text/format/TextColor.java
zml2008: I think the non-static method should be the primary one here -- we own `TextColor`, so it makes sense to put methods directly on it.

zml2008:
this should be updated at some point to support lerping HSV colours as well.

api/src/test/java/net/kyori/adventure/text/format/TextColorTest.java
```diff
@@ -97,7 +97,13 @@ void testExtractComponents() {
...
 
   @Test
-  void testEquality() {
+  void testLerp() {
+    assertEquals(TextColor.color(0x808080), TextColor.lerp(0.50f, TextColor.color(0xffffff), TextColor.color(0x000000)));
+    assertEquals(TextColor.color(0x3399FF), TextColor.lerp(0.50f, TextColor.color(0x3366FF), TextColor.color(0x33CCFF)));
+  }
+
+  @Test
+  public void testEquality() {
```
kashike, line 106:
```suggestion
  void testEquality() {
```

kezz (approved):
Could be worth adding some more tests (perhaps for `t=0` and `t=1`) but it's just simple maths so might not be worth the effort. Looks great though!

Issue #227, opened 2020-12-07
Make use of `Contract` annotation from JetBrains java-annotations?


kashike:
https://github.com/JetBrains/java-annotations/blob/master/common/src/main/java/org/jetbrains/annotations/Contract.java

PR #228, opened 2020-12-08
Add JetBrains Contract annotations to some methods

lucko:
#227


api/src/main/java/net/kyori/adventure/bossbar/BossBar.java
```diff
@@ -251,7 +260,7 @@ default float percent() {
    * @return the flags
    * @since 4.0.0
    */
-  @NonNull Set<Flag> flags();
+  @UnmodifiableView @NonNull Set<Flag> flags();
```
kashike, line 263: Fairly sure the `Set` returned here is mutable?

zml2008: luck changed that as part of this PR. Modifying flags wouldn't have worked in the first place, since the listener wouldn't have been called.

kashike: Makes sense.

kashike (approved):

PR #230, opened 2020-12-10
api: Add 'Builder' suffix to Component factory methods that return a builder

### LINKED ISSUE #222: Make static methods that create objects distinct from those that create their builders.
_This issue is taken from https://github.com/SpongePowered/SpongeAPI/pull/2275#issuecomment-739912355_

One large pain point that I've encountered is that there are factory methods that are named the same as their builder methods. Coupled with the convenience methods on the immutable variants, the method that should be used to create an object is extremely unclear (another issue will be opened on the convenience methods).

Take the following two methods from your library in isolation - what someone booting an IDE and typing `Component.text(` might see.

```java
Component.text();
Component.text(String);
```

They both look like factory methods to me: my assumption is that they would both create `Component`s, one empty, one containing the string. However, it turns out that the first is a `TextComponent.Builder`, and one is a `TextComponent`. What's bad, however, is that my assumption can actually hold - if I use these as `ComponentLike`s in something that accepts `ComponentLike`s, then I will indeed end up with `Component.empty()` (a method that's actually okay but totally eclipsed by this) and my text component. So now we have two methods that don't do the same thing, but can be treated as such, but are actually totally different ways to do the same thing.

While both can actually be used in the same way, with the same methods names, they really shouldn't be. In theory, I can do the following:

```java
Component.text().content("Text").color(NamedTextColor.GREEN).style(Style.style().decorate(TextDecoration.BOLD).decorate(TextDecoration.ITALIC).build());
Component.text("").content("Text").color(NamedTextColor.GREEN).decorate(TextDecoration.BOLD).style(Style.style(TextDecoration.BOLD).decoration(TextDecoration.ITALIC, TextDecoration.State.TRUE)); // slighly facetious with the starting method, but it proves a point
```

and you get the same result - one uses the builders, one does not - and because both the builder and the standard component implement `ComponentLike`, I can use these within other calls and would be no wiser to the fact they are two totally different objects.

I propose the following:

* Deprecate any static methods that return builders and are named the same as factory static methods that _do not_ return builders.
* Create replacement methods suffixed with `Builder` to clearly mark that those methods return a different type.

zml2008:
Fixes #222, see there for discussion on the concept

### DIFF
--- a/api/src/main/java/net/kyori/adventure/text/Component.java
+++ b/api/src/main/java/net/kyori/adventure/text/Component.java
@@ -135,10 +135,23 @@ public interface Component extends ComponentBuilderApplicable, ComponentLike, Ex
    * Creates a block NBT component builder.
    *
    * @return a builder
-   * @since 4.0.0
+   * @since 4.4.0
    */
-  static BlockNBTComponent.@NonNull Builder blockNBT() {
+  static BlockNBTComponent.@NonNull Builder blockNBTBuilder() {
     return new BlockNBTComponentImpl.BuilderImpl();
+  }
+
+
+  /**
+   * Creates a block NBT component builder.
+   *
+   * @return a builder
+   * @since 4.0.0
+   * @deprecated for removal since 4.4.0, see {@link #blockNBTBuilder()}
+   */
+  @Deprecated
+  static BlockNBTComponent.@NonNull Builder blockNBT() {
+    return blockNBTBuilder();
   }
 
   /**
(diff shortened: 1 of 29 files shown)


api/src/main/java/net/kyori/adventure/text/Component.java
```diff
@@ -131,14 +131,27 @@
...
+    return new BlockNBTComponentImpl.BuilderImpl();
+  }
+
+
   /**
    * Creates a block NBT component builder.
    *
    * @return a builder
    * @since 4.0.0
+   * @deprecated for removal since 4.4.0, see {@link #blockNBTBuilder()}
```
kashike, line 150: Or this?
```suggestion
   * @deprecated for removal since 4.4.0, in favour of {@link #blockNBTBuilder()}
```

zml2008: i was just trying to match existing deprecations :p

kashike (approved):

lucko (changes requested):
I've put my comments re: this change on the original issue: https://github.com/KyoriPowered/adventure/issues/222#issuecomment-742416165

I'm against merging this as-is.

zml2008:
Due to opposition and several proposed alternatives, I'm going to close this PR. If the issue is ever revisited, any other approach would effectively be starting from scratch anyways.

PR #240, opened 2020-12-14
nbt: ability to add multiple tags to a Compound tag

kashike:


nbt/src/main/java/net/kyori/adventure/nbt/CompoundBinaryTag.java
```diff
@@ -44,6 +44,19 @@
...
 
+  /**
+   * Creates a compound tag populated with {@code tags}.
+   *
+   * <p>If {@code tags} is empty, {@link #empty()} will be returned.</p>
+   *
+   * @return a compound tag
+   * @since 4.4.0
+   */
+  static @NonNull CompoundBinaryTag from(final @NonNull Map<String, BinaryTag> tags) {
```
SirYwell, line 55: Is there a reason for using `Map<String, BinaryTag>` instead of `Map<String, ? extends BinaryTag>`?

PR #243, opened 2020-12-14
nbt: Compound removal methods

kashike:


nbt/src/main/java/net/kyori/adventure/nbt/CompoundTagSetter.java
```diff
@@ -42,6 +44,27 @@
...
+
+  /**
+   * Removes a tag.
+   *
+   * @param key the key
+   * @param removed a consumer that accepts the removed tag
+   * @return a compound tag
+   * @since 4.0.0
+   */
+  @NonNull R remove(final @NonNull String key, final @Nullable Consumer<BinaryTag> removed);
```
octylFractal, line 66: Consider making this `@NonNull`, and delegating with `this.remove(key, r -> {});` instead since the lambda should effectively be a constant, and there would be no branch.

kashike: Is `@Nullable` so that people don't need to do something like:
```java
if(result != null) {
  compound.remove("foo", result);
} else {
  compound.remove("foo");
}
```
etc

zml2008 (changes requested):
mostly good, just going to have JD opinions here :)

nbt/src/main/java/net/kyori/adventure/nbt/CompoundTagSetter.java
kashike: The [same method](https://github.com/KyoriPowered/adventure/blob/master/nbt/src/main/java/net/kyori/adventure/nbt/ListBinaryTag.java#L147) in `ListBinaryTag` accepts a `@Nullable` `Consumer`, so we will do the same here for consistency.

nbt/src/main/java/net/kyori/adventure/nbt/CompoundTagSetter.java
```diff
@@ -52,6 +54,27 @@
    */
   @NonNull R put(final @NonNull Map<String, ? extends BinaryTag> tags);
 
+  /**
+   * Removes a tag.
+   *
+   * @param key the key
+   * @return a compound tag
```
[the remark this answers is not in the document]
kashike: I changed the method to return the same tag if `key` is not in it, so this has not been changed.

zml2008 (approved):

Issue #246, opened 2020-12-15
nbt: ListBinaryTag is missing getList


octylFractal:
Basically what the title says, there's no `#getList()` helper like there is for `CompoundBinaryTag`. Maybe it's just not that common, but I think it can't hurt to be consistent.

PR #247, opened 2020-12-15
nbt: ListBinaryTag.getList

kashike:
fixes GH-246

### DIFF
--- a/nbt/src/main/java/net/kyori/adventure/nbt/ListBinaryTag.java
+++ b/nbt/src/main/java/net/kyori/adventure/nbt/ListBinaryTag.java
@@ -367,6 +367,63 @@ public interface ListBinaryTag extends ListTagSetter<ListBinaryTag, BinaryTag>, 
   }
 
   /**
+   * Gets a list.
+   *
+   * @param index the index
+   * @return the list, or an empty list if the tag at index {@code index} is not a list tag
+   * @since 4.4.0
+   */
+  default @NonNull ListBinaryTag getList(final @NonNegative int index) {
+    return this.getList(index, null, ListBinaryTag.empty());
+  }
+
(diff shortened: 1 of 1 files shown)


nbt/src/main/java/net/kyori/adventure/nbt/ListBinaryTag.java
```diff
@@ -366,6 +366,62 @@ default double getDouble(final @NonNegative int index, final double defaultValue
     return defaultValue;
   }
 
+  /**
+   * Gets a list.
+   *
+   * @param index the index
+   * @return the list, or an empty list
```
zml2008, line 373: when is the empty list returned? when `index` is out of bounds? or when the tag at `index` is not a list tag? if it's the second, what happens when `index` is out of bounds?

nbt/src/main/java/net/kyori/adventure/nbt/ListBinaryTag.java
```diff
@@ -366,6 +366,62 @@ default double getDouble(final @NonNegative int index, final double defaultValue
...
+    return this.getList(index, null, defaultValue);
+  }
+
+  /**
+   * Gets a list.
+   *
+   * <p>If {@code elementType} is non-{@code null} and the {@link ListBinaryTag} at index {@code index} does not match, {@code defaultValue} will be returned.</p>
+   *
+   * @param index the index
+   * @param elementType the element type of the list
```
zml2008, line 409: since there are two lists in context here, i'd like to see the JD be super explicit about which list is relevant for any operation or parameter.

zml2008 (changes requested):
I think the API here is a good addition, would like some more clarity in docs though :)

PR #248, opened 2020-12-15
nbt: ListBinaryTag.add(Iterable)

kashike:


nbt/src/main/java/net/kyori/adventure/nbt/ListBinaryTagImpl.java
```diff
@@ -64,34 +64,42 @@ public int size() {
...
+    noAddEnd(tag);
+    if(this.elementType != BinaryTagTypes.END) {
+      mustBeSameType(tag, this.elementType);
+    }
+    return this.edit(tags -> tags.add(tag), tag.type());
+  }
+
+  @Override
+  public @NonNull ListBinaryTag add(final Iterable<? extends BinaryTag> tagsToAdd) {
+    final BinaryTagType<?> type = ListBinaryTagImpl.mustBeSameType(tagsToAdd);
```
zml2008, line 97: add a check here that `tagsToAdd` is non-empty maybe, and if it is empty return `this` to save the copy?

nbt/src/main/java/net/kyori/adventure/nbt/ListBinaryTag.java
```diff
@@ -46,6 +46,16 @@
     return ListBinaryTagImpl.EMPTY;
   }
 
+  /**
+   * Gets an empty list tag.
```
zml2008, line 50: you forgot to update this :p

nbt/src/main/java/net/kyori/adventure/nbt/ListTagSetter.java
```diff
@@ -41,4 +41,13 @@
    * @since 4.0.0
    */
   @NonNull R add(final T tag);
+
+  /**
+   * Adds tags.
```
zml2008, line 46:
```suggestion
* Adds multiple tags
```

zml2008 (changes requested):

nbt/src/main/java/net/kyori/adventure/nbt/ListBinaryTagImpl.java
kashike: `Iterable` has no `isEmpty` or `size` method?

zml2008: you could either create an Iterator and check hasNext before going into `edit()`, or only do the optimization for inputs that are also a `Collection` i guess

PR #249, opened 2020-12-15
 nbt: Introduce a new way to read and write binary tags, with support for (legacy) named roots.

kashike:


zml2008 (changes requested):
I don't really like the duplication here -- NamedBinaryTagIO is the exact same as BinaryTagIO, but with customizable name field.

What do you think of splitting up the input/output stream choice from the choice of how the root tag is handled, with something like Configurate's ConfigurationLoader/AbstractConfigurationLoader.Builder, something similar to what the SNBT reader + writer do for options, or something else?

This would also open the door to a configurable serializer if we add options like restricting the allowed tags (if someone wanted to work with really old MC), or supporting Bedrock's format.

kashike:
> What do you think of splitting up the input/output stream choice from the choice of how the root tag is handled, with something like Configurate's ConfigurationLoader/AbstractConfigurationLoader.Builder, something similar to what the SNBT reader + writer do for options, or something else?

That sounds reasonable to me, though I am not sure what you have in mind for such?

zml2008 (approved):

nbt/src/main/java/net/kyori/adventure/nbt/BinaryTagIO.java
```diff
@@ -182,11 +200,283 @@ public static void writeCompressedOutputStream(final @NonNull CompoundBinaryTag
...
+        return new GZIPInputStream(is);
+      }
+
+      @Override
+      @NonNull OutputStream compress(final @NonNull OutputStream os) throws IOException {
+        return new GZIPOutputStream(os);
+      }
+    };
+
+    abstract @NonNull InputStream decompress(final @NonNull InputStream is) throws IOException;
```
zml2008, line 478: these methods should have JD, looks like.

i assume your intent here is others could implement Compression to write NBT with some other format, rather than gzip?

kashike: These two methods are package-private

zml2008: ah -- i've spent too much time looking at interfaces i gues :)

nbt/src/main/java/net/kyori/adventure/nbt/BinaryTagIO.java
```diff
@@ -182,11 +200,283 @@ public static void writeCompressedOutputStream(final @NonNull CompoundBinaryTag
...
+  }
+
+  /**
+   * {@link CompoundBinaryTag} reader.
+   *
+   * @since 4.4.0
+   */
+  public interface Reader {
+    /**
+     * Reads a binary tag from {@code path}.
```
octylFractal, line 218: Should specify that this is equivalent to `read(path, null)`.

nbt/src/main/java/net/kyori/adventure/nbt/BinaryTagWriterImpl.java
```diff
@@ -0,0 +1,92 @@
...
+    }
+  }
+
+  @Override
+  public void writeNamed(final Map.@NonNull Entry<String, CompoundBinaryTag> tag, @NonNull OutputStream output, final BinaryTagIO.@Nullable Compression compression) throws IOException {
+    output = closeShield(output);
+    if(compression != null) {
+      output = compression.compress(output);
+    }
+    try(final DataOutputStream dos = new DataOutputStream(new BufferedOutputStream(new GZIPOutputStream(closeShield(output))))) {
```
octylFractal, line 81: bug here with `GZIPOutputStream` always applied, you might want to add some tests to cover this :P

octylFractal (changes requested):
Personally I would rather have a `Compression.NONE` rather than `@Nullable Compression`, but that's just me.

PR #252, opened 2020-12-25
Include empty components in translatable arguments

Electroid:
Consider the following scenario: 
```java
MessageFormat format = new MessageFormat("{1} was killed by {2}"); // key = player.death

Component message = translatable()
  .key("player.death")
  .args(empty(), text("Bob"), text("Alice"))
  .build()
```

If you translate `message` today, you'll get `"Alice was killed by {2}"`. This is because `AbstractComponent#asComponents` filters out `Component#empty`. While this works in most situations, it causes index-skew with `TranslatableComponent`.


Electroid:
Also worth noting, this is a regression from `3.x`

zml2008 (approved):

kashike (approved):

PR #257, opened 2021-01-11
serializer-gson: Change output ordering to match the order of Vanilla components

### LINKED ISSUE #256: JSON component serialization order issue
The order in which adventure serializes components is out of order compared to vanilla, vanilla logic
(Paper IChatBaseComponent#serialize#L258, stores NBT in the order of

- chat modifiers
- extra
- component data

Adventure however stores the items in the order
- component data
- extra
- chat modifiers

while this is fine for component comparison given that this relies on the actual data and not the order, this fails inside of vanilla for aspects like item comparisons as those are done by comparing the NBT, which stores the json string

zml2008:
Fixes #256

### DIFF
--- a/text-serializer-gson/src/main/java/net/kyori/adventure/text/serializer/gson/StyleSerializer.java
+++ b/text-serializer-gson/src/main/java/net/kyori/adventure/text/serializer/gson/StyleSerializer.java
@@ -38,6 +38,8 @@ import com.google.gson.stream.JsonReader;
 import java.io.IOException;
 import java.io.StringReader;
 import java.lang.reflect.Type;
+import java.util.EnumSet;
+import java.util.Set;
 import net.kyori.adventure.key.Key;
 import net.kyori.adventure.text.Component;
 import net.kyori.adventure.text.event.ClickEvent;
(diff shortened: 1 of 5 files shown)


text-serializer-gson/src/main/java/net/kyori/adventure/text/serializer/gson/StyleSerializer.java
```diff
@@ -49,7 +51,27 @@
...
-  private static final TextDecoration[] DECORATIONS = TextDecoration.values();
+  @SuppressWarnings("checkstyle:NoWhitespaceAfter")
+  private static final TextDecoration[] DECORATIONS = {
+    // The order here is important -- MC does string comparisons of some serialized components
+    // so we have to make sure our order matches Vanilla
+    TextDecoration.BOLD,
+    TextDecoration.ITALIC,
+    TextDecoration.UNDERLINED,
+    TextDecoration.STRIKETHROUGH,
+    TextDecoration.OBFUSCATED,
```
kashike, line 62: remove , at end

kashike (changes requested):

kashike (approved):

PR #258, opened 2021-01-11
api: Rename LinearComponents.linear to Component.compose

### LINKED ISSUE #225: Improve documentation and discoverabilty of LinearComponents.linear()
_This issue is taken from https://github.com/SpongePowered/SpongeAPI/pull/2275#issuecomment-739912355_

`LinearComponents.linear()` is clearly designed to ease the burden of creating a tree like structure, but is hindered by the fact that it's not obvious what it is, why it's there, how it works. The documentation says this:

https://github.com/KyoriPowered/adventure/blob/3af760f73830902f1a546f905ef1565814d9bc50/api/src/main/java/net/kyori/adventure/text/LinearComponents.java#L39-L45

which tells me nothing about it and devalues the existence of the function. Until recently, I was under the impression that it was mostly to be used to effectively join multiple components together - however, based on recent conversations, it actually aspires that it can be used for more than that - like including colours and styles. I have no idea how that works - does it work like Sponge API 7's `Text.of()` when it comes to colours and styles?

It's also not a particularly descriptive name: to most people it's not going to be obvious that "linear" components means the same as this `Text.of(...)` - people I know only know of it because I told them about it (and even then, I should just use a `join` method)! It _might_ be better on the `Component` interface - but it certainly needs fleshing out in terms of documentation and what it can do.

zml2008:
Closes #225


api/src/main/java/net/kyori/adventure/text/Component.java
```diff
@@ -157,6 +158,58 @@
...
+  /**
+   * Compose a series of elements (styles and contents) into a component.
+   *
+   * <p>Styles apply to all components after them until a conflicting style is discovered</p>
+   *   <pre>
+   *     Component message = LinearComponents.linear(NamedTextColor.RED, translatable("welcome.message"), TextDecoration.BOLD, text(" SERVER));
+   *   </pre>
+   * In this example all the text is red, but only the last word is bold.
+   *   <pre>
+   *     Component message = LinearComponents.linear(NamedTextColor.GREEN, text("I am green. "), NamedTextColor.GRAY, text("I am gray."));
```
KingOfSquares, line 170: `LinearComponents.linear` should be switched with the new method

api/src/main/java/net/kyori/adventure/text/AbstractComponent.java
```diff
@@ -73,6 +73,10 @@
     return newList;
   }
 
+  static IllegalStateException nothingComponentLike() {
+    return new IllegalStateException("Cannot build component linearly - nothing component-like was given");
```
kashike, line 77: Should either re-word to drop "linearly", or rename method.

kashike (changes requested):

lucko (changes requested):
Overall I agree with this change :)

However, I'm not a huge fan of the name `compose`. (I raised this in our dev chat but figured I should say here too!)

In my mind, compose is more akin to `Function#compose` / functional composition than joining/concatenating text elements together.

I would prefer `join` - but it was noted that this would conflict with the existing method. I personally don't think that's a huge issue (the methods have a similar behaviour) but I also suggested `joinWithInlineStyle` or just `joinInline` in response.

Interested to hear if anyone else agrees (happy ofc to settle for the current proposal if it's just me 😆 ) or if there's an even better name I may be overlooking. :)

PR #259, opened 2021-01-13
API addition for translations

PrimordialMoros:
Adds a method to check if a translation key has any registered translations.

This allows one to easily handle a missing translation key using a custom fallback.

### DIFF
--- a/api/src/main/java/net/kyori/adventure/translation/TranslationRegistry.java
+++ b/api/src/main/java/net/kyori/adventure/translation/TranslationRegistry.java
@@ -72,6 +72,15 @@ public interface TranslationRegistry extends Translator {
   }
 
   /**
+   * Checks if any translations are explicitly registered for the specified key.
+   *
+   * @param key a translation key
+   * @return whether the registry contains a value for the translation key
+   * @since 4.7.0
+   */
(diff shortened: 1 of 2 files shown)


zml2008:
It's not possible to know serverside whether or not a translation exists - the translatable component supports any key defined in a client resource pack as well.

PrimordialMoros:
I was thinking about plugin registered keys that are unlikely to exist on client resource packs. Perhaps, should mention this limitation in the documentation, or change the method name to better reflect that?

api/src/main/java/net/kyori/adventure/translation/TranslationRegistry.java
```diff
@@ -71,6 +71,15 @@
...
   }
 
+  /**
+   * Checks if any translations have been registered for the specified key.
+   *
+   * @param key a translation key
+   * @return whether the key has any registered translations
+   * @since 4.3.1
+   */
+  boolean hasTranslation(final @NonNull String key);
```
kashike, line 81: `contains`?

PrimordialMoros: Updated it

zml2008 (approved):

Issue #135, opened 2020-08-15
Legacy serializer should match client behavior


octylFractal:
Currently the legacy serializer skips translatable components, rather than rendering the key like the vanilla client. I request that this be changed to match the vanilla client, so that WorldEdit's abuse of this feature renders properly when converted to legacy formatting :P

PR #262, opened 2021-01-21
api: add support for pixel width calculation

### LINKED ISSUE #59: Pixel width
The ability to find the pixel-width of a single character, or the content of an entire component (and children).
```java
public interface PixelWidthSource {
  int width(final @NonNull Component component);

  int width(final char character);
}
```

KingOfSquares:
This PR adds support for pixel width calculation in adventure. (Closes #59)

I've provided a function able to calculate width for the default Minecraft font, but a custom function could also be used.

I have a few problems about my design which I hope I will be able to solve with some feedback:

1. The use of the term "resolve"
I opted to call the small functions users can insert into a source to turn components into text `ComponentResolver`s, trying to avoid "render" since I felt that belonged to the renderers, which has different behavior and standards. I think "resolve" works ok, but unsure if it's the best word to use.
2. How to distribute a source
Right now sources has to be made using one of the static methods in `PixelWidthSource`, which means users gets the responsibility of distributing the source internally. Generally adventure seems to stray away from this (Except for the platform audience creation), and I'm not sure how to solve the problem if it needs to be solved. The context necessary for calculation is hard to generalize further than it currently is.

Any additional feedback is also appreciated! 😎 👍

### DIFF
--- a/api/src/main/java/net/kyori/adventure/text/width/CharacterWidthFunction.java
+++ b/api/src/main/java/net/kyori/adventure/text/width/CharacterWidthFunction.java
@@ -1,0 +1,43 @@
+/*
+ * This file is part of adventure, licensed under the MIT License.
+ *
+ * Copyright (c) 2017-2021 KyoriPowered
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in all
+ * copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+ * SOFTWARE.
+ */
+package net.kyori.adventure.text.width;
+
+import net.kyori.adventure.text.format.Style;
+
+/**
+ * A function that takes a character(represented by its UTF-16 codepoint) and a {@link Style} and returns
+ * the characters width as an {@code int}.
+ *
+ * @since 4.7.0
+ */
+@FunctionalInterface
+public interface CharacterWidthFunction {
+  /**
+   * Gets the width for the given character(represented by its UTF-16 codepoint). {@code char}s will
+   * automatically be converted to codepoints.
+   *
+   * @since 4.7.0
+   */
+  int widthOf(int codepoint, Style style);
+}
--- a/api/src/main/java/net/kyori/adventure/text/width/PixelWidthSource.java
+++ b/api/src/main/java/net/kyori/adventure/text/width/PixelWidthSource.java
@@ -40,7 +40,7 @@ import org.jetbrains.annotations.Nullable;
  * <p>Providing context to this source is only necessary if a custom {@code Function<CX, CharacterWidthFunction>} is provided on creation</p>
  *
  * @param <CX> a context type (player, server, locale)
- * @since 4.7.0
+ * @since 4.10.0
  */
 @ApiStatus.NonExtendable
 public interface PixelWidthSource<CX> extends Buildable<PixelWidthSource<CX>, PixelWidthSource.Builder<CX>> {
@@ -48,13 +48,15 @@ public interface PixelWidthSource<CX> extends Buildable<PixelWidthSource<CX>, Pi
   /**
    * A pixel width source calculating width using the provided flattener and character width function.
    *
-   * <p>A null value results in using the basic counterpart</p>
+   * <p>A null value results in using a basic counterpart:</p>
+   * <p>{@link net.kyori.adventure.text.flattener.ComponentFlattener#basic()}</p>
+   * <p>{@link net.kyori.adventure.text.width.DefaultCharacterWidthFunction#INSTANCE}</p>
    *
    * @param flattener a flattener used to turn components into linear text
    * @param function a function that provides a character width function
    * @param <CX> context a context type (player, server, locale)
    * @return a pixel width source
-   * @since 4.7.0
+   * @since 4.10.0
    */
   static <CX> @NotNull PixelWidthSource<CX> pixelWidthSource(final @Nullable ComponentFlattener flattener, final @Nullable Function<@Nullable CX, CharacterWidthFunction> function) {
     return new PixelWidthSourceImpl<>(flattener == null ? ComponentFlattener.basic() : flattener, function == null ? cx -> DefaultCharacterWidthFunction.INSTANCE : function);
@@ -66,7 +68,7 @@ public interface PixelWidthSource<CX> extends Buildable<PixelWidthSource<CX>, Pi
    * @param component a component
    * @param context the context of this calculation
    * @return the pixel width of the component
-   * @since 4.7.0
+   * @since 4.10.0
    */
   float width(final @NotNull Component component, final @Nullable CX context);
 
@@ -75,7 +77,7 @@ public interface PixelWidthSource<CX> extends Buildable<PixelWidthSource<CX>, Pi
    *
    * @param component a component
    * @return the pixel width of the component
-   * @since 4.7.0
+   * @since 4.10.0
    */
   default float width(final @NotNull Component component) {
     return this.width(component, null);
@@ -88,7 +90,7 @@ public interface PixelWidthSource<CX> extends Buildable<PixelWidthSource<CX>, Pi
    * @param style the style of the string
    * @param context the context of this calculation
    * @return the pixel width of the string
-   * @since 4.7.0
+   * @since 4.10.0
    */
   float width(final @NotNull String string, final @NotNull Style style, final @Nullable CX context);
 
@@ -98,7 +100,7 @@ public interface PixelWidthSource<CX> extends Buildable<PixelWidthSource<CX>, Pi
    * @param string a string
    * @param style the style of the string
    * @return the pixel width of the string
-   * @since 4.7.0
+   * @since 4.10.0
    */
   default float width(final @NotNull String string, final @NotNull Style style) {
     return this.width(string, style, null);
@@ -111,7 +113,7 @@ public interface PixelWidthSource<CX> extends Buildable<PixelWidthSource<CX>, Pi
    * @param style the style of the character
    * @param context the context of this calculation
    * @return the pixel width of the character
-   * @since 4.7.0
+   * @since 4.10.0
    */
   float width(final char character, final @NotNull Style style, final @Nullable CX context);
 
@@ -121,7 +123,7 @@ public interface PixelWidthSource<CX> extends Buildable<PixelWidthSource<CX>, Pi
    * @param character a character
    * @param style the style of the character
    * @return the pixel width of the character
-   * @since 4.7.0
+   * @since 4.10.0
    */
   default float width(final char character, final @NotNull Style style) {
     return this.width(character, style, null);
@@ -134,7 +136,7 @@ public interface PixelWidthSource<CX> extends Buildable<PixelWidthSource<CX>, Pi
    * @param style the style of the character
    * @param context the context of this calculation
    * @return the pixel width of the character
-   * @since 4.7.0
+   * @since 4.10.0
    */
   float width(final int codepoint, final @NotNull Style style, final @Nullable CX context);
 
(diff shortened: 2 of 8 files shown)


api/src/main/java/net/kyori/adventure/text/width/PixelWidthSource.java
```diff
@@ -0,0 +1,138 @@
...
+  /**
+   * Calculates the pixel width of a character, given a context.
+   *
+   * @param c a character
+   * @param style the style of the character
+   * @param context the context of this calculation
+   * @return the pixel width of the character
+   * @since 4.5.0
+   */
+  int width(final char c, final @NonNull Style style, final @NonNull CX context);
```
kashike, line 103:
```suggestion
  int width(final char character, final @NonNull Style style, final @NonNull CX context);
```

api/src/main/java/net/kyori/adventure/text/width/PixelWidthSourceImpl.java
```diff
@@ -0,0 +1,368 @@
...
+
+  @Override
+  public int width(final char c, final @NonNull Style style, final @NonNull CX context) {
+    return this.function.apply(context).applyAsInt(c, style);
+  }
+
+  @SuppressWarnings(value="unchecked")
+  @Override
+  public <CO extends Component> void addResolver(final @NonNull Class<CO> resolveFor, final @NonNull ComponentResolver<CO, CX> resolver) {
+
```
kashike, line 200:
```suggestion
```

api/src/main/java/net/kyori/adventure/text/width/PixelWidthSourceImpl.java
```diff
@@ -0,0 +1,368 @@
...
+        this.entityNBTResolvers.add((ComponentResolver<EntityNBTComponent, CX>) resolver);
+        return;
+      } else if(resolveFor.isAssignableFrom(StorageNBTComponent.class)) {
+        this.storageNBTResolvers.add((ComponentResolver<StorageNBTComponent, CX>) resolver);
+        return;
+      }
+    }
+
+    throw new UnsupportedOperationException("Invalid Component");
+
```
kashike, line 229:
```suggestion
```

api/src/main/java/net/kyori/adventure/util/CharToIntFunction.java
```diff
@@ -0,0 +1,41 @@
...
+package net.kyori.adventure.util;
+
+/**
+ * A function that takes a {@code char} and a value {@code U} as input and produces an {@code int}.
+ *
+ * @since 4.4.0
+ */
+
+@FunctionalInterface
+public interface CharToIntFunction<U> {
```
kashike, line 33: Should probably make this a more specific name, really.

kashike (changes requested):

api/src/main/java/net/kyori/adventure/text/width/PixelWidthSource.java
```diff
@@ -0,0 +1,138 @@
...
+   */
+  <CO extends Component> void addResolver(final @NonNull Class<CO> resolveFor, final @NonNull ComponentResolver<CO, CX> resolver);
+
+  /**
+   * Something that can either resolve a {@link Component} into a {@link TextComponent}, or return {@code null} if not possible.
+   *
+   * @param <CX> a context (player, server, locale)
+   * @since 4.5.0
+   */
+  interface ComponentResolver<CO extends Component, CX> {
```
zml2008, line 122: I think this component resolver system could be broken out to be more broadly applicable for operations that need to work with linear text -- think the legacy (#135) and plain serializers, or selector and NBT components that need server-side resolving (KyoriPowered/adventure-platform#8)

api/src/main/java/net/kyori/adventure/text/width/PixelWidthSourceImpl.java
```diff
@@ -0,0 +1,368 @@
...
+  private final PixelWidthRenderer renderer = new PixelWidthRenderer();
+
+  private final Function<CX, CharToIntFunction<Style>> function;
+  private final Function<CX, Locale> localeFunction;
+
+  /**
+   * Get the pixel width of the given character in the Minecraft font, excluding the space between
+   * characters and drop shadow. Handles alphanumerics and most common english punctuation.
+   */
+  static final CharToIntFunction<Style> DEFAULT_FONT_WIDTH = (c, style) -> {
```
zml2008, line 83: This should be something generated by `adventure-vanilla-data` or otherwise directly by the client, not a magic block of code that's been copied around.

api/src/main/java/net/kyori/adventure/text/width/PixelWidthSourceImpl.java
```diff
@@ -0,0 +1,368 @@
...
+
+    map.forEach((string, style) -> string.chars().forEach(c -> length.getAndAdd(this.width((char) c, style, context))));
+
+    return length.get();
+  }
+
+  @Override
+  public int width(final @NonNull String string, final @NonNull Style style, final @NonNull CX context) {
+    int length = 0;
+    for(final char c : string.toCharArray()) length += this.width(c, style, context);
```
zml2008, line 188: this won't handle code points outside of the BMP properly

api/src/main/java/net/kyori/adventure/text/width/PixelWidthSourceImpl.java
```diff
@@ -0,0 +1,368 @@
...
+  }
+
+  /**
+   * Flattens a {@link TextComponent} and <em>all</em> it's {@link TextComponent} children,
+   * wherever they are, into a map where the content and style of each component is linked together.
+   */
+  private Map<String, Style> flattenComponents(final @NonNull TextComponent parent){
+    //By this point, we know that all components in this component tree are TextComponents,
+    //so some casting is acceptable
+    final Map<String, Style> map = new HashMap<>();
```
zml2008, line 252: `flattenComponents` should take a map that it adds to, rather than creating a new Map at each recursion

api/src/main/java/net/kyori/adventure/text/width/PixelWidthSourceImpl.java
```diff
@@ -0,0 +1,368 @@
...
+
+      if(!(component0 instanceof TextComponent)) component0 = this.renderUsingResolver(PixelWidthSourceImpl.this.translatableResolvers, component, context);
+
+      return component0 != null ? (TextComponent) component0 : text(component.key(), component.style()).children(component.children());
+    }
+
+    //null -> unsuccessful
+    private <CO extends Component> @Nullable TextComponent renderUsingResolver(final @NonNull Set<ComponentResolver<CO, CX>> resolvers, final @NonNull CO component, final @NonNull CX context){
+      final AtomicReference<TextComponent> result = new AtomicReference<>(null);
+      resolvers.forEach(r -> {
```
zml2008, line 348: you should just use a normal loop here

api/src/main/java/net/kyori/adventure/text/width/PixelWidthSourceImpl.java
```diff
@@ -0,0 +1,368 @@
...
+      return rendered != null ? rendered : text(component.keybind(), component0.style()).children(component0.children());
+    }
+
+    @Override
+    protected @NonNull TextComponent renderScore(final @NonNull ScoreComponent component, final @NonNull CX context) {
+      final ScoreComponent component0 = (ScoreComponent) super.renderScore(component, context);
+
+      final TextComponent rendered = this.renderUsingResolver(PixelWidthSourceImpl.this.scoreResolvers, component, context);
+
+      return rendered != null ? rendered : text(component.name() + "_" + component.objective() + "_" + component.value(), component0.style()).children(component0.children());
```
zml2008, line 319: is this anywhere accurate for a guess?

zml2008 (changes requested):
thanks for proposing this -- would like to see some changes to make the rendering features here a bit more generally applicable, but it has the potential to be a good addition

api/src/main/java/net/kyori/adventure/text/width/PixelWidthSourceImpl.java
kashike: (basic multilingual plane)

api/src/main/java/net/kyori/adventure/text/width/PixelWidthSourceImpl.java
KingOfSquares: I forgot to mention that in the main post. I was unsure weather to "guess" using info the component has without rendering, or to just ignore it whenever it does not get resolved

api/src/main/java/net/kyori/adventure/text/width/PixelWidthSourceImpl.java
Electroid: I think the API should provide a default, otherwise it's going to be too complicated for the average API user to use it.

api/src/main/java/net/kyori/adventure/text/width/PixelWidthSourceImpl.java
```diff
@@ -0,0 +1,368 @@
...
+      return;
+    } else if(resolveFor.isAssignableFrom(NBTComponent.class)) {
+      if(resolveFor.isAssignableFrom(BlockNBTComponent.class)) {
+        this.blockNBTResolvers.add((ComponentResolver<BlockNBTComponent, CX>) resolver);
+        return;
+      } else if(resolveFor.isAssignableFrom(EntityNBTComponent.class)) {
+        this.entityNBTResolvers.add((ComponentResolver<EntityNBTComponent, CX>) resolver);
+        return;
+      } else if(resolveFor.isAssignableFrom(StorageNBTComponent.class)) {
+        this.storageNBTResolvers.add((ComponentResolver<StorageNBTComponent, CX>) resolver);
```
Electroid, line 223: The more I look at this, the more I think `width` should just be method of `TextComponent`. I think the current approach (in totality) is overly complex. It's not practical to calculate the `width` of non-text.

zml2008: hm that would play well into my hope for "two systems" -- some system for component 'flattening' is still important to have, then this system for width calculations.

Though -- since text components can have non-text children, I don't think such a method could be safely put on TextComponent

api/src/main/java/net/kyori/adventure/text/width/PixelWidthSourceImpl.java
zml2008: I never said it shouldn't be part of the API -- this should just be something reproducible based on data for any font, not magic.

api/src/main/java/net/kyori/adventure/util/CharToIntFunction.java
KingOfSquares: What do you mean, like `CharacterWidthFunction`?

api/src/main/java/net/kyori/adventure/text/width/PixelWidthSourceImpl.java
KingOfSquares: Width is only calculated on text, but other components may need rendering before text representative of their content can be displayed. The resolvers added here does not calculate anything, they just open the possibility of resolving components so the text they represent can be used for a precise calculation.

api/src/main/java/net/kyori/adventure/text/width/PixelWidthSourceImpl.java
KingOfSquares: If handling of supplementary characters is needed we can't use `char` in the functions calculating the width for each symbol, which makes it less intuitive for people to make their own (for resource packs++). Do we actually need to support that?

zml2008: People all around the world play Minecraft, and they usually like to have things work correctly when they speak their language.

api/src/main/java/net/kyori/adventure/text/width/PixelWidthSourceImpl.java
zml2008: You should try to match whatever the Vanilla client does for unresolvable content, to give the most accurate results.

zml2008:
Another note: There are some formatting issues in here that Checkstyle didn't catch. I've updated the adventure checkstyle to start resolving that, so if you rebase this PR you should get some feedback to fix.

api/src/main/java/net/kyori/adventure/util/CharToIntFunction.java
kashike: Yes.

api/src/main/java/net/kyori/adventure/text/width/CharacterWidthFunction.java
```diff
@@ -0,0 +1,43 @@
...
+package net.kyori.adventure.text.width;
+
+import net.kyori.adventure.text.format.Style;
+
+/**
+ * A function that takes a {@code char} and a {@link Style} as input and produces an {@code int}.
+ *
+ * @since 4.4.0
+ */
+
```
kashike, line 33:
```suggestion
```

api/src/main/java/net/kyori/adventure/text/width/CharacterWidthFunction.java
```diff
@@ -0,0 +1,43 @@
...
+@FunctionalInterface
+public interface CharacterWidthFunction {
+
```
kashike, line 36:
```suggestion
```

api/src/main/java/net/kyori/adventure/text/width/CharacterWidthFunction.java
```diff
@@ -0,0 +1,43 @@
...
+ * SOFTWARE.
+ */
+package net.kyori.adventure.text.width;
+
+import net.kyori.adventure.text.format.Style;
+
+/**
+ * A function that takes a {@code char} and a {@link Style} as input and produces an {@code int}.
+ *
+ * @since 4.4.0
```
kashike, line 31:
```suggestion
 * @since 4.6.0
```
Applies to all occurrences in this PR.

kashike (changes requested):

KingOfSquares:
Decided to finally fix the last comment from zml with generating a proper default function for calculating. I created a little tool (that's kinda scuffed rn but it works) that I used to generate a java class which I copied over. 

I forgot to update the tests so ill do that first...

zml2008:
After some discussion, we've decided that it'll be best for this to be developed further as an external library. Thanks for your time!

PR #265, opened 2021-01-22
nbt: support 1.16 format

extclp:


nbt/src/main/java/net/kyori/adventure/nbt/BinaryTagTypes.java
```diff
@@ -118,10 +121,12 @@
     }
   }, (tag, output) -> {
     output.writeByte(tag.elementType().id());
-    final int size = tag.size();
-    output.writeInt(size);
-    for(final BinaryTag item : tag) {
-      BinaryTagType.write(item.type(), item, output);
+    if(tag.elementType() != END){
```
zml2008, line 124: wrong formatting, look at what the rest of the file does

zml2008 (changes requested):
This is incorrect. You conditionally read + write the length field -- Mojang's implemenattion unconditionally reads and writes this field.

Please add tests for this as well -- ideally testing against a NBT file containing an empty list produced by the Mojang implementation of the format.

PR #279, opened 2021-02-12
text-serializer-legacy: Vanilla implementation for URL conversion

Gabik21:
Fixes a bug where legacy text containing URLs without protocol specification would cause a crash on old minecraft versions.
Modern minecraft versions just refuse to open the URL dialog without this.


zml2008:
This doesn't work with non-`http` URL schemes. While those are prohibited by a vanilla client, we don't want to prevent their use in a modded environment, or pass differently invalid data to the client by prepending a second scheme to an otherwise valid URL. Perhaps test with a regex that matches the scheme pattern (defined in section 3.1 of [RFC2396](https://www.ietf.org/rfc/rfc2396.txt)) instead.

Gabik21:
Makes sense, added that.

zml2008 (approved):

kashike (approved):

PR #280, opened 2021-02-18
api: Add asHSV to RGBLike and use HSV for color comparisons in NamedTextColor.nearestTo

jpenilla:
The code for RGB to HSV conversion is based on the example code [here](https://www.cs.rit.edu/~ncs/color/t_convert.html), modified to behave more like java.awt.Color.RGBtoHSB.

In my limited testing, this algorithm for color distance seems to match gray less often, leading to a more "correct" result for color down sampling.

comparison:

original:
original
original
old downsample:
old
old
new downsample:
original
new

### DIFF
--- a/api/src/main/java/net/kyori/adventure/util/HSVLikeImpl.java
+++ b/api/src/main/java/net/kyori/adventure/util/HSVLikeImpl.java
@@ -23,46 +23,52 @@
  */
 package net.kyori.adventure.util;
 
-import org.checkerframework.checker.nullness.qual.NonNull;
-import org.checkerframework.common.value.qual.IntRange;
+import net.kyori.examination.string.StringExaminer;
+import org.checkerframework.checker.nullness.qual.Nullable;
 
-/**
- * Something that can provide red, green, and blue colour components.
- *
- * @since 4.0.0
- */
-public interface RGBLike {
-  /**
-   * Gets the red component.
-   *
-   * @return the red component
-   * @since 4.0.0
-   */
-  @IntRange(from = 0x0, to = 0xff) int red();
+import java.util.Objects;
 
-  /**
-   * Gets the green component.
-   *
-   * @return the green component
-   * @since 4.0.0
-   */
-  @IntRange(from = 0x0, to = 0xff) int green();
+final class HSVLikeImpl implements HSVLike {
+  private final float h;
+  private final float s;
+  private final float v;
 
-  /**
-   * Gets the blue component.
-   *
-   * @return the blue component
-   * @since 4.0.0
-   */
-  @IntRange(from = 0x0, to = 0xff) int blue();
+  HSVLikeImpl(final float h, final float s, final float v) {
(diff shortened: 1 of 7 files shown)


kashike (changes requested):

api/src/main/java/net/kyori/adventure/util/RGBLike.java
```diff
@@ -54,4 +55,46 @@
...
    */
   @IntRange(from = 0x0, to = 0xff) int blue();
+
+  /**
+   * Converts the color represented by this RGBLike to the HSV color space. Result values will be in the range [0, 1].
+   *
+   * @return an array of three elements containing the hue, saturation, and value (in that order), of the color represented by this RGBLike
+   * @since 4.6.0
+   */
+  default float@NonNull[] asHSV() {
```
kashike, line 65: On second thought... do we want to add `HSVLike` instead of just returning a `float` array?
```java
interface HSVLike {
  float h();
  float s();
  float v();
}
```

kashike (changes requested):

api/src/main/java/net/kyori/adventure/util/RGBLike.java
jpenilla: Would `HSVLike` be something `TextColor` would extend, or would it have a separate implementation? The problem with `TextColor` extending it would be we need to convert RGB -> HSV each time we call `h`, `s`, and `v` unless we cache/precompute the HSV values for every `TextColor`.

kashike: No, standalone interface + class.

api/src/main/java/net/kyori/adventure/util/HSVLikeImpl.java
```diff
@@ -0,0 +1,74 @@
...
+  public float v() {
+    return this.v;
+  }
+
+  @Override
+  public boolean equals(final @Nullable Object other) {
+    if(this == other) return true;
+    if(!(other instanceof HSVLikeImpl)) return false;
+    final HSVLikeImpl that = (HSVLikeImpl) other;
+    return Float.compare(that.h, this.h) == 0 && Float.compare(that.s, this.s) == 0 && Float.compare(that.v, this.v) == 0;
```
kashike, line 62: Should be able to use [`ShadyPines.equals`](https://github.com/KyoriPowered/adventure/blob/master/api/src/main/java/net/kyori/adventure/util/ShadyPines.java#L69-L79) here?

kashike (changes requested):

api/src/main/java/net/kyori/adventure/util/HSVLikeImpl.java
jpenilla: Didn't know of that method until now, but looks like yes

zml2008 (approved):

kashike (approved):

Issue #286, opened 2021-02-26
Add a collector to join components together out of a Stream


zml2008:
Based on https://gitlab.com/stellardrift/featherchat/-/blob/master/src/main/java/ca/stellardrift/featherchat/TextFormats.java#L77 -- we'd need to add the abiilty to check if a builder has children or not, but would be fairly straightforward

cc @ItsDoot

PR #292, opened 2021-02-28
api: Add methods to join components using a collector

zml2008:
Closes GH-286

### DIFF
--- a/api/src/main/java/net/kyori/adventure/text/ComponentBuilder.java
+++ b/api/src/main/java/net/kyori/adventure/text/ComponentBuilder.java
@@ -23,6 +23,7 @@
  */
 package net.kyori.adventure.text;
 
+import java.util.List;
 import java.util.Set;
 import java.util.function.Consumer;
 import java.util.function.Function;
@@ -154,6 +155,14 @@ public interface ComponentBuilder<C extends BuildableComponent<C, B>, B extends 
    */
   @Contract("_ -> this")
   @NonNull B mapChildrenDeep(final @NonNull Function<BuildableComponent<?, ?>, ? extends BuildableComponent<?, ?>> function);
+
+  /**
+   * Get an unmodifiable list containing all children currently in this builder.
+   *
+   * @return the list of children
+   * @since 4.6.0
+   */
+  @NonNull List<Component> children();
 
   /**
    * Sets the style.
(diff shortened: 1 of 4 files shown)


api/src/main/java/net/kyori/adventure/text/ComponentBuilder.java
```diff
@@ -155,6 +156,14 @@
   @Contract("_ -> this")
   @NonNull B mapChildrenDeep(final @NonNull Function<BuildableComponent<?, ?>, ? extends BuildableComponent<?, ?>> function);
 
+  /**
+   * Get an unmodifiable list containing all children currently in this builder.
+   *
+   * @return the list of children
+   * @since 4.6.0
+   */
+  @NonNull List<Component> children();
```
kashike, line 165: Is this method necessary?

zml2008: it's not strictly required -- the collector could just work with the builder implementation -- but it feels weird to not expose functionality the collector needs to the api.

api/src/test/java/net/kyori/adventure/text/AbstractComponentTest.java
```diff
@@ -376,4 +377,40 @@ void testBuilderApplyDeep() {
...
+      .append(Component.space())
+      .append(Component.text("World"))
+      .build();
+
+    assertEquals(expected, joined);
+
+  }
+
+  @Test
+  void testJoinEmpty() {
```
kashike, line 413:
```suggestion
  void testCollectOnEmptyStreamReturnsEmptyComponent() {
```

api/src/test/java/net/kyori/adventure/text/AbstractComponentTest.java
```diff
@@ -376,4 +377,40 @@ void testBuilderApplyDeep() {
...
+    ).collect(Component.toComponent(Component.space()));
+
+    final Component expected = Component.text()
+      .append(Component.text("Hello", NamedTextColor.RED))
+      .append(Component.space())
+      .append(Component.text("World"))
+      .build();
+
+    assertEquals(expected, joined);
+
```
kashike, line 409:
```suggestion
```

kashike (changes requested):

api/src/main/java/net/kyori/adventure/text/ComponentBuilder.java
kashike: I'm not really a fan of exposing getters on builders... but it looks like we do for `content()` as well, so I guess this is okay.

kashike (approved):

PR #296, opened 2021-03-02
serializer-gson: Add deserializers for JsonElement, Reader and JsonReader

### LINKED ISSUE #293: GsonSerialiser: add deserialise(JsonElement)
The ability to create a Component from a JsonElement would be a cool thing to have.

kezz:
This PR adds three new `deserialize` methods in `GsonComponentSerializer` and the respective implementation that enable the deserialisation of `JsonReader`, `JsonElement` and `Reader` objects.

I'm not sure if this was the best way to add them as they won't be accessible using the `ComponentSerializer` interface. I suppose each new deserialisation method could be put in it's own `ComponentSerializer` but it feels best suited to the `GsonComponentSerializer`.

I also haven't introduced any tests yet. I was not sure what the best way to go about introducing tests would be. Perhaps a separate class or mutating the existing ones to test against all the deserialisers? I'd appreciate some pointers!

Merging this PR will close #293.

### DIFF
--- a/text-serializer-gson/src/main/java/net/kyori/adventure/text/serializer/gson/GsonComponentSerializerImpl.java
+++ b/text-serializer-gson/src/main/java/net/kyori/adventure/text/serializer/gson/GsonComponentSerializerImpl.java
@@ -25,7 +25,12 @@ package net.kyori.adventure.text.serializer.gson;
 
 import com.google.gson.Gson;
 import com.google.gson.GsonBuilder;
+
+import java.io.Reader;
 import java.util.function.UnaryOperator;
+
+import com.google.gson.JsonElement;
+import com.google.gson.stream.JsonReader;
 import net.kyori.adventure.key.Key;
 import net.kyori.adventure.text.BlockNBTComponent;
 import net.kyori.adventure.text.Component;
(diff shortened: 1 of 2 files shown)


zml2008 (approved):
lgtm, not sure how relevant tests are given you're just delegating to gson here.

not entirely sold on how necessary these are, but the implementation is fine at least

kezz:
> not entirely sold on how necessary these are, but the implementation is fine at least

Yeah I understand what you mean! There's nothing here that can't be done with just a few lines of code anyway - I suppose it makes catching exceptions a bit easier and more unified?

text-serializer-gson/src/main/java/net/kyori/adventure/text/serializer/gson/GsonComponentSerializerImpl.java
```diff
@@ -90,6 +93,27 @@
...
+    return component;
+  }
+
+  @Override
+  public @NonNull Component deserialize(final @NonNull JsonReader input) {
+    final Component component = this.serializer().fromJson(input, Component.class);
+    if(component == null) throw ComponentSerializerImpl.notSureHowToDeserialize(input);
+    return component;
+  }
+
```
kashike, line 116: Is there really a need for these two separate methods?

kashike (changes requested):

text-serializer-gson/src/main/java/net/kyori/adventure/text/serializer/gson/GsonComponentSerializerImpl.java
kezz: Not particularly, but `JsonReader` doesn't implement `Reader` so they're two completely different objects.

PR #301, opened 2021-03-04
api: Mark ScoreComponent's value as deprecated since it is no longer supported

zml2008:
Mojang removed this field from 1.16, so it will be ignored when used there.

### DIFF
--- a/api/src/main/java/net/kyori/adventure/text/ScoreComponentImpl.java
+++ b/api/src/main/java/net/kyori/adventure/text/ScoreComponentImpl.java
@@ -36,6 +36,7 @@ import static java.util.Objects.requireNonNull;
 final class ScoreComponentImpl extends AbstractComponent implements ScoreComponent {
   private final String name;
   private final String objective;
+  @Deprecated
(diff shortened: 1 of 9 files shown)


api/src/main/java/net/kyori/adventure/text/ScoreComponentImpl.java
```diff
@@ -68,11 +68,13 @@
   }
 
   @Override
+  @Deprecated
   public @Nullable String value() {
     return this.value;
```
kashike, line 73: Deprecate the field too, please.

kashike (changes requested):

kashike (approved):

KingOfSquares:
When you are committing a change for the score value, could you maybe change the javadoc on the ScoreComponent as well?
https://github.com/KyoriPowered/adventure/blob/e27430962ca2648321ac00f9c39af9b908364d0a/api/src/main/java/net/kyori/adventure/text/ScoreComponent.java#L41

I've discovered that the value is _always_ used if present, even if another score is found.

Issue #114, opened 2020-07-20
Add methods to compact text


Katrix:
Would be neat to have a method that would try to pack the Text tree as compact as possible while keeping the formatting. I'd be happy to port my implementation from [MineJson](https://github.com/Katrix/MineJson/blob/master/minejsonText/src/main/scala/net/katsstuff/minejson/text/textObjs.scala#L85)

PR #308, opened 2021-03-13
api: Add support for 'compacting' components

zml2008:
Closes GH-114

This ports the implementation of component compacting over from the original Kotlin provided by @CrushedPixel (see that [here](https://paste.gg/p/zml/89fb51d013064ad6b6bb279c7486d378)) 

While this component optimization is expensive enough that it shouldn't be used on every message sent, it can be useful for ensuring components of unknown origin (such as loaded from a configuration, or built from multiple sources) are as compact as possible before being sent to users.

Some things to discuss:

- The name: Does `compact` make sense, or should we use something else like `optimize` or `pack`?
- Should a setting be added to the serializers (mainly the Configurate serializer, but possibly others) to auto-compact when loading?

### DIFF
--- a/api/src/main/java/net/kyori/adventure/text/Component.java
+++ b/api/src/main/java/net/kyori/adventure/text/Component.java
@@ -1916,6 +1916,14 @@ public interface Component extends ComponentBuilderApplicable, ComponentLike, Ex
   @NotNull Component replaceText(final @NotNull TextReplacementConfig config);
 
   /**
+   * Create a new component with any redundant style elements or children removed.
+   *
+   * @return the optimized component
+   * @since 4.9.0
+   */
+  @NotNull Component compact();
+
+  /**
    * Finds and replaces text within any {@link Component}s using a string literal.
    *
    * @param search a string literal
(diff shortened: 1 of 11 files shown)


octylFractal (approved):
kewl, maybe call it `optimize` if you're gonna call it that internally anyways

api/src/main/java/net/kyori/adventure/text/ComponentCompacting.java
```diff
@@ -0,0 +1,152 @@
...
+final class ComponentCompacting {
+  private static final TextDecoration[] DECORATIONS = TextDecoration.values();
+
+  private ComponentCompacting() {
+  }
+
+  static @NonNull Component optimize(final AbstractComponent component, final @Nullable Style parentStyle) {
+    Component optimized = component.children(Collections.emptyList());
+
+    if(parentStyle != null) {
```
kashike, line 45:
```suggestion
    if (parentStyle != null) {
```
(entire pull request needs reformatting)

kashike (changes requested):

api/src/main/java/net/kyori/adventure/text/Component.java
```diff
@@ -1693,6 +1693,14 @@ default boolean hasStyling() {
   @Contract(pure = true)
   @NonNull Component replaceText(final @NonNull TextReplacementConfig config);
 
+  /**
+   * Create a new component with any redundant style elements or children removed.
+   *
+   * @return the optimized component
+   * @since 4.8.0
```
kashike, line 1700:
```suggestion
   * @since 4.9.0
```

kashike (changes requested):

kashike (approved):

PR #310, opened 2021-03-15
api: Add component visitor

kezz:
This PR adds three simple methods to visit a component alongside some simple unit tests. These three methods are:

- `visit(Consumer<Component>)`: calls `#accept` on the consumer for each element in the component,
- `visitUntil(Predicate<Component>)`: calls `#test` on the predicate for each element in the component until the predicate returns `true`, and
- `visitWhile(Predicate<Component>)`: calls `#test` on the predicate for each element in the component while the predicate returns true.

I'm not sure if this could be written in such a way that we can guarantee the order in which components can be visited. I'm not sure I'm knowledgeable enough of the internals to know if that's something that can be done or even should be done. I'll investigate further if needs be.

### DIFF
--- a/api/src/main/java/net/kyori/adventure/text/Component.java
+++ b/api/src/main/java/net/kyori/adventure/text/Component.java
@@ -31,6 +31,7 @@ import java.util.Map;
 import java.util.Set;
 import java.util.function.Consumer;
 import java.util.function.Function;
+import java.util.function.Predicate;
 import java.util.function.UnaryOperator;
 import java.util.regex.Pattern;
 import java.util.stream.Collector;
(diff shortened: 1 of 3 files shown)


kezz:
Just spotted a few issues with this and improvements that could be made but I'll wait and see what everyone thinks of the idea of this PR before taking it further!

api/src/main/java/net/kyori/adventure/text/Component.java
```diff
@@ -1656,6 +1657,42 @@ default boolean hasStyling() {
...
+
+  /**
+   * Calls the provided visitor for each component and all of the component's children in
+   * this component while the visitor returns {@code true}. There is no guarantee of the
+   * order in which each component will be visited.
+   *
+   * @param visitor the component visitor
+   * @since 4.8.0
+   */
+  void visitWhile(final @NonNull Predicate<Component> visitor);
```
Electroid, line 1694: Why don't we use the `Iterator`/`Iterable` pattern here? That would be a more standard way of traversing the tree.

```java
Iterator<Component> iterator();
Iterator<Component> iterator(/* some config? */);
```

If you want to bail early, you just stop iterating. And this would work nicely with loops.

kezz: Nice idea - it would be relatively simple to mock up a `ComponentIterator` that nicely traverses a component structure.

kezz:
Superseded by #339.

Issue #315, opened 2021-03-27
Add a way to send an Audience a sound playing from an entity


kezz:
Put simply, it would be nice to have an Adventure way to send an [entity sound effect](https://wiki.vg/Protocol#Entity_Sound_Effect) packet that follows a given entity.

PR #316, opened 2021-03-27
api: Add support for playing sound with an "emitter"

kezz:
This is a simple PR that adds the concept of a sound "player". The interface for `Sound.Player` is intentionally empty, as the way the client/server identifies a sound player is with the entity ID, an implementation detail that some platforms may not wish to expose. 

The general idea is that a native implementation will implement `Sound.Player` on their entity class, allowing it to be passed directly as an argument into the `playSound` method. For platforms, they could provide utility methods to convert their entity class to an implementation of the `Sound.Player` interface which provides an entity object, or some other form of identifier. This can then be used to convert to the platform specific way of sending the entity sound effect packet.

Alternatively, the `Sound.Player` could provide an integer ID, but this is just exposing what is essentially a magic number, not something as constant as, say, a UUID. I'm also not 100% sold on the `Player` part, I think it's potentially too confusing with the actual player. Perhaps something like `Sound.Maker` or `Sound.EntitySource`?

### DIFF
--- a/api/src/main/java/net/kyori/adventure/audience/Audience.java
+++ b/api/src/main/java/net/kyori/adventure/audience/Audience.java
@@ -479,6 +479,17 @@ public interface Audience extends Pointered {
   }
 
   /**
+   * Plays a sound from a player, usually an entity.
+   *
+   * @param sound a sound
+   * @param player a player
+   * @see Sound
+   * @since 4.8.0
+   */
+  default void playSound(final @NonNull Sound sound, final Sound.@NonNull Player player) {
+  }
+
+  /**
    * Stops a sound, or many sounds.
    *
    * @param stop a sound stop
--- a/api/src/main/java/net/kyori/adventure/audience/Audience.java
+++ b/api/src/main/java/net/kyori/adventure/audience/Audience.java
@@ -479,14 +479,14 @@ public interface Audience extends Pointered {
   }
 
   /**
-   * Plays a sound from a player, usually an entity.
+   * Plays a sound from an emitter, usually an entity.
    *
    * @param sound a sound
-   * @param player a player
+   * @param emitter an emitter
    * @see Sound
    * @since 4.8.0
    */
-  default void playSound(final @NonNull Sound sound, final Sound.@NonNull Player player) {
+  default void playSound(final @NonNull Sound sound, final Sound.@NonNull Emitter emitter) {
   }
 
   /**
--- a/api/src/main/java/net/kyori/adventure/audience/Audience.java
+++ b/api/src/main/java/net/kyori/adventure/audience/Audience.java
@@ -488,12 +488,12 @@ public interface Audience extends Pointered {
    *   In this case the sound will be played at the location of the emitter and will not follow them.
    * </p>
    *
+   * <p>To play a sound that follows the recipient, use {@link Sound.Emitter#self()}.</p>
+   *
    * <p><b>Note</b>: Due to <a href="https://bugs.mojang.com/browse/MC-138832">MC-138832</a>, the volume and pitch may be ignored when using this method.</p>
    *
    * @param sound a sound
    * @param emitter an emitter
-   * @see Sound
-   * @see Sound.Emitter
    * @since 4.8.0
    */
   default void playSound(final @NotNull Sound sound, final Sound.@NotNull Emitter emitter) {
--- a/api/src/main/java/net/kyori/adventure/audience/Audience.java
+++ b/api/src/main/java/net/kyori/adventure/audience/Audience.java
@@ -445,7 +445,9 @@ public interface Audience extends Pointered {
   }
 
   /**
-   * Plays a sound.
+   * Plays a sound at the location of the recipient of the sound.
+   *
+   * <p>To play a sound that follows the recipient, use {@link #playSound(Sound, Sound.Emitter)} with {@link Sound.Emitter#self()}.</p>
    *
    * @param sound a sound
    * @see Sound
@@ -482,8 +484,8 @@ public interface Audience extends Pointered {
    * Plays a sound from an emitter, usually an entity.
    *
    * <p>
-   *   Sounds played from an emitter will follow the entity unless the sound is custom sound.
-   *   In the case where a custom sound is provided, the sound will be played at the location of the emitter at the time of calling.
+   *   Sounds played using this method will follow the emitter unless the sound is a custom sound.
+   *   In this case the sound will be played at the location of the emitter and will not follow them.
    * </p>
    *
    * <p><b>Note</b>: Due to <a href="https://bugs.mojang.com/browse/MC-138832">MC-138832</a>, the volume and pitch may be ignored when using this method.</p>
@@ -491,6 +493,7 @@ public interface Audience extends Pointered {
    * @param sound a sound
    * @param emitter an emitter
    * @see Sound
+   * @see Sound.Emitter
    * @since 4.8.0
    */
   default void playSound(final @NotNull Sound sound, final Sound.@NotNull Emitter emitter) {
--- a/api/src/main/java/net/kyori/adventure/audience/Audience.java
+++ b/api/src/main/java/net/kyori/adventure/audience/Audience.java
@@ -481,12 +481,19 @@ public interface Audience extends Pointered {
   /**
    * Plays a sound from an emitter, usually an entity.
    *
+   * <p>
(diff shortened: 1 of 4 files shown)


api/src/main/java/net/kyori/adventure/sound/Sound.java
```diff
@@ -189,4 +189,13 @@
     @Override
     @NonNull Key key();
   }
+
+  /**
+   * A player of sounds.
+   *
+   * @since 4.8.0
+   */
+  interface Player {
```
kashike, line 198: Not sure how I feel about this.

@zml2008?

api/src/main/java/net/kyori/adventure/audience/ForwardingAudience.java
```diff
@@ -120,6 +120,11 @@ default void playSound(final @NonNull Sound sound, final double x, final double
     for(final Audience audience : this.audiences()) audience.playSound(sound, x, y, z);
   }
 
+  @Override
+  default void playSound(final @NonNull Sound sound, final Sound.@NonNull Player player) {
```
kashike, line 124: Missing an override in `ForwardingAudience.Single`

kashike (changes requested):

api/src/main/java/net/kyori/adventure/sound/Sound.java
kezz: Perhaps `Performer`?

MiniDigger: Sound Source?

kezz: Emitter is nice - Source clashes with `Sound.Source` which is unfortunate..

kashike: `Emitter` works for me.

zml2008: Emitter sounds good to me too

kezz: Emitter it is!

api/src/main/java/net/kyori/adventure/sound/Sound.java
```diff
@@ -189,4 +189,13 @@
     @Override
     @NonNull Key key();
   }
+
+  /**
+   * A emitter of sounds.
```
kashike, line 194:
```suggestion
   * An emitter of sounds.
```

api/src/main/java/net/kyori/adventure/sound/Sound.java
```diff
@@ -189,4 +189,13 @@
...
+   *
+   * @since 4.8.0
+   */
+  interface Emitter {
+
```
kashike, line 199:
```suggestion
```

kashike (changes requested):

kashike (approved):

kezz:
Just a thought: it's worth noting that whilst this does fix https://github.com/KyoriPowered/adventure/issues/315, it doesn't "fix" the subsequent problem of having a method to play sounds coming from the audience member (unless an implementation decides to use `Audience#playSound(Sound)` for that).

The issue arises from the fact that if you have a `ForwardingAudience`, you can't run code like `forwardingAudience.playSound(sound, forwardingAudience)` or something similar, you'd have to iterate through each member of the audience and play the sound to them directly. This isn't a terrible solution, as in most implementations you'd be able to do code like:
```java
for (Player player : world.getPlayers()) {
    player.playSound(sound, player);
}
```

However, this does mean that there isn't a cross-platform way to signal you'd like a sound to be played *from* each member of the audience and it is a potential problem with Adventure that there isn't a way to indicate this. Firstly, I believe that it should be specified exactly what happens with the `Audience#playSound(Sound)` method to avoid confusion across implementations as to how the sound would be played. Additionally, a method that mimics `Audience#playSound(Sound)` could be created to allow playing sound using each audience member as the `Emitter`, perhaps, as was suggested in https://github.com/PaperMC/Paper/pull/5472, an additional boolean parameter to specify if the sound should follow the receiver or not.

kashike:
`playSoundFromEmitter`?

kezz:
> `playSoundFromEmitter`?

Yeah I suppose that could work - I think it does potentially confuse the idea of `Emitter` as opposed to the "receiver" of a sound.

api/src/main/java/net/kyori/adventure/audience/Audience.java
```diff
@@ -451,6 +451,17 @@ default void playSound(final @NonNull Sound sound) {
   default void playSound(final @NonNull Sound sound, final double x, final double y, final double z) {
   }
 
+  /**
+   * Plays a sound from an emitter, usually an entity.
```
zml2008, line 455: I'd like to see this clarify what it means to play a sound "from" an entity -- it means that the location of the sound will follow the location of that entity, right?

zml2008 (changes requested):

kezz:
Rebased and added further clarification to the Javadocs for the added method/interface.

zml2008 (approved):

kezz:
Following on from zml's comment in https://github.com/KyoriPowered/adventure/issues/330#issuecomment-855286397, I have updated this PR to add `Emitter#self()` which, when passed into `Audience#playSound(Sound, Emitter)` should cause the sound to be played from the recipient of the sound.

api/src/main/java/net/kyori/adventure/audience/Audience.java
```diff
@@ -466,6 +468,25 @@ default void playSound(final @NotNull Sound sound) {
...
+   *   Sounds played using this method will follow the emitter unless the sound is a custom sound.
+   *   In this case the sound will be played at the location of the emitter and will not follow them.
+   * </p>
+   *
+   * <p><b>Note</b>: Due to <a href="https://bugs.mojang.com/browse/MC-138832">MC-138832</a>, the volume and pitch may be ignored when using this method.</p>
+   *
+   * @param sound a sound
+   * @param emitter an emitter
+   * @see Sound
+   * @see Sound.Emitter
```
zml2008, line 484:
```suggestion
   * @see Sound.Emitter#self() to make a sound follow the receiver
```

(no need to reference the specific classes, they're already linked because of being method parameters)

zml2008 (changes requested):

api/src/main/java/net/kyori/adventure/audience/Audience.java
kezz: Yeah I only included this `@see` because every other method in `Audience` has one for method params or not. Also, this would make the text for the link into `to make a sound follow the receiver`, is this what you want? Happy to just add it into the body of the docs if not.

zml2008: yeah, that should change -- I've got the commit locally to remove some of the especially redundant `@see`s, and will push once this PR makes it in.

oops -- didn't realize that. Probably the best choice is `@see <code>Sound.Emitter.self()</code> to make a sound that follows the receiver`?

api/src/main/java/net/kyori/adventure/sound/SoundImpl.java
```diff
@@ -31,6 +31,8 @@
 import org.jetbrains.annotations.Nullable;
 
 abstract class SoundImpl implements Sound {
+  static final Emitter SELF = new Emitter() {};
```
kashike, line 34:
```suggestion
  static final Emitter EMITTER_SELF = new Emitter() {};
```

kashike: Add a `toString` override.

kashike (changes requested):

zml2008 (approved):

api/src/main/java/net/kyori/adventure/sound/SoundImpl.java
```diff
@@ -31,6 +31,13 @@
 import org.jetbrains.annotations.Nullable;
 
 abstract class SoundImpl implements Sound {
+  static final Emitter EMITTER_SELF = new Emitter() {
+    @Override
+    public String toString() {
+      return "SoundEmitterSelf";
```
kashike, line 37:
```suggestion
      return "SelfSoundEmitter";
```

kashike (changes requested):

kashike (approved):

PR #318, opened 2021-03-27
api: Add a way to play a sound from a coordinate

### LINKED ISSUE #317: Add a way to play a sound from a native coordinate/position/location
Although the `Audience#playSound(Sound, double, double, double)` method is fine, it would be nice to add an interface similar to `Pos` in `BlockNBT` that looks something like: 
```java
interface Location {
    double x();
    double y();
    double z();
}
```

This would allow native implementations to implement this object, whatever it's name may be, and pass it into a new method in audience `Audience#playSound(Sound, Location)`.

kezz:
This is an example implementation of #317 using a method similar to the `RGBLike` interface used to allow the use of native platform colour objects.

Closes #317.

### DIFF
--- a/api/src/main/java/net/kyori/adventure/audience/Audience.java
+++ b/api/src/main/java/net/kyori/adventure/audience/Audience.java
@@ -35,7 +35,7 @@ import net.kyori.adventure.sound.SoundStop;
 import net.kyori.adventure.text.Component;
 import net.kyori.adventure.text.ComponentLike;
 import net.kyori.adventure.title.Title;
-import net.kyori.adventure.util.CoordinateLike;
+import net.kyori.adventure.util.Vector3dLike;
 import org.jetbrains.annotations.NotNull;
 
 /**
@@ -461,7 +461,7 @@ public interface Audience extends Pointered {
    * @param location a location
    * @since 4.8.0
    */
-  default void playSound(final @NonNull Sound sound, final @NonNull CoordinateLike location) {
+  default void playSound(final @NonNull Sound sound, final @NonNull Vector3dLike location) {
     this.playSound(sound, location.x(), location.y(), location.z());
   }
 
--- a/api/src/main/java/net/kyori/adventure/audience/Audience.java
+++ b/api/src/main/java/net/kyori/adventure/audience/Audience.java
@@ -36,6 +36,7 @@ import net.kyori.adventure.text.Component;
 import net.kyori.adventure.text.ComponentLike;
 import net.kyori.adventure.title.Title;
 import net.kyori.adventure.util.Vector3dLike;
+import net.kyori.adventure.util.Vector3iLike;
 import org.jetbrains.annotations.NotNull;
 
 /**
@@ -462,7 +463,21 @@ public interface Audience extends Pointered {
    * @see Sound
    * @since 4.8.0
    */
-  default void playSound(final @NonNull Sound sound, final @NonNull Vector3dLike pos) {
+  @ForwardingAudienceOverrideNotRequired
+  default void playSound(final @NotNull Sound sound, final @NotNull Vector3iLike pos) {
+    this.playSound(sound, pos.x(), pos.y(), pos.z());
+  }
+
+  /**
+   * Plays a sound at a position.
+   *
+   * @param sound a sound
+   * @param pos a position
+   * @see Sound
+   * @since 4.8.0
+   */
+  @ForwardingAudienceOverrideNotRequired
+  default void playSound(final @NotNull Sound sound, final @NotNull Vector3dLike pos) {
     this.playSound(sound, pos.x(), pos.y(), pos.z());
   }
 
--- a/api/src/main/java/net/kyori/adventure/audience/Audience.java
+++ b/api/src/main/java/net/kyori/adventure/audience/Audience.java
@@ -35,6 +35,7 @@ import net.kyori.adventure.sound.SoundStop;
 import net.kyori.adventure.text.Component;
 import net.kyori.adventure.text.ComponentLike;
 import net.kyori.adventure.title.Title;
+import net.kyori.adventure.util.LocationLike;
 import net.kyori.adventure.util.Vector3dLike;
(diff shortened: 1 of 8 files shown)


api/src/main/java/net/kyori/adventure/util/CoordinateLike.java
```diff
@@ -0,0 +1,68 @@
...
+ * SOFTWARE.
+ */
+package net.kyori.adventure.util;
+
+/**
+ * Something that can provide x, y, and z components.
+ *
+ * @since 4.8.0
+ */
+public interface CoordinateLike {
```
kashike, line 31:
```suggestion
public interface Vector3dLike {
```

api/src/main/java/net/kyori/adventure/util/CoordinateLikeImpl.java
```diff
@@ -0,0 +1,56 @@
...
+ * SOFTWARE.
+ */
+package net.kyori.adventure.util;
+
+/**
+ * Something that can provide x, y, and z components.
+ *
+ * @since 4.8.0
+ */
+class CoordinateLikeImpl implements CoordinateLike {
```
kashike, line 26:
```suggestion
final class Vector3dImpl implements Vector3dLike {
```

kashike: Requires `toString` (using examination), `equals`, and `hashCode`.

kashike (changes requested):

api/src/main/java/net/kyori/adventure/audience/Audience.java
```diff
@@ -438,6 +439,17 @@ default void hideBossBar(final @NonNull BossBar bar) {
...
 
+  /**
+   * Plays a sound at a location.
+   *
+   * @param sound a sound
+   * @param location a location
+   * @since 4.8.0
+   */
+  default void playSound(final @NonNull Sound sound, final @NonNull Vector3dLike location) {
+    this.playSound(sound, location.x(), location.y(), location.z());
```
kashike, line 450:
```suggestion
   * Plays a sound at a position.
   *
   * @param sound a sound
   * @param pos a position
   * @since 4.8.0
   */
  default void playSound(final @NonNull Sound sound, final @NonNull Vector3dLike pos) {
    this.playSound(sound, pos.x(), pos.y(), pos.z());
```

kashike (changes requested):

api/src/main/java/net/kyori/adventure/util/Vector3dLike.java
```diff
@@ -0,0 +1,82 @@
...
+  /**
+   * Creates a new Vector3dLike.
+   *
+   * @param x the x coordinate
+   * @param y the y coordinate
+   * @param z the z coordinate
+   * @return a new CoordinateLike
+   * @since 4.8.0
+   */
+  static Vector3dLike of(final double x, final double y, final double z) {
```
zml2008, line 46: I don't know if adventure itself should provide an implementation of this interface.

MiniDigger:
if we add Vector3dLike, shouldn't we also provide Vector3iLike?

api/src/main/java/net/kyori/adventure/util/Vector3dLike.java
kezz: There's essentially zero chance that anyone implementing Adventure won't provide an object that can implement this. I suppose it can't hurt but happy to remove this.

kezz:
> if we add Vector3dLike, shouldn't we also provide Vector3iLike?

It's still `x, y, z` right? We could generify this into just `Vector3Like`, that's how Unity names their corresponding object for example.

kashike:
> > if we add Vector3dLike, shouldn't we also provide Vector3iLike?
> 
> It's still `x, y, z` right? We could generify this into just `Vector3Like`, that's how Unity names their corresponding object for example.

No. Different return types.

kezz:
> Different return types.
Ah right. In that case yes I agree.

kezz:
I've just rebased this with the addition of `Vector3iLike` and the removal of the implementation for `Vector3dLike`. However, this does need further consideration as to how we would or if this PR even should handle worlds. Many platforms have objects that represent a location in a given world and this could produce unexpected behavior with code like this (using Paper as an example):

```java
Location location = new Location(Bukkit.getWorld("test"), 0.0, 0.0, 0.0);
Bukkit.getServer().playSound(sound, location); // why can't i do this? or why is it playing for people not in the "test" world?
```

I would be interested to hear everyone's thoughts on this, but I feel like the best way is what I have just added in this PR which is to add *another* location-ish object that does not provide any location in order to let platforms implement this as they feel fit. Would be happy to pull this last commit into another PR if necessary.

Additionally, the current implementation of the `playSound` method that takes a `Vector3iLike` just turns each integer into a double directly, should this be instead using the center of a location (i.e. adding 0.5 to each integer)?

api/src/main/java/net/kyori/adventure/util/LocationLike.java
```diff
@@ -0,0 +1,34 @@
...
+package net.kyori.adventure.util;
+
+/**
+ * Something that represents a location in a specific world/dimension/instance/etc.
+ *
+ * <p>Implementations should implement this for position-based classes which are more specific than {@link Vector3iLike} or {@link Vector3dLike}.</p>
+ *
+ * @since 4.8.0
+ */
+public interface LocationLike {
```
zml2008, line 33: not a fan of an empty interface here -- and tbh not sure how important it is to have a worldly variant of `playSound`

zml2008 (changes requested):

kezz:
Been thinking about this one a bit more, and I'm just going to close this PR. I haven't seen any need for something like this in Adventure yet (nor has anyone else expressed an interest). Stateful API (that requires the developer to have information about the platform at hand, such as location, world, etc.) is perhaps just not as useful in Adventure anyway.

PR #321, opened 2021-03-31
WIP: Add weather API

kezz:
# Weather
This PR adds a weather module to Adventure, providing an API for controlling the weather. The current implementation is more of a proof of concept then a final design. I'm not even 100% sure if such a thing fits in Adventure like this, or if it could be its own repo or something completely different. I'm most interested in hearing what everyone thinks!

## Holder
Similar to `Audience` the weather module includes the concept of a `WeatherHolder`, that being something that can have weather applied and checked. The current weather can be obtained from a holder using `WeatherHolder#weather()` and can be set using `WeatherHolder#weather(Weather)`.

Additionally, the concept of a `ChildWeatherHolder` exists. This represents an object that either has its own weather, or has the weather of it's parent. For example, a player could have their own weather overriding that of the world's weather.

## Weather
The `Weather` object itself is constructed using a type and a length. The type is one of rain, clear or thunder, mimicking the `/weather` command in vanilla Minecraft. The length object either holds an exact duration or is marked as random, mimicking the natural flow of weather in whatever platform you are using.

## Examples
Here's some examples of how the system would work.

```java
// clear the weather for a random amount of time
world.weather(Weather.clear());

// nope let's make it rain for exactly 10 minutes
world.weather(Weather.rain(Length.of(Duration.ofMinutes(5))));

// assuming player is a child of world, this would be true
var isTrue = player.weather().type() == Type.RAIN;

// however, we can override the weather for the child and make it permanent
player.weather(Weather.clear(Length.forever()));

// now the child has it's own weather, this will be false
var isFalse = player.weather.type() == Type.RAIN;
```

### DIFF
--- a/weather/src/main/java/net/kyori/adventure/weather/Weather.java
+++ b/weather/src/main/java/net/kyori/adventure/weather/Weather.java
@@ -1,0 +1,224 @@
+/*
+ * This file is part of adventure, licensed under the MIT License.
+ *
+ * Copyright (c) 2017-2021 KyoriPowered
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in all
+ * copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+ * SOFTWARE.
+ */
+package net.kyori.adventure.weather;
+
+import java.time.Duration;
+import java.util.stream.Stream;
+import net.kyori.examination.Examinable;
+import net.kyori.examination.ExaminableProperty;
+import org.checkerframework.checker.nullness.qual.NonNull;
+import org.checkerframework.checker.nullness.qual.Nullable;
+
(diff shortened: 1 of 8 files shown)


weather/src/main/java/net/kyori/adventure/weather/Weather.java
```diff
@@ -0,0 +1,224 @@
...
+  }
+
+  /**
+   * Rain with a specific length.
+   *
+   * @param length the length
+   * @return the weather
+   * @since 4.8.0
+   */
+  static @NonNull Weather rain(Weather.@NonNull Length length) {
```
kashike, line 80:
```suggestion
  static @NonNull Weather rain(final @NonNull Length length) {
```

weather/src/main/java/net/kyori/adventure/weather/Weather.java
```diff
@@ -0,0 +1,224 @@
...
+  }
+
+  /**
+   * Thunder with a specific length.
+   *
+   * @param length the length
+   * @return the weather
+   * @since 4.8.0
+   */
+  static @NonNull Weather thunder(Weather.@NonNull Length length) {
```
kashike, line 101:
```suggestion
  static @NonNull Weather thunder(final @NonNull Length length) {
```

weather/src/main/java/net/kyori/adventure/weather/Weather.java
```diff
@@ -0,0 +1,224 @@
...
+
+  /**
+   * Weather with a specific type and length.
+   *
+   * @param type the type
+   * @param length the length
+   * @return the weather
+   * @since 4.8.0
+   */
+  static @NonNull Weather weather(final @NonNull Type type, Weather.@NonNull Length length) {
```
kashike, line 113:
```suggestion
  static @NonNull Weather weather(final @NonNull Type type, final @NonNull Length length) {
```

weather/src/main/java/net/kyori/adventure/weather/Weather.java
```diff
@@ -0,0 +1,224 @@
...
+  Weather.@NonNull Length length();
+
+  /**
+   * Gets the type of this weather.
+   *
+   * @return the type
+   * @see Type
+   * @since 4.8.0
+   */
+  @NonNull Type type();
```
kashike, line 132: Move this above the `length()` getter.

weather/src/main/java/net/kyori/adventure/weather/Weather.java
```diff
@@ -0,0 +1,224 @@
...
+     * @since 4.8.0
+     */
+    RAIN,
+
+    /**
+     * Thunder.
+     *
+     * @since 4.8.0
+     */
+    THUNDER
```
kashike, line 167:
```suggestion
    THUNDER;
```

kashike (changes requested):
codestyle

Electroid:
I would agree with @TheMolkaPL here.

kezz:
I wonder if this is something that might be better suited in another repo, perhaps `adventure-extra` or something with a different name entirely to handle weather alongside other aspects of Minecraft that could be abstracted into an API like this? Each part would be its own module so that they could be implemented part by part if desired.

Electroid:
Appreciate the rationale. Deciding what's in and out of scope is one of the most challenging parts of maintaining a project. I think though we have some consensus that weather, even in an "extra" project, is out of scope.

PR #327, opened 2021-04-08
api: Expose ComponentLike -> Component conversion method

kashike:


api/src/main/java/net/kyori/adventure/text/ComponentLike.java
```diff
@@ -33,6 +38,45 @@
...
+   * @since 4.8.0
+   */
+  static @NonNull List<Component> asComponents(final @NonNull List<? extends ComponentLike> likes, final @Nullable Predicate<? super Component> filter) {
+    if(likes.isEmpty()) {
+      // We do not need to create a new list if the one we are copying is empty - we can
+      // simply just return our known-empty list instead.
+      return Collections.emptyList();
+    }
+    final int size = likes.size();
+    final List<Component> components = new ArrayList<>(size);
```
octylFractal, line 69: Consider calling `components.trimToSize();` or whatever it is on `ArrayList`, since this might way over-allocate.

octylFractal (approved):
kewl

api/src/main/java/net/kyori/adventure/text/ComponentLike.java
```diff
@@ -33,6 +38,55 @@
...
+        components.add(component);
+      }
+    }
+    if(components != null) {
+      // https://github.com/KyoriPowered/adventure/pull/327#discussion_r631420264
+      // we pre-size the list, but filtering might make the actual size much smaller
+      components.trimToSize();
+    }
+    // if we filtered all elements out, just use an empty list instead
+    if(components == null || components.isEmpty()) return Collections.emptyList();
```
octylFractal, line 86: Isn't the latter half of this condition impossible?

octylFractal (approved):

PR #328, opened 2021-04-09
Add sound source provider

kezz:
This PR adds an interface similar to `Sound.Type` which allows native implementations to use their existing source enums in the Adventure API.


api/src/main/java/net/kyori/adventure/sound/Sound.java
```diff
@@ -172,6 +214,21 @@
...
+     * @since 4.8.0
+     */
+    public interface Provider {
+      /**
+       * Gets the source.
+       *
+       * @return the source
+       * @since 4.8.0
+       */
+      @NonNull Source source();
```
kashike, line 230: Perhaps a more specific name here? `soundSource`

kashike (changes requested):

kashike (approved):

PR #329, opened 2021-04-09
api: More ways to stop sound

kezz:
This PR adds the following methods that simplify the process of manipulating sounds.

- `Audience#stopSound(Sound)`
Allows developers to stop specific sounds without creating a `SoundStop.
- `Sound#stopper`
Gets a `SoundStop` that will stop the sound.
- ` SoundStop#sound(Sound)`
Gets a `SoundStop` that will stop a specific sound.

### DIFF
--- a/api/src/main/java/net/kyori/adventure/sound/Sound.java
+++ b/api/src/main/java/net/kyori/adventure/sound/Sound.java
@@ -192,7 +192,7 @@ public interface Sound extends Examinable {
    * @return the sound stop
    * @since 4.8.0
    */
-  @NonNull SoundStop stopper();
+  @NotNull SoundStop asStop();
 
   /**
    * The sound source.
--- a/api/src/main/java/net/kyori/adventure/sound/Sound.java
+++ b/api/src/main/java/net/kyori/adventure/sound/Sound.java
@@ -187,6 +187,14 @@ public interface Sound extends Examinable {
   float pitch();
 
   /**
+   * Gets the {@link SoundStop} that will stop this specific sound.
+   *
+   * @return the sound stop
+   * @since 4.8.0
+   */
+  @NonNull SoundStop stopper();
+
+  /**
    * The sound source.
    *
    * @since 4.0.0
--- a/api/src/main/java/net/kyori/adventure/sound/SoundImpl.java
+++ b/api/src/main/java/net/kyori/adventure/sound/SoundImpl.java
@@ -59,7 +59,7 @@ abstract class SoundImpl implements Sound {
 
   @Override
   public @NotNull SoundStop asStop() {
-    if (this.stop == null) this.stop = SoundStop.sound(this);
+    if (this.stop == null) this.stop = SoundStop.namedOnSource(this.name(), this.source());
     return this.stop;
   }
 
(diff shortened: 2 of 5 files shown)


api/src/main/java/net/kyori/adventure/sound/SoundImpl.java
```diff
@@ -34,11 +34,13 @@
   private final Source source;
   private final float volume;
   private final float pitch;
+  private final SoundStop stopper;
```
kashike, line 37:
```suggestion
  private @MonotonicNonNull SoundStop stop;
```

api/src/main/java/net/kyori/adventure/sound/SoundImpl.java
```diff
@@ -34,11 +34,13 @@
...
 
   SoundImpl(final @NonNull Source source, final float volume, final float pitch) {
     this.source = source;
     this.volume = volume;
     this.pitch = pitch;
+    this.stopper = SoundStop.sound(this);
```
kashike, line 43:
```suggestion
```

api/src/main/java/net/kyori/adventure/sound/SoundImpl.java
```diff
@@ -56,6 +58,11 @@ public float pitch() {
     return this.pitch;
   }
 
+  @Override
+  public @NonNull SoundStop stopper() {
+    return this.stopper;
```
kashike, line 63:
```suggestion
    if(this.stop == null) {
        this.stop = SoundStop.sound(this);
    }
    return this.stop;
```

kashike (changes requested):

api/src/main/java/net/kyori/adventure/sound/Sound.java
```diff
@@ -144,6 +144,14 @@
    */
   float pitch();
 
+  /**
+   * Gets the {@link SoundStop} that will stop this specific sound.
+   *
+   * @return the sound stop
+   * @since 4.8.0
+   */
+  @NonNull SoundStop stopper();
```
kashike, line 153: Unsure of this name - @zml2008?

zml2008: hm, maybe `asStop`?

kezz: Yeah I was a bit conflicted about this, I chose stopper as it returns the thing that stops the sound and is quite clear that it wouldn't actually stop the sound (as opposed to, say, `stop`).

I like `asStop` although I don't think it's as clear as `stopper`, not that it matters too much. Happy to change if needs be!

zml2008: @kashike i don't have a super strong opinion either way at the moment, would you decide (or flip a coin?)

kezz: `asStop` is actually more in keeping with the rest of the sound API (`asSource` for example).

zml2008: alright then, let's go with `asStop`

kezz:
Renamed and rebased :)

zml2008 (approved):

api/src/main/java/net/kyori/adventure/sound/SoundStop.java
```diff
@@ -175,6 +175,18 @@
...
   }
 
+  /**
+   * Stops a specific sound.
+   *
+   * @param sound the sound
+   * @return a sound stopper
+   * @since 4.8.0
+   */
+  static @NotNull SoundStop sound(final @NotNull Sound sound) {
```
kashike, line 185: I think we can remove this method - just call `namedOnSource` from within `SoundImpl`

kashike (changes requested):

PR #330, opened 2021-04-09
api: Clarify Audience#playSound methods

kezz:
Following my comment in https://github.com/KyoriPowered/adventure/pull/316#issuecomment-816734116, this PR clarifies exactly what the `Audience#playSound(Sound)` method does, essentially binding implementations into a contract as to how they implement this method. Additionally, a new method is added that allows "both" behaviours of the `Audience#playSound` method to be run.

### DIFF
--- a/api/src/main/java/net/kyori/adventure/audience/Audience.java
+++ b/api/src/main/java/net/kyori/adventure/audience/Audience.java
@@ -451,12 +451,14 @@ public interface Audience extends Pointered {
    * @since 4.0.0
    */
   default void playSound(final @NotNull Sound sound) {
-    this.playSound(sound, false);
   }
 
   /**
    * Plays a sound.
+   *
    * <p>Custom sounds cannot be made to follow the receiver and they will fall back to playing at the location of the receiver.</p>
+   *
+   * <p><b>Note</b>: Due to <a href="https://bugs.mojang.com/browse/MC-138832">MC-138832</a>, the volume and pitch may be ignored when using this method to follow the receiver.</p>
    *
    * @param sound a sound
    * @param followReceiver if the sound should follow the receiver
@@ -464,6 +466,7 @@ public interface Audience extends Pointered {
    * @since 4.8.0
    */
   default void playSound(final @NotNull Sound sound, final boolean followReceiver) {
+    if (!followReceiver) this.playSound(sound);
   }
 
   /**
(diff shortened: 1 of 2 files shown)


api/src/main/java/net/kyori/adventure/audience/Audience.java
```diff
@@ -429,13 +429,26 @@ default void hideBossBar(final @NonNull BossBar bar) {
...
+  /**
+   * Plays a sound.
+   * <p>Custom sounds cannot be made to follow the receiver and they will fall back to playing at the location of the receiver.</p>
+   *
+   * @param sound a sound
+   * @param followReceiver if the sound should follow the receiver
+   * @see Sound
+   * @since 4.8.0
+   */
+  default void playSound(final @NonNull Sound sound, final boolean followReceiver) {
```
octylFractal, line 451: It makes me a bit uncomfortable that things that previously overrode `playSound` will now give no behavior when `playSound(sound, false)` is called, but without some crazy shenanigans it's hard to fix that.

kezz: Wouldn't be that hard to fix, we can make this method default to call the old `playSound` method and leave that one as is. I believe this is how it's done elsewhere in the codebase - I think in the boss bar listener percent/progress thing?

api/src/main/java/net/kyori/adventure/audience/Audience.java
```diff
@@ -429,13 +429,26 @@ default void hideBossBar(final @NonNull BossBar bar) {
...
    * @see Sound
    * @since 4.0.0
    */
   default void playSound(final @NonNull Sound sound) {
+    this.playSound(sound, false);
+  }
+
+  /**
+   * Plays a sound.
+   * <p>Custom sounds cannot be made to follow the receiver and they will fall back to playing at the location of the receiver.</p>
```
zml2008, line 444:
```suggestion
   * Plays a sound.
   * 
   * <p>Custom sounds cannot be made to follow the receiver and they will fall back to playing at the location of the receiver.</p>
```

zml2008 (changes requested):

kezz:
Rebased and moved the overloading so as not to break anything if any implementations haven't yet updated upon a new update that includes this PR.

zml2008:
looking at this together with #316, I think this might be better implemented by having a special `Emitter`, accessed through `Emitter.self()`, that is treated as always the receiver, rather than a separate boolean parameter.

kezz:
> looking at this together with #316, I think this might be better implemented by having a special `Emitter`, accessed through `Emitter.self()`, that is treated as always the receiver, rather than a separate boolean parameter.

This is a brilliant idea tbh - I'll add that to #316 and modify this PR to simply add the clarification to the Javadocs.

PR #331, opened 2021-04-10
Build updates and checkstyle improvements

zml2008:
This bumps us up to Gradle 7 and Indra 2

~~IntelliJ is a bit funky about its handling for multi-release, so it shows all our `exports` declarations as errors, but they compile fine. This is probably unavoidable for the time being.~~ Modules have been split into another PR

I've migrated the project over to a YAML format for Gradle's version catalogs -- we'll want to decide on styling.

### DIFF
--- a/gradle/libs.versions.yaml
+++ b/gradle/libs.versions.yaml
@@ -1,5 +1,14 @@
 metadata:
   format: {version: 1.0}
+  polyglot-extensions: [plugins]
+
+plugins:
+  net.kyori.indra: &indra 2.0.5
+  net.kyori.indra.checkstyle: *indra
+  net.kyori.indra.license-header: *indra
+  net.kyori.indra.publishing.sonatype: *indra
+  com.adarshr.test-logger: 3.0.0
+  org.jetbrains.kotlin.jvm: 1.4.32
 
 dependencies:
   # Runtime libraries
--- a/gradle/libs.versions.yaml
+++ b/gradle/libs.versions.yaml
@@ -1,0 +1,27 @@
+metadata:
+  format: {version: 1.0}
+
+dependencies:
+  # Runtime libraries
+  checkerQual: "org.checkerframework:checker-qual:3.13.0"
+  configurate3: "org.spongepowered:configurate-core:3.7.2"
+  configurate4: "org.spongepowered:configurate-core:4.1.1"
+  examination-api: {group: net.kyori, name: examination-api, version: &examination 1.1.0}
+  examination-string: {group: net.kyori, name: examination-string, version: *examination}
+  google-gson: "com.google.code.gson:gson:2.8.0"
+  jetbrainsAnnotations: "org.jetbrains:annotations:21.0.1"
+  kotlin-stdlib8: {group: "org.jetbrains.kotlin", name: "kotlin-stdlib-jdk8"}
+
+  # Checkstyle
+  stylecheck: "ca.stellardrift:stylecheck:0.1"
+
+  # Testing
+  google-guava: {group: com.google.guava, name: guava, version: &guava 30.1.1-jre}
+  google-guavaTest: {group: com.google.guava, name: guava-testlib, version: *guava}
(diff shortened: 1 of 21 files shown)


api/src/main/java/net/kyori/adventure/util/ShadyPines.java
```diff
@@ -77,4 +85,44 @@ public static boolean equals(final double a, final double b) {
...
+   *
+   * @param original the original map
+   * @param <K> key type
+   * @param <V> value type
+   * @return the immutable snapshot map
+   * @since 4.8.0
+   */
+  public static <K, V> @NonNull Map<K, V> immutableMap(final @NonNull Map<K, V> original) {
+    return Collections.unmodifiableMap(new HashMap<>(original));
+  }
```
kashike, line 127: Perhaps we should add a new class for all the collection-related methods?

api/src/main/java9/module-info.java
```diff
@@ -0,0 +1,47 @@
...
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+ * SOFTWARE.
+ */
+@SuppressWarnings({"requires-automatic", "requires-transitive-automatic"})
+module net.kyori.adventure {
+  requires transitive net.kyori.adventure.key;
+  requires transitive net.kyori.examination.api;
+  requires transitive net.kyori.examination.string;
+  requires static org.checkerframework.checker.qual;
+  requires static org.jetbrains.annotations;
```
kashike, line 30: Does it complain about automatic modules at all?

gradle/libs.versions.yaml
```diff
@@ -0,0 +1,27 @@
+metadata:
+  format: {version: 1.0}
+
+dependencies:
+  # Runtime libraries
+  checker: "org.checkerframework:checker-qual:3.12.0"
```
kashike, line 6:
```suggestion
  checker-qual: "org.checkerframework:checker-qual:3.12.0"
```

kashike (changes requested):

gradle/libs.versions.yaml
zml2008: this would produce an accessor of `libs.checker.qual` -- do you want that, or something like `libs.checkerQual`?

api/src/main/java9/module-info.java
zml2008: it gives a warning, that I think I've managed to suppress. Might have to suppress it with a compiler argument though.

api/src/main/java/net/kyori/adventure/util/ShadyPines.java
zml2008: could do yeah -- that'd make sense with another PR you have open at the moment I think?

api/src/main/java9/module-info.java
```diff
@@ -0,0 +1,47 @@
...
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+ * SOFTWARE.
+ */
+@SuppressWarnings({"requires-automatic", "requires-transitive-automatic"})
+module net.kyori.adventure {
+  requires transitive net.kyori.adventure.key;
+  requires transitive net.kyori.examination.api;
+  requires transitive net.kyori.examination.string;
+  requires static org.checkerframework.checker.qual;
```
A248, line 29: Shouldn't these directives be `static transitive`?

A248 (commented):
Very cool to see

kashike:
https://github.com/KyoriPowered/adventure/commit/077b23654ffff32f0bbc572a6067a0cd37cc6209

zml2008:
Very little of this is still relevant. What is will be broken out into separate PRs.

PR #339, opened 2021-04-20
api: Component iterator

kezz:
Following the discussion in https://github.com/KyoriPowered/adventure/pull/310#discussion_r599837951, this PR makes the `Component` class implement `Iterable<Component>`, allowing developers to iterate through a component and their children recursively. As component are immutable, the iterator does not support removal and it will throw an exception as per the definition of the class where removal isn't supported.

There are two types of iterator included, a depth first and a breadth first one. As the breadth first iterator is slightly more performant, this is the iterator type that is returned by the implementation of the `Iterable` class in the `Component` class. The other iterator can be obtained from `ComponentIterator.iterator(component, ComponentIterator.Type.DEPTH_FIRST)`. Additionally, this class also contains a method that returns an `Iterable` using a similar method, in order to allow for each looping over a component using the other iterator type.

### DIFF
--- a/api/src/main/java/net/kyori/adventure/text/Component.java
+++ b/api/src/main/java/net/kyori/adventure/text/Component.java
@@ -49,6 +49,7 @@ import net.kyori.adventure.translation.Translatable;
 import net.kyori.adventure.util.Buildable;
 import net.kyori.adventure.util.ForwardingIterator;
 import net.kyori.adventure.util.IntFunction2;
+import net.kyori.adventure.util.MonkeyBars;
 import net.kyori.examination.Examinable;
 import org.jetbrains.annotations.ApiStatus;
 import org.jetbrains.annotations.Contract;
@@ -2023,12 +2024,26 @@ public interface Component extends ComponentBuilderApplicable, ComponentLike, Ex
    * Returns an iterable view of this component.
    *
    * @param type the type
+   * @param flags the flags
    * @return the iterable
    * @since 4.9.0
    */
-  default @NotNull Iterable<Component> iterable(final @NotNull ComponentIteratorType type) {
+  default @NotNull Iterable<Component> iterable(final @NotNull ComponentIteratorType type, final @NotNull ComponentIteratorFlag@Nullable... flags) {
+    return this.iterable(type, flags == null ? Collections.emptySet() : MonkeyBars.enumSet(ComponentIteratorFlag.class, flags));
+  }
+
+  /**
+   * Returns an iterable view of this component.
+   *
+   * @param type the type
+   * @param flags the flags
+   * @return the iterable
+   * @since 4.9.0
+   */
+  default @NotNull Iterable<Component> iterable(final @NotNull ComponentIteratorType type, final @NotNull Set<ComponentIteratorFlag> flags) {
     Objects.requireNonNull(type, "type");
-    return new ForwardingIterator<>(() -> this.iterator(type), () -> this.spliterator(type));
+    Objects.requireNonNull(flags, "flags");
+    return new ForwardingIterator<>(() -> this.iterator(type, flags), () -> this.spliterator(type, flags));
   }
 
   /**
@@ -2036,12 +2051,27 @@ public interface Component extends ComponentBuilderApplicable, ComponentLike, Ex
    *
    * <p>As components are immutable, this iterator does not support removal.</p>
    *
-   * @param type the type of the iterator
+   * @param type the type
+   * @param flags the flags
    * @return the iterator
    * @since 4.9.0
    */
-  default @NotNull Iterator<Component> iterator(final @NotNull ComponentIteratorType type) {
-    return new ComponentIterator(this, type);
+  default @NotNull Iterator<Component> iterator(final @NotNull ComponentIteratorType type, final @NotNull ComponentIteratorFlag@Nullable... flags) {
+    return this.iterator(type, flags == null ? Collections.emptySet() : MonkeyBars.enumSet(ComponentIteratorFlag.class, flags));
+  }
+
+  /**
+   * Returns an iterator for this component.
+   *
+   * <p>As components are immutable, this iterator does not support removal.</p>
+   *
+   * @param type the type
+   * @param flags the flags
+   * @return the iterator
+   * @since 4.9.0
+   */
+  default @NotNull Iterator<Component> iterator(final @NotNull ComponentIteratorType type, final @NotNull Set<ComponentIteratorFlag> flags) {
+    return new ComponentIterator(this, Objects.requireNonNull(type, "type"), Objects.requireNonNull(flags, "flags"));
   }
 
   /**
@@ -2049,12 +2079,27 @@ public interface Component extends ComponentBuilderApplicable, ComponentLike, Ex
    *
    * <p>The resulting spliterator has the {@link Spliterator#IMMUTABLE}, {@link Spliterator#NONNULL} and {@link Spliterator#ORDERED} characteristics.</p>
    *
-   * @param type the type of the underlying iterator
-   * @return the iterator
+   * @param type the type
+   * @param flags the flags
+   * @return the spliterator
    * @since 4.9.0
    */
-  default @NotNull Spliterator<Component> spliterator(final @NotNull ComponentIteratorType type) {
-    return Spliterators.spliteratorUnknownSize(this.iterator(type), Spliterator.IMMUTABLE & Spliterator.NONNULL & Spliterator.ORDERED);
+  default @NotNull Spliterator<Component> spliterator(final @NotNull ComponentIteratorType type, final @NotNull ComponentIteratorFlag@Nullable... flags) {
+    return this.spliterator(type, flags == null ? Collections.emptySet() : MonkeyBars.enumSet(ComponentIteratorFlag.class, flags));
+  }
+
+  /**
+   * Returns a spliterator for this component.
+   *
+   * <p>The resulting spliterator has the {@link Spliterator#IMMUTABLE}, {@link Spliterator#NONNULL} and {@link Spliterator#ORDERED} characteristics.</p>
+   *
+   * @param type the type
+   * @param flags the flags
+   * @return the spliterator
+   * @since 4.9.0
+   */
+  default @NotNull Spliterator<Component> spliterator(final @NotNull ComponentIteratorType type, final @NotNull Set<ComponentIteratorFlag> flags) {
+    return Spliterators.spliteratorUnknownSize(this.iterator(type, flags), Spliterator.IMMUTABLE & Spliterator.NONNULL & Spliterator.ORDERED);
   }
 
   /**
--- a/api/src/main/java/net/kyori/adventure/text/Component.java
+++ b/api/src/main/java/net/kyori/adventure/text/Component.java
@@ -2024,8 +2024,7 @@ public interface Component extends ComponentBuilderApplicable, ComponentLike, Ex
    *
    * @param type the type
    * @return the iterable
-   * @see ComponentIteratorType
-   * @since 4.8.0
+   * @since 4.9.0
    */
   default @NotNull Iterable<Component> iterable(final @NotNull ComponentIteratorType type) {
     Objects.requireNonNull(type, "type");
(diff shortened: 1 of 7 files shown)


api/src/main/java/net/kyori/adventure/text/Component.java
```diff
@@ -1656,6 +1658,30 @@ default boolean hasStyling() {
   @Contract(pure = true)
   @NonNull Component replaceText(final @NonNull TextReplacementConfig config);
 
+  /**
+   * Returns a breadth-first iterator over this component and it's children.
+   * <p>As components are immutable, this iterator does not support removal.</p>
```
kashike, line 1663:
```suggestion
   * Returns a breadth-first iterator over this component and it's children.
   *
   * <p>As components are immutable, this iterator does not support removal.</p>
```

api/src/main/java/net/kyori/adventure/text/Component.java
```diff
@@ -1656,6 +1658,30 @@ default boolean hasStyling() {
...
+   * @since 4.8.0
+   */
+  @Override
+  default @NonNull Iterator<Component> iterator() {
+    return ComponentIterator.iterator(this, ComponentIterator.Type.BREADTH_FIRST);
+  }
+
+  /**
+   * Returns a spliterator from this component's iterator.
+   * <p>The resulting iterator has the {@link Spliterator#IMMUTABLE} and {@link Spliterator#NONNULL} characteristics.</p>
```
kashike, line 1675:
```suggestion
   * Returns a spliterator from this component's iterator.
   *
   * <p>The resulting iterator has the {@link Spliterator#IMMUTABLE} and {@link Spliterator#NONNULL} characteristics.</p>
```

api/src/main/java/net/kyori/adventure/text/Component.java
```diff
@@ -1656,6 +1658,30 @@ default boolean hasStyling() {
...
+   *
+   * @return the iterator
+   * @since 4.8.0
+   */
+  @Override
+  default @NonNull Spliterator<Component> spliterator() {
+    return Spliterators.spliteratorUnknownSize(this.iterator(), Spliterator.IMMUTABLE & Spliterator.NONNULL);
```
kashike, line 1682: Could probably use `ORDERED` too?

api/src/main/java/net/kyori/adventure/text/ComponentIterator.java
```diff
@@ -0,0 +1,126 @@
...
+import java.util.Deque;
+import java.util.Iterator;
+import java.util.List;
+import java.util.NoSuchElementException;
+import java.util.Objects;
+import org.checkerframework.checker.nullness.qual.NonNull;
+
+/**
+ * An iterator that traverses a component and it's children.
+ * <p>As components are immutable, this iterator does not support removal.</p>
```
kashike, line 36:
```suggestion
 * An iterator that traverses a component and it's children.
 *
 * <p>As components are immutable, this iterator does not support removal.</p>
```

api/src/main/java/net/kyori/adventure/text/ComponentIterator.java
```diff
@@ -0,0 +1,126 @@
...
+ *
+ * @see Component#iterator()
+ * @since 4.8.0
+ */
+public final class ComponentIterator implements Iterator<Component> {
```
kashike, line 41: Should be package-private, with `Component` having the way to create them.

api/src/main/java/net/kyori/adventure/text/ComponentIterator.java
```diff
@@ -0,0 +1,126 @@
...
+
+      final List<Component> children = next.children();
+      if(!children.isEmpty()) {
+        if(this.bfs) {
+          this.queue.addAll(next.children());
+        } else {
+          for(int i = children.size() - 1; i >= 0; i--) {
+            this.queue.addFirst(children.get(i));
+          }
+        }
```
kashike, line 95: Should probably extract this logic to each type.

kashike (changes requested):

api/src/main/java/net/kyori/adventure/text/ComponentIterator.java
kezz: I'm not sure about how I've implemented this now, with it being moved to an enum. Do you think it's worth completely extracting the types of iteration more than this?

api/src/main/java/net/kyori/adventure/util/DelegatingIterable.java
```diff
@@ -0,0 +1,63 @@
...
+import java.util.function.Supplier;
+import org.checkerframework.checker.nullness.qual.NonNull;
+
+/**
+ * An iterable that delegates the {@link #iterator()} and {@link #spliterator()} calls to a {@link Supplier}.
+ *
+ * @param <T> the type of the iterable
+ * @since 4.8.0
+ */
+public final class DelegatingIterable<T> implements Iterable<T> {
```
kashike, line 38: Not sure if this is the best name for this class?

api/src/main/java/net/kyori/adventure/util/DelegatingIterable.java
```diff
@@ -0,0 +1,63 @@
...
+  private final Supplier<Iterator<T>> iteratorSupplier;
+  private final Supplier<Spliterator<T>> spliteratorSupplier;
```
kashike, line 40: Can drop `Supplier` from the name of the fields + constructor parameters.

api/src/test/java/net/kyori/adventure/text/ComponentIteratorTest.java
```diff
@@ -0,0 +1,109 @@
...
+package net.kyori.adventure.text;
+
+import net.kyori.adventure.text.event.HoverEvent;
+import net.kyori.adventure.text.format.NamedTextColor;
+import org.junit.jupiter.api.Test;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.fail;
+
+public class ComponentIteratorTest {
```
kashike, line 33:
```suggestion
class ComponentIteratorTest {
```

kashike (changes requested):

api/src/main/java/net/kyori/adventure/util/DelegatingIterable.java
kezz: Perhaps `ForwardingIterable`?

kezz:
Just a note that these tests will fail as the hover event iterator isn't implemented correctly at the moment. Need to get some more testing done tomorrow.

kezz:
As far as I'm concerned this is now "done". However, I do have one remaining question.

For the `XXX_WITH_HOVER` iterator types, should the components in the hover event be considered children or siblings of the current component? I believe it makes sense to consider them as siblings of the component, meaning that regardless of if it's a breadth or depth first iteration the component will always get appended next. However, I'd be interested to hear others' thoughts on the matter.

api/src/main/java/net/kyori/adventure/text/Component.java
```diff
@@ -1656,6 +1660,76 @@ default boolean hasStyling() {
...
+   * @since 4.8.0
+   */
+  default @NonNull Iterable<Component> iterable(final @NonNull ComponentIteratorType type) {
+    Objects.requireNonNull(type, "type");
+    if(type == ComponentIteratorType.defaultType()) return this;
+    return new ForwardingIterator<>(() -> this.iterator(type), () -> this.spliterator(type));
+  }
+
+  /**
+   * Returns an iterator over this component and it's children using the default iterator type.
```
zml2008, line 1678:
```suggestion
   * Returns an iterator over this component and its children using the default iterator type.
```

api/src/main/java/net/kyori/adventure/text/Component.java
```diff
@@ -92,7 +96,7 @@
  * @since 4.0.0
  */
 @ApiStatus.NonExtendable
-public interface Component extends ComponentBuilderApplicable, ComponentLike, Examinable, HoverEventSource<Component> {
+public interface Component extends ComponentBuilderApplicable, ComponentLike, Examinable, HoverEventSource<Component>, Iterable<Component> {
```
zml2008, line 99: imo `Component` should not directly implement `Iterable` (at least, not from the API side) -- it's a tree, not a sequence. Having separate methods that request a view as an `Iterable` or `Iterator` makes it clear that the flat `Iterator` is one of many possible views for a component.

zml2008 (changes requested):

api/src/main/java/net/kyori/adventure/text/Component.java
kezz: Yup that makes sense actually. I'd almost err towards removing this alongside the methods that don't take an iteration type to force people to decide what sort of iteration they want to do.

lucko: I agree with @zml2008 - implementing `Iterable` seems inappropriate for Components

kezz:
Rebased to latest and removed the iterable implementation and default methods. I think forcing developers to choose an iteration type will prevent headaches in the future.

api/src/main/java/net/kyori/adventure/text/Component.java
```diff
@@ -108,7 +111,6 @@
    * @since 4.8.0
    */
   BiPredicate<? super Component, ? super Component> EQUALS_IDENTITY = (a, b) -> a == b;
-
```
kashike, line 111: ?

kezz: 🤨

api/src/main/java/net/kyori/adventure/text/Component.java
```diff
@@ -1872,6 +1874,47 @@ default boolean hasStyling() {
...
+   * <p>The resulting spliterator has the {@link Spliterator#IMMUTABLE}, {@link Spliterator#NONNULL} and {@link Spliterator#ORDERED} characteristics.</p>
+   *
+   * @param type the type of the underlying iterator
+   * @return the iterator
+   * @see ComponentIteratorType
+   * @since 4.8.0
+   */
+  default @NotNull Spliterator<Component> spliterator(final @NotNull ComponentIteratorType type) {
+    return Spliterators.spliteratorUnknownSize(this.iterator(type), Spliterator.IMMUTABLE & Spliterator.NONNULL & Spliterator.ORDERED);
+  }
```
kashike, line 1916: Do we want to move these methods to their own class instead, and access them statically?

kashike (changes requested):

api/src/main/java/net/kyori/adventure/text/Component.java
zml2008: I think that is unnecessary -- they're one-liners in here already.

zml2008 (approved):
This should be good to go as soon as the one whitespace diff in `Component` is cleaned up.

kezz:
It would be nice to be able to iterate through builders as well, to access a mutable state to make modifying whilst iterating easier.

kezz:
Just rebased this to 4.9.0 in addition to the since tag changes. One last question I have about this PR is should we be allowing users to specify their own iteration strategy or is using the enum enough? I can't think of any different iteration strategies but there's no reason why we can't just mock up an interface and use that instead of the enum.

zml2008:
I think the enum should be fine here

zml2008 (approved):

kezz (commented):
This PR is incorrectly missing the arguments of translatable components.

zml2008:
I'm really not sure if translatable component arguments should be included -- they're not really children until the component is actually rendered down, they're just associated with the component.

Perhaps it would be more appropriate to have including translatable components as an additional flag? Similar to hover events.

That brings up something else (that you definitely might have mentioned before) -- would it make sense to split the depth/breadth-first option from additional 'flags', like including events or translatable arguments?

The API for that would look something like:
```java

interface Component {
  Iterable<Component> iterable(final ComponentIteratorType type, final ComponentIteratorFlag... flags);
}
```

kezz:
Yup, I completely agree. A flag system would definitely be better than what we've got at the moment here.

PR #342, opened 2021-04-22
api: Add joinWithFinalSeparator

kezz:
This PR adds a sister method of `Component#join` which is functionally the same but with the final separator being separate from the rest. This allows for joining components into strings like `1, 2 or 3`.

Tests are also added and the current `join` method has been modified to just overload the current.

Open for feedback on the name as it feels a bit shit :grimacing:

### DIFF
--- a/api/src/main/java/net/kyori/adventure/text/Component.java
+++ b/api/src/main/java/net/kyori/adventure/text/Component.java
@@ -146,17 +146,57 @@ public interface Component extends ComponentBuilderApplicable, ComponentLike, Ex
    */
   @Contract(value = "_, _ -> new", pure = true)
   static @NonNull TextComponent join(final @NonNull ComponentLike separator, final Iterable<? extends ComponentLike> components) {
+    return joinWithFinalSeparator(separator, separator, components);
+  }
+
+  /**
+   * Joins {@code components} using {@code separator} with {@code finalSeparator} between the last two components.
+   *
+   * @param separator the separator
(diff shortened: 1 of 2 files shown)


api/src/main/java/net/kyori/adventure/text/Component.java
```diff
@@ -146,17 +146,57 @@
...
       return Component.empty();
     }
+
     final TextComponent.Builder builder = text();
-    while(it.hasNext()) {
-      builder.append(it.next());
-      if(it.hasNext()) {
-        builder.append(separator);
+    ComponentLike component = it.next();
+    while(true) {
```
kashike, line 184: Is there any other way to do this? This feels a bit ugly.

kezz: I can't *think* of any way to do it without a sized collection that we can iterate over with an indexed loop. I suppose we could make this take a list instead of an iterable? There might be another way to write it, perhaps iterating while the component variable isn't null and setting it to null when the iterator runs out of elements? Or perhaps just replacing this with `it.hasNext()`, but then there's duplicate calls to it which isn't a massive issue I suppose.

kezz:
Another idea, what if we make a `JoinConfiguration` object similar to the replacement configuration bit? This could also hold a prefix and suffix.

kashike:
Sure, you can try that.

kezz:
Superseded by #344.

PR #343, opened 2021-04-26
ComponentSerializer services

kashike:


api/src/main/java/net/kyori/adventure/util/Services0.java
```diff
@@ -0,0 +1,35 @@
...
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+ * SOFTWARE.
+ */
+package net.kyori.adventure.util;
+
+import java.util.ServiceLoader;
+
+final class Services0 {
```
octylFractal, line 28: This should probably be documented to describe its role. As-is, it just seems like a weird split, but I know it's for Bukkit et al.

text-serializer-legacy/src/main/java/net/kyori/adventure/text/serializer/legacy/LegacyComponentSerializer.java
```diff
@@ -261,4 +263,39 @@
...
+     * Provides a {@link LegacyComponentSerializer} using {@link #SECTION_CHAR}.
+     *
+     * @return a {@link LegacyComponentSerializer}
+     * @since 4.8.0
+     */
+    @ApiStatus.Internal
+    @NonNull LegacyComponentSerializer legacySection();
+
+    /**
+     * Completes the building process of {@link Builder}.
```
octylFractal, line 293: This is not called on completion, but initialization, unless you mean completion of the `Builder` constructor which is not at all clear.

text-serializer-plain/src/main/java/net/kyori/adventure/text/serializer/plain/PlainTextComponentSerializer.java
```diff
@@ -0,0 +1,130 @@
...
+     * Provides a {@link PlainTextComponentSerializer}.
+     *
+     * @return a {@link PlainTextComponentSerializer}
+     * @since 4.8.0
+     */
+    @ApiStatus.Internal
+    @NonNull PlainTextComponentSerializer plainTextSimple();
+
+    /**
+     * Completes the building process of {@link Builder}.
```
octylFractal, line 122: Same question about "completes" here.

octylFractal (changes requested):
unkewl

api/src/main/java/net/kyori/adventure/util/Services.java
```diff
@@ -0,0 +1,64 @@
...
+    final Iterator<P> it = loader.iterator();
+    while(it.hasNext()) {
+      try {
+        final P provider = it.next();
+        if(it.hasNext()) {
+          throw new IllegalStateException("Expected to find one " + type + " service provider, found multiple");
+        }
+        return Optional.of(provider);
+      } catch(final Throwable t) {
+        // ignored
```
A248, line 59: Would it be okay to not swallow exceptions here? Debugging is painstaking enough already. Not to mention that this defeats the point of the `IllegalStateException` thrown here.

text-serializer-legacy/src/main/java/net/kyori/adventure/text/serializer/legacy/LegacyComponentSerializerImpl.java
```diff
@@ -471,6 +488,7 @@ private void applyFullFormat() {
     private ComponentFlattener flattener = ComponentFlattener.basic();
 
     BuilderImpl() {
+      BUILDER.accept(this); // let service provider touch the builder before anybody else touches it
```
A248, line 491: Specifically this change could irreversibly affect behavior.

A248 (commented):
As I understand, this could change the way certain component serializers are configured, possibly in a way that is not reversible by the API user.

I assume this isn't considered a breaking change for the static helper methods - GsonComponentSerializer.gson(), colorDownsamplingGson(), LegacyComponentSerializer.legacySection(), legacyAmpersand(), and the like - because the exact configuration of those serializers is not specified. (Though perhaps this lack of exact requirements could be documented)

For LegacyComponentSerializer.builder(), however, the component service has the ability to touch the builder before anything happens to it. This could mean that certain options on the LegacyComponentSerializer could be set by the service, but not *unset* by the user. For example, `useUnusualXRepeatedCharacterHexFormat` has no method to undo it.

text-serializer-legacy/src/main/java/net/kyori/adventure/text/serializer/legacy/LegacyComponentSerializerImpl.java
kashike: This only provides defaults - anything can be changed by using the builder methods like normal.

kashike:
> I assume this isn't considered a breaking change for the static helper methods - GsonComponentSerializer.gson(), colorDownsamplingGson(), LegacyComponentSerializer.legacySection(), legacyAmpersand(), and the like - because the exact configuration of those serializers is not specified. (Though perhaps this lack of exact requirements could be documented)
> 

Correct - not considered a breaking change.

> For LegacyComponentSerializer.builder(), however, the component service has the ability to touch the builder before anything happens to it. This could mean that certain options on the LegacyComponentSerializer could be set by the service, but not _unset_ by the user. For example, `useUnusualXRepeatedCharacterHexFormat` has no method to undo it.

This is true - we could introduce a method to allow this. But as is currently, no platforms do anything except providing the default `ComponentFlattener` to legacy and plain-text serializers.

zml2008 (approved):

PR #344, opened 2021-04-26
api: Advanced Component joining

kezz:
This PR adds a new class, `JoinConfiguration`, which introduces a more advanced method of joining components. Essentially, this class would allow you to create a component like `Online players: Bob, John, Harry and Martin.` from a simple collection of names with ease. A join configuration contains a series of component parts:

- A separator to go between each component,
- A final separator to go between the final two components,
- A prefix to be prepended to the result,
- A suffix to be appended to the result, and
- An convertor to change each joined `ComponentLike` into a `Component`.
- A predicate to determine if each component is to be included.

There are also some unit tests to ensure everything works correctly.

### DIFF
--- a/api/src/main/java/net/kyori/adventure/text/Component.java
+++ b/api/src/main/java/net/kyori/adventure/text/Component.java
@@ -145,7 +145,7 @@ public interface Component extends ComponentBuilderApplicable, ComponentLike, Ex
    * @param components the components
    * @return a text component
    * @since 4.0.0
-   * @deprecated for removal since 4.8.0, use {@link #join(JoinConfiguration, ComponentLike...)} with a {@link JoinConfiguration} instead.
+   * @deprecated for removal since 4.9.0, use {@link #join(JoinConfiguration, ComponentLike...)} with a {@link JoinConfiguration} instead.
    */
   @Deprecated
   @Contract(value = "_, _ -> new", pure = true)
@@ -160,7 +160,7 @@ public interface Component extends ComponentBuilderApplicable, ComponentLike, Ex
    * @param components the components
    * @return a text component
    * @since 4.0.0
-   * @deprecated for removal since 4.8.0, use {@link #join(JoinConfiguration, Iterable)} with a {@link JoinConfiguration} instead.
+   * @deprecated for removal since 4.9.0, use {@link #join(JoinConfiguration, Iterable)} with a {@link JoinConfiguration} instead.
    */
   @Deprecated
   @Contract(value = "_, _ -> new", pure = true)
@@ -178,7 +178,7 @@ public interface Component extends ComponentBuilderApplicable, ComponentLike, Ex
    * @param components the components
    * @return a text component
    * @see JoinConfiguration
-   * @since 4.8.0
+   * @since 4.9.0
    */
   @Contract(pure = true)
   static @NotNull Component join(final @NotNull JoinConfiguration config, final @NotNull ComponentLike @NotNull... components) {
@@ -192,7 +192,7 @@ public interface Component extends ComponentBuilderApplicable, ComponentLike, Ex
    * @param components the components
    * @return a text component
    * @see JoinConfiguration
-   * @since 4.8.0
+   * @since 4.9.0
    */
   @Contract(pure = true)
   static @NotNull Component join(final @NotNull JoinConfiguration config, final @NotNull Iterable<? extends ComponentLike> components) {
--- a/api/src/main/java/net/kyori/adventure/text/Component.java
+++ b/api/src/main/java/net/kyori/adventure/text/Component.java
@@ -146,10 +146,12 @@ public interface Component extends ComponentBuilderApplicable, ComponentLike, Ex
    * @param components the components
    * @return a text component
    * @since 4.0.0
+   * @deprecated for removal since 4.8.0, use {@link #join(JoinConfig, ComponentLike...)} with a {@link JoinConfig} instead.
    */
+  @Deprecated
   @Contract(value = "_, _ -> new", pure = true)
   static @NotNull TextComponent join(final @NotNull ComponentLike separator, final @NotNull ComponentLike@NotNull... components) {
-    return join(separator, Arrays.asList(components));
+    return join(JoinConfig.separator(separator.asComponent()), components);
   }
 
   /**
@@ -159,20 +161,74 @@ public interface Component extends ComponentBuilderApplicable, ComponentLike, Ex
    * @param components the components
    * @return a text component
    * @since 4.0.0
+   * @deprecated for removal since 4.8.0, use {@link #join(JoinConfig, Iterable)} with a {@link JoinConfig} instead.
    */
+  @Deprecated
   @Contract(value = "_, _ -> new", pure = true)
   static @NotNull TextComponent join(final @NotNull ComponentLike separator, final Iterable<? extends ComponentLike> components) {
+    return join(JoinConfig.separator(separator.asComponent()), components);
+  }
+
+  /**
+   * Joins {@code components} using the configuration in {@code config}.
+   *
+   * @param config the join configuration
+   * @param components the components
+   * @return a text component
+   * @see JoinConfig
+   * @since 4.8.0
+   */
+  @Contract(value = "_, _ -> new", pure = true)
+  static @NonNull TextComponent join(final @NonNull JoinConfig config, final @NonNull ComponentLike@NonNull... components) {
+    return join(config, Arrays.asList(components));
+  }
+
+  /**
+   * Joins {@code components} using the configuration in {@code config}.
+   *
+   * @param config the join configuration
+   * @param components the components
+   * @return a text component
+   * @see JoinConfig
+   * @since 4.8.0
+   */
+  @Contract(value = "_, _ -> new", pure = true)
+  static @NonNull TextComponent join(final @NonNull JoinConfig config, final @NonNull Iterable<? extends ComponentLike> components) {
     final Iterator<? extends ComponentLike> it = components.iterator();
+
     if (!it.hasNext()) {
-      return Component.empty();
+      if(config.prefix() == null && config.suffix() == null) return Component.empty();
+
+      final TextComponent.Builder builder = text();
+      if(config.prefix() != null) builder.append(config.prefix());
+      if(config.suffix() != null) builder.append(config.suffix());
+      return builder.build();
     }
+
     final TextComponent.Builder builder = text();
-    while (it.hasNext()) {
-      builder.append(it.next());
-      if (it.hasNext()) {
-        builder.append(separator);
+    if(config.prefix() != null) builder.append(config.prefix());
+
+    final boolean hasSeparator = config.separator() != null;
+    final ComponentLike lastSeparator = config.lastSeparator() == null ? config.separator() : config.lastSeparator();
+
+    ComponentLike component = it.next();
+    while (component != null) {
+      builder.append(config.operator().apply(component));
+
+      if(!it.hasNext()) {
+        component = null;
+      } else {
+        component = it.next();
+
+        if (it.hasNext()) {
+          if(hasSeparator) builder.append(config.separator());
+        } else {
+          if(lastSeparator != null) builder.append(lastSeparator);
+        }
       }
     }
+
+    if(config.suffix() != null) builder.append(config.suffix());
     return builder.build();
   }
 
--- a/api/src/main/java/net/kyori/adventure/text/Component.java
+++ b/api/src/main/java/net/kyori/adventure/text/Component.java
@@ -25,7 +25,6 @@ package net.kyori.adventure.text;
 
 import java.util.Arrays;
 import java.util.Collections;
-import java.util.Iterator;
 import java.util.List;
 import java.util.Map;
 import java.util.Objects;
@@ -168,7 +167,7 @@ public interface Component extends ComponentBuilderApplicable, ComponentLike, Ex
   static @NotNull TextComponent join(final @NotNull ComponentLike separator, final Iterable<? extends ComponentLike> components) {
     final Component component = join(JoinConfiguration.separator(separator), components);
 
-    if(component instanceof TextComponent) return (TextComponent) component;
+    if (component instanceof TextComponent) return (TextComponent) component;
     return Component.text().append(component).build();
   }
 
@@ -182,7 +181,7 @@ public interface Component extends ComponentBuilderApplicable, ComponentLike, Ex
    * @since 4.8.0
    */
   @Contract(pure = true)
-  static @NonNull Component join(final @NonNull JoinConfiguration config, final @NonNull ComponentLike@NonNull... components) {
+  static @NotNull Component join(final @NotNull JoinConfiguration config, final @NotNull ComponentLike @NotNull... components) {
     return join(config, Arrays.asList(components));
   }
 
@@ -196,57 +195,8 @@ public interface Component extends ComponentBuilderApplicable, ComponentLike, Ex
    * @since 4.8.0
    */
   @Contract(pure = true)
-  static @NonNull Component join(final @NonNull JoinConfiguration config, final @NonNull Iterable<? extends ComponentLike> components) {
-    final Iterator<? extends ComponentLike> it = components.iterator();
-    final Component prefix = config.prefix();
-    final Component suffix = config.suffix();
-    final UnaryOperator<Component> operator = config.operator();
-
-    if (!it.hasNext()) {
-      if(prefix == null && suffix == null) return Component.empty();
-
-      final TextComponent.Builder builder = text();
-      if(prefix != null) builder.append(prefix);
-      if(suffix != null) builder.append(suffix);
-      return builder.build();
-    }
-
-    ComponentLike component = it.next();
-    int componentsSeen = 0;
-
-    if(!it.hasNext() && prefix == null && suffix == null) return operator.apply(component.asComponent());
-
-    final Component separator = config.separator();
-    final boolean hasSeparator = separator != null;
-
-    final TextComponent.Builder builder = text();
-    if(prefix != null) builder.append(prefix);
-
-    while (component != null) {
-      builder.append(operator.apply(component.asComponent()));
-      componentsSeen++;
-
-      if(!it.hasNext()) {
-        component = null;
-      } else {
-        component = it.next();
-
-        if (it.hasNext()) {
-          if(hasSeparator) builder.append(separator);
-        } else {
-          Component lastSeparator = null;
-
-          if(componentsSeen > 1) lastSeparator = config.lastSeparatorIfSerial();
-          if(lastSeparator == null) lastSeparator = config.lastSeparator();
-          if(lastSeparator == null) lastSeparator = config.separator();
-
-          if(lastSeparator != null) builder.append(lastSeparator);
-        }
-      }
-    }
-
-    if(suffix != null) builder.append(suffix);
-    return builder.build();
+  static @NotNull Component join(final @NotNull JoinConfiguration config, final @NotNull Iterable<? extends ComponentLike> components) {
+    return JoinConfigurationImpl.join(config, components);
   }
 
   /**
--- a/api/src/main/java/net/kyori/adventure/text/Component.java
+++ b/api/src/main/java/net/kyori/adventure/text/Component.java
@@ -195,25 +195,32 @@ public interface Component extends ComponentBuilderApplicable, ComponentLike, Ex
   @Contract(value = "_, _ -> new", pure = true)
   static @NonNull TextComponent join(final @NonNull JoinConfiguration config, final @NonNull Iterable<? extends ComponentLike> components) {
     final Iterator<? extends ComponentLike> it = components.iterator();
+    final ComponentLike prefix = config.prefix();
+    final ComponentLike suffix = config.suffix();
 
     if (!it.hasNext()) {
-      if(config.prefix() == null && config.suffix() == null) return Component.empty();
+      if(prefix == null && suffix == null) return Component.empty();
 
       final TextComponent.Builder builder = text();
-      if(config.prefix() != null) builder.append(config.prefix());
-      if(config.suffix() != null) builder.append(config.suffix());
+      if(prefix != null) builder.append(prefix);
+      if(suffix != null) builder.append(suffix);
       return builder.build();
     }
 
     final TextComponent.Builder builder = text();
-    if(config.prefix() != null) builder.append(config.prefix());
+    if(prefix != null) builder.append(prefix);
 
-    final boolean hasSeparator = config.separator() != null;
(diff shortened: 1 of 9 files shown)


api/src/main/java/net/kyori/adventure/text/JoinConfig.java
```diff
@@ -0,0 +1,258 @@
...
+ * This defaults to an identity operator.</p>
+ *
+ * <p>Note that if the final separator is omitted, the normal separator will be used instead.
+ * To omit the final separator, but still include normal separators, use {@link Component#empty()}.</p>
+ *
+ * @see Component#join(JoinConfig, Iterable)
+ * @see Component#join(JoinConfig, ComponentLike...)
+ * @since 4.8.0
+ */
+public interface JoinConfig extends Buildable<JoinConfig, JoinConfig.Builder>, Examinable {
```
kashike, line 55:
```suggestion
public interface JoinConfiguration extends Buildable<JoinConfig, JoinConfig.Builder>, Examinable {
```

kashike (changes requested):

api/src/test/java/net/kyori/adventure/text/JoinTest.java
```diff
@@ -0,0 +1,159 @@
...
+package net.kyori.adventure.text;
+
+import java.util.Collections;
+import java.util.stream.IntStream;
+import net.kyori.adventure.text.format.NamedTextColor;
+import org.junit.jupiter.api.Test;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+
+public class JoinTest {
```
kashike, line 33:
```suggestion
class JoinTest {
```

api/src/main/java/net/kyori/adventure/text/JoinConfig.java
```diff
@@ -0,0 +1,258 @@
...
+   * @since 4.8.0
+   */
+  interface Builder extends Buildable.Builder<JoinConfig> {
+    /**
+     * Gets the prefix of this join configuration builder.
+     *
+     * @return the prefix
+     * @since 4.8.0
+     */
+    @Nullable ComponentLike prefix();
```
kashike, line 179: No getters in builders.

kashike (changes requested):

kezz:
Just made all those requested changes in addition to cleaning up the Javadoc a bit. I just have two further questions:

1. Could we replace the dl/dt/dd tags with just a normal unordered list? I know these tags are used elsewhere for similar purposes but they render terribly on web browsers and don't look very nice imo.
2. Is it okay that the use of the full word in `JoinConfiguration` conflicts with the shortened use of it elsewhere? For example, `TextReplacementConfig`.

api/src/main/java/net/kyori/adventure/text/Component.java
```diff
@@ -143,20 +145,74 @@
...
+   * @see JoinConfiguration
+   * @since 4.8.0
+   */
+  @Contract(value = "_, _ -> new", pure = true)
+  static @NonNull TextComponent join(final @NonNull JoinConfiguration config, final @NonNull Iterable<? extends ComponentLike> components) {
     final Iterator<? extends ComponentLike> it = components.iterator();
+
     if(!it.hasNext()) {
-      return Component.empty();
+      if(config.prefix() == null && config.suffix() == null) return Component.empty();
```
kashike, line 184: Extract `config.prefix(), `config.suffix()`, and others to local variables.

api/src/main/java/net/kyori/adventure/text/JoinConfiguration.java
```diff
@@ -0,0 +1,222 @@
...
+   */
+  @Nullable ComponentLike lastSeparator();
+
+  /**
+   * Gets the operator of this join configuration.
+   *
+   * @return the operator
+   * @since 4.8.0
+   */
+  @NotNull UnaryOperator<ComponentLike> operator();
```
kashike, line 143: Should this be `UnaryOperator<? super ComponentLike>`?

api/src/main/java/net/kyori/adventure/text/JoinConfiguration.java
```diff
@@ -0,0 +1,222 @@
...
+   *
+   * @param components the components
+   * @return a text component
+   * @see Component#join(JoinConfiguration, Iterable)
+   * @since 4.8.0
+   */
+  @Contract(value = "_ -> new", pure = true)
+  default @NonNull TextComponent join(final @NonNull Iterable<? extends ComponentLike> components) {
+    return Component.join(this, components);
+  }
```
kashike, line 169: Not sure I like these being here.

api/src/test/java/net/kyori/adventure/text/JoinTest.java
```diff
@@ -0,0 +1,179 @@
...
+
+import java.util.Collections;
+import java.util.stream.IntStream;
+import net.kyori.adventure.text.format.NamedTextColor;
+import org.junit.jupiter.api.Test;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+
+class JoinTest {
+
```
kashike, line 34:
```suggestion
```suggestion
```
```

File: api/src/main/java/net/kyori/adventure/text/JoinConfiguration.java (hunk 5 above)
Comment (kezz) [reply api/src/main/java/net/kyori/adventure/text/JoinConfiguration.java:143]: I'm not sure - what would adding this do?

File: api/src/main/java/net/kyori/adventure/text/JoinConfiguration.java [hunk 8]
```diff
@@ -0,0 +1,222 @@
...
+   */
+  @Nullable ComponentLike prefix();
+
+  /**
+   * Gets the suffix of this join configuration.
+   *
+   * @return the suffix
+   * @since 4.8.0
+   */
+  @Nullable ComponentLike suffix();
```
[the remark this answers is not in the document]
kashike: Should probably return `Component`, yes.

BomBardyGamer (approved):
Looks great to me, would be very usefulness to be able to join components very easily. I think it might be useful though to be able to configure this from just a plain string as well as from component like objects, instead of having to construct a text component every time.

kezz:
Changes since last commit:
- Support for creating lists separated by an Oxford comma for all you weirdos that use it
- Javadocs have been improved to look less ugly
- All methods now return `Component` instead of `ComponentLike`

---

> Looks great to me, would be very usefulness to be able to join components very easily. I think it might be useful though to be able to configure this from just a plain string as well as from component like objects, instead of having to construct a text component every time.

Yeah I'm not sure about this - there's no place in the API that accepts components that also accepts methods to "easily" create components from strings, etc.

api/src/main/java/net/kyori/adventure/text/JoinConfiguration.java
```diff
@@ -0,0 +1,300 @@
...
+  @NotNull UnaryOperator<Component> operator();
+
+  /**
+   * Gets the last separator that will be used instead of the normal last separator in the case where there
+   * are more than two components being joined. This can be used to mimic a serial (or Oxford) comma.
+   *
+   * @return the separator
+   * @since 4.8.0
+   */
+  @Nullable Component lastSeparatorIfSerial();
```
kashike, line 167: Move above `operator` method.

api/src/main/java/net/kyori/adventure/text/JoinConfiguration.java
```diff
@@ -0,0 +1,300 @@
...
+  interface Builder extends Buildable.Builder<JoinConfiguration> {
+    /**
+     * Sets the prefix of this join configuration builder.
+     *
+     * @param prefix the prefix
+     * @return this builder
+     * @since 4.8.0
+     */
+    @Contract("_ -> this")
+    @NonNull Builder prefix(final @Nullable Component prefix);
```
kashike, line 183: I don't think we need this method (or others that accept `Component`) in here - `ComponentLike` is good enough, just do the converting to `Component` in the implementation.

kashike (changes requested):

api/src/main/java/net/kyori/adventure/text/Component.java
```diff
@@ -130,10 +130,12 @@
    * @param components the components
    * @return a text component
    * @since 4.0.0
+   * @deprecated for removal since 4.8.0, use {@link #join(JoinConfiguration, ComponentLike...)} with a {@link JoinConfiguration} instead.
    */
+  @Deprecated
```
astei, line 135: I don't think we should deprecate this method - it is perfectly fine as a convenient shorthand.

api/src/main/java/net/kyori/adventure/text/Component.java
```diff
@@ -143,20 +145,86 @@
    * @param components the components
    * @return a text component
    * @since 4.0.0
+   * @deprecated for removal since 4.8.0, use {@link #join(JoinConfiguration, Iterable)} with a {@link JoinConfiguration} instead.
    */
+  @Deprecated
```
astei, line 150: Ditto

api/src/main/java/net/kyori/adventure/text/Component.java
```diff
@@ -143,20 +145,86 @@
...
+   * Joins {@code components} using the configuration in {@code config}.
+   *
+   * @param config the join configuration
+   * @param components the components
+   * @return a text component
+   * @see JoinConfiguration
+   * @since 4.8.0
+   */
+  @Contract(value = "_, _ -> new", pure = true)
+  static @NonNull TextComponent join(final @NonNull JoinConfiguration config, final @NonNull Iterable<? extends ComponentLike> components) {
```
astei, line 180: Would this API be more idiomatic as a builder, or a modifiable (albeit immutable) `ComponentJoiner` object?

```java
  private static final ComponentJoiner PLUGIN_JOINER = Component.joiner(Component.text(", "))
    .prefix(prefix)
    .suffix(suffix);

  // ...

  public static void sendPluginList(Audience target) {
    target.sendMessage(PLUGIN_JOINER.join(getPluginNameComponents());
  }
```

astei (changes requested):

api/src/main/java/net/kyori/adventure/text/Component.java
astei: There is precedent here: see Guava `Joiner` and `Splitter` which uses the pattern I've described here.

api/src/main/java/net/kyori/adventure/text/Component.java
kashike: Text replacement methods got the same treatment - deprecated in favour of configuration. It is easier to maintain `JoinConfiguration` and not worrying about updating things here too.

api/src/main/java/net/kyori/adventure/text/Component.java
kashike: This PR is matching how we do this in the rest of the Adventure API, however.

astei: I'm still not a big fan of it. I guess it's something that can be revisited in Adventure 5, because I still believe this pattern feels unnatural to write.

kezz: I do agree that a shorthand method to join directly from a join configuration would be nice, an earlier version of the PR did include one. Although I do understand that the syntax doesn't differ too much whether you're doing `config.join(components)` or `join(config, components)` and the latter is certainly more in keeping with the rest of the API, such as the text replacement system.

api/src/main/java/net/kyori/adventure/text/JoinConfiguration.java
kezz: Speaking of, can anyone think of a better name for this? The current one is not very descriptive at all but the descriptive alternative `lastSeparatorIfThreeOrMoreComponentsAreBeingJoined()` is much, much worse.

kezz:
Remaining questions and thoughts I have about this PR:
- Should an additional static method of creation for the configuration be made that includes the `lastSeparatorIfSerial` parameter (i.e. `JoinConfiguration#separators(ComponentLike separator, ComponentLike lastSeparator, ComponentLike lastSeparatorIfSerial)`.
- Should these static methods just be completely dropped in favour of the builder? The equivalent `TextReplacementConfig` doesn't have any static methods like this.
- Is `JoinConfiguration` a good name or should it be `JoinConfig` in keeping with `TextReplacementConfig`?
- Should a utility method be created to allow `config.join(components)` to work?
- Can anyone think of a better name than `lastSeparatorIfSerial`?
- Do the most recent changes to the join system make sense? I believe it would be logical that if you put in a single component with no prefix or suffix that you would receive the exact component that you supplied. This does mean that the old join methods need a bit of hackery to wrap them up in a text component if the new join methods didn't return anything. This might seem messier, but imo it is a lot more logical and prevents the addition of unneeded bulk.

api/src/main/java/net/kyori/adventure/text/Component.java
zml2008: This brings up a good point -- perhaps we should un-deprecate `replaceText` methods for the most simple case of Pattern -> Component, as a convenience option.

api/src/main/java/net/kyori/adventure/text/JoinConfiguration.java
```diff
@@ -0,0 +1,236 @@
...
+
+    /**
+     * Sets the operator of this join configuration builder.
+     *
+     * @param operator the operator
+     * @return this builder
+     * @since 4.8.0
+     */
+    @Contract("_ -> this")
+    @NonNull Builder operator(final @NotNull UnaryOperator<Component> operator);
```
zml2008, line 234: Can you clarify that the operator does not receive the `prefix`, `suffix` or any of the `separator`s?

api/src/main/java/net/kyori/adventure/text/Component.java
```diff
@@ -143,20 +145,91 @@
...
+   * @param config the join configuration
+   * @param components the components
+   * @return a text component
+   * @see JoinConfiguration
+   * @since 4.8.0
+   */
+  @Contract(pure = true)
+  static @NonNull Component join(final @NonNull JoinConfiguration config, final @NonNull Iterable<? extends ComponentLike> components) {
     final Iterator<? extends ComponentLike> it = components.iterator();
+    final Component prefix = config.prefix();
```
zml2008, line 185: This logic is becoming more complex. What about moving the implementation into `JoinConfigurationImpl`?

zml2008 (changes requested):

lucko:
Looks good but I agree with @astei re: the deprecations

api/src/main/java/net/kyori/adventure/text/Component.java
kezz: Another solution is to just add convenience methods to `TextReplacementConfig` similar to the `separator` methods in `JoinConfiguration`, for example.

kezz:
Rebased and applied zml's suggestions.

As for the deprecation of the "shorthand" methods, I believe this PR should be merged with the deprecations in place, and the removal of the deprecations for this and the `replaceText` methods should be done together in another PR to avoid inconsistencies and to provide a more pertinent place for discussion.

api/src/main/java/net/kyori/adventure/text/Component.java
zml2008: hm, I think if we were doing this from the start that would be good, but in the interest of minimizing API churn, I think maintaining the older methods would be the better way to go for both this and replacement.

zml2008:
Looking back at some of your comments:

- `JoinConfig` is probably a better name
- The static methods on `JoinConfig` should be dropped, in favor of the existing methods on `Component`
- Why return `TextComponent` rather than just `Component`? The `content` of the result will always be empty, right?

kezz:
> * `JoinConfig` is probably a better name

Agreed, it's definitely more in keeping with `TextReplacementConfig`.

> * The static methods on `JoinConfig` should be dropped, in favor of the existing methods on `Component`

Do you mean the methods like `separators()` and `noSeparators()`? Or the previously added `join` methods that have since been removed.

> * Why return `TextComponent` rather than just `Component`? The `content` of the result will always be empty, right?

Yes, the return here is only done because that's what the old methods returned. `TextComponent` is never returned for any of the new methods added in this PR.

zml2008:
> Do you mean the methods like separators() and noSeparators()? Or the previously added join methods that have since been removed.

Yeah, I think those last two should be removed too.

> Yes, the return here is only done because that's what the old methods returned. TextComponent is never returned for any of the new methods added in this PR.

oh oops, nvm that then

api/src/main/java/net/kyori/adventure/text/JoinConfiguration.java
```diff
@@ -0,0 +1,239 @@
...
+   * Gets the last separator that will be used instead of the normal last separator in the case where there
+   * are more than two components being joined. This can be used to mimic a serial (or Oxford) comma.
+   *
+   * @return the separator
+   * @since 4.8.0
+   */
+  @Nullable Component lastSeparatorIfSerial();
+
+  /**
+   * Gets the operator of this join configuration.]
```
kezz, line 161:
```suggestion
   * Gets the operator of this join configuration.
```

kezz:
Just rebased this for a 4.9.0 release. I still think the questions regarding deprecation are valid, but I do still think they are better served in another PR that would undeprecate both sets of methods or one that provides shorthand methods for `TextReplacementConfig`.

api/src/main/java/net/kyori/adventure/text/Component.java
zml2008: Thinking about this again, and with #418 I think replacing all these methods with one that takes a `JoinConfiguration` is the least awkward way forward.

api/src/main/java/net/kyori/adventure/text/Component.java
```diff
@@ -159,21 +160,43 @@
...
+    return Component.text().append(component).build();
+  }
+
+  /**
+   * Joins {@code components} using the configuration in {@code config}.
+   *
+   * @param config the join configuration
+   * @param components the components
+   * @return a text component
+   * @see JoinConfiguration
```
zml2008, line 180: add @see for #noSeparators, #separator, and #separators here, replacing the one for `JoinConfiguration`.

zml2008 (changes requested):
Can you deprecate `TextComponent.ofChildren` as well, to be replaced by `join(JoinConfiguration.noSeparator())`?

kezz:
Thoughts on a predicate to include/exclude components from being joined?

api/src/main/java/net/kyori/adventure/text/JoinConfigurationImpl.java
```diff
@@ -0,0 +1,234 @@
...
+      ExaminableProperty.of("operator", this.operator)
+    );
+  }
+
+  @Override
+  public String toString() {
+    return this.examine(StringExaminer.simpleEscaping());
+  }
+
+  @Contract(pure = true)
```
kezz, line 115: Unneeded annotation.

api/src/main/java/net/kyori/adventure/text/JoinConfigurationImpl.java
```diff
@@ -0,0 +1,234 @@
...
+    if (!it.hasNext()) {
+      if (prefix == null && suffix == null) return Component.empty();
+
+      final TextComponent.Builder builder = Component.text();
+      if (prefix != null) builder.append(prefix);
+      if (suffix != null) builder.append(suffix);
+      return builder.build();
+    }
+
+    ComponentLike component = it.next();
```
kezz, line 134: We should be ensuring that this component is non-null, or documenting that we allow non-null elements and handling that.

api/src/main/java/net/kyori/adventure/text/JoinConfigurationImpl.java
```diff
@@ -0,0 +1,234 @@
...
+    if (!it.hasNext() && prefix == null && suffix == null) return operator.apply(component.asComponent());
+
+    final Component separator = config.separator();
+    final boolean hasSeparator = separator != null;
+
+    final TextComponent.Builder builder = Component.text();
+    if (prefix != null) builder.append(prefix);
+
+    while (component != null) {
+      builder.append(operator.apply(component.asComponent()));
```
kezz, line 146: Similarly, we should be asserting that this isn't null or documenting that we allow it and handling it.

text-serializer-legacy/src/main/java/net/kyori/adventure/text/serializer/legacy/LegacyComponentSerializerImpl.java
```diff
@@ -234,7 +234,7 @@ private TextComponent extractUrl(final TextComponent component) {
     if (this.urlReplacementConfig == null) return component;
     final Component newComponent = component.replaceText(this.urlReplacementConfig);
     if (newComponent instanceof TextComponent) return (TextComponent) newComponent;
-    return TextComponent.ofChildren(newComponent);
+    return Component.text().append(newComponent).build();
```
zml2008, line 237: Is there a reason you didn't use the replacement recommended in the JD for `ofChildren`? If that issue exists here, what does that say for other users of the API? If it's not good enough for us, why should we ask others to migrate?

kezz: Well, the annoying part about this methods and others like it is that they return `TextComponent` instead of `Component`. So in this situation I have to choice of using the new join method and then wrapping it in a text component (like I have done elsewhere) or just appending it as one child. In hindsight, this should be using an `instanceof` check before the wrapping is done as in a lot of cases this component will already be a `TextComponent`.

api/src/main/java/net/kyori/adventure/text/JoinConfigurationImpl.java
```diff
@@ -164,7 +163,7 @@ public String toString() {
         }
       }
 
-      builder.append(convertor.apply(component));
+      builder.append(Objects.requireNonNull(convertor.apply(component), "Null output from \"convertor\" is not allowed"));
```
zml2008, line 166: `convertor` or `converter`?

zml2008: turns out it's either! english is great!

zml2008 (approved):

api/src/main/java/net/kyori/adventure/text/JoinConfiguration.java
```diff
@@ -0,0 +1,269 @@
...
+     * @param lastSeparator the last separator
+     * @return this builder
+     * @since 4.9.0
+     */
+    @Contract("_ -> this")
+    @NotNull Builder lastSeparator(final @Nullable ComponentLike lastSeparator);
+
+    /**
+     * Sets the last separator that will be used instead of the normal last separator in the case where there
+     * are more than two components being joined. This can be used to mimic a serial (or Oxford) comma.
```
zml2008, line 236:
```suggestion
     * are more than two components being joined. 
     * 
     * <p>This can be used to mimic a serial (or Oxford) comma.</p>
```

PR #354, opened 2021-05-06
text-serializer-gson: Add color stripping serializer

kezz:
This PR adds an additional way of serializing components that strips them of all color.

In order to facilitate this, a new enum is created that contains the different ways of processing colors.


kashike:
Err... why, exactly?

kezz:
> Err... why, exactly?

Essentially as a way to fix MC-64867 or to implement the feature in other software.

zml2008:
I really think this is something better done as a pre-processing step, rather than as part of the component serializer. Downsampling was only really put in the serializer for backwards compatibility, and because adventure was written before 1.16 was released.

kezz:
Actually, the more I think about this, the more I agree. This should be a serialiser-agnostic step in the rendering/serialising process, especially if/when JSON serialises are created, it becomes a pain to have to implement this in multiple places rather than duplicating this logic in multiple places.

PR #355, opened 2021-05-06
Component serializer typo

kezz:
My biggest PR yet...

### DIFF
--- a/api/src/main/java/net/kyori/adventure/text/serializer/ComponentSerializer.java
+++ b/api/src/main/java/net/kyori/adventure/text/serializer/ComponentSerializer.java
@@ -27,6 +27,7 @@ import net.kyori.adventure.text.Component;
 import org.checkerframework.checker.nullness.qual.NonNull;
 import org.checkerframework.checker.nullness.qual.Nullable;
 import org.checkerframework.checker.nullness.qual.PolyNull;
+import org.jetbrains.annotations.ApiStatus;
 import org.jetbrains.annotations.Contract;
 
 /**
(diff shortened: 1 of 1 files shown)


api/src/main/java/net/kyori/adventure/text/serializer/ComponentSerializer.java
```diff
@@ -55,9 +55,25 @@
    * @param input the input
    * @return the component if {@code input} is non-null, otherwise {@code null}
    * @since 4.7.0
+   * @deprecated Use the correctly spelt {@link #deserializeOrNull(Object)}.
```
kashike, line 58:
```suggestion
   * @deprecated for removal since 4.8.0, use {@link #deserializeOrNull(Object)} instead
```

kashike (changes requested):

api/src/main/java/net/kyori/adventure/text/serializer/ComponentSerializer.java
kezz: With `forRemoval = true` in the annotation as well?

kashike: No, we are not on Java 9+.

kezz: Added! Although it does feel weird having `@Contract` at the top :')

kashike (approved):

kashike (approved):

PR #364, opened 2021-05-12
api: Add contains method that allows alternative methods of equality comparison

### LINKED ISSUE #363: Component contains check doesn't check identical but separately created components
The following test will fail despite seeming like it should work.

```java
final Component c0 = Component.text("best!");
final Component c1 = Component.text()
  .append(Component.text("Nero "))
  .append(Component.text(" are the ").append(c0))
  .build();

assertTrue(c1.contains(Component.text("best!")));
```

As far as I'm concerned there are a few possible solutions to this:
1. Modify the Javadoc to note that this is how the check works.
2. Swap the identity check for a true equality check.
3. Create a new method, `Component#contains(Component, boolean)` which can be used to define the type of check that is made and make the old method overload this one.

I kinda prefer options 1 or 3 because doing a true equality check every time here is going to be an awful lot of recursion which would suck.

kashike:
fixes #363

### DIFF
--- a/api/src/main/java/net/kyori/adventure/text/Component.java
+++ b/api/src/main/java/net/kyori/adventure/text/Component.java
@@ -28,7 +28,9 @@ import java.util.Collections;
 import java.util.Iterator;
 import java.util.List;
 import java.util.Map;
+import java.util.Objects;
 import java.util.Set;
+import java.util.function.BiPredicate;
 import java.util.function.Consumer;
 import java.util.function.Function;
 import java.util.function.UnaryOperator;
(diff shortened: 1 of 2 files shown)


api/src/main/java/net/kyori/adventure/text/Component.java
```diff
@@ -93,6 +95,19 @@
...
 @ApiStatus.NonExtendable
 public interface Component extends ComponentBuilderApplicable, ComponentLike, Examinable, HoverEventSource<Component> {
+  /**
+   * A predicate that checks equality of two {@code Component}s using {@link Objects#equals(Object, Object)}.
+   *
+   * @since 4.8.0
+   */
+  BiPredicate<? super Component, ? super Component> EQUALS = Objects::equals;
+  /**
+   * A predicate that checks equality of two {@code Component}s.
```
kezz, line 105: Might be worth adding in the javadoc that this uses `==`.

PR #365, opened 2021-05-13
api: `Audience#foreach` and `Audience#filter`

kashike:


api/src/main/java/net/kyori/adventure/audience/Audience.java
```diff
@@ -133,6 +135,34 @@
...
+  /**
+   * Executes an action against all audiences.
+   *
+   * <p>If you implement {@code Audience} and not {@link ForwardingAudience} in your own code, and your audience forwards to
+   * other audiences, then you <b>must</b> override this method and provide each audience to {@code action}.</p>
+   *
+   * @param action the action
+   * @since 4.8.0
+   */
+  default void foreach(final @NonNull Consumer<? super Audience> action) {
```
kezz, line 162: Is there a reason this isn't `forEach`?

kezz (commented):
I think this is a decent idea - there's a lot of good use cases for it at least. However, I wonder if it might just be worth implementing `Iterable` at this point?

api/src/main/java/net/kyori/adventure/audience/Audience.java
kashike: To avoid conflicts with `Iterable`.

kashike, line 162: Would it be better to call this `forEachAudience`?

api/src/main/java/net/kyori/adventure/audience/Audience.java
```diff
@@ -133,6 +135,34 @@
...
 
+  /**
+   * Filters this audience.
+   *
+   * <p>The returned {@code Audience} may be the same, or a completely different one.</p>
+   *
+   * @param filter the filter
+   * @since 4.8.0
+   */
+  default @NonNull Audience filter(final @NonNull Predicate<? super Audience> filter) {
```
kashike, line 146: Would it be better to call this `filterAudience`?

api/src/main/java/net/kyori/adventure/audience/Audience.java
kezz: Personally, I'd prefer the more specific names. Less change for a clash and it also means the ugly `foreach` doesn't have to exist!

Draycia: I prefer `forEeachAudience` it's much clearer, having `foreach` and `forEach` in a class would get confusing super fast.
Having it be `foreach` may also mess with people who have their IDE autocomplete for them, as they may get the method they don't want, and not notice the capitalisation difference.

kezz:
I mentioned this in Discord and it got lost to time so I'll just repeat here. It could be a good idea to just provide a method to get the audience as an iterable directly. Maybe exposing a stream as well could be a more "standard" way of implementing the filter system.

api/src/main/java/net/kyori/adventure/audience/Audience.java
Draycia: `filterAudience`, to be consistent with above

api/src/main/java/net/kyori/adventure/audience/ForwardingAudience.java
```diff
@@ -55,6 +60,30 @@
   @ApiStatus.OverrideOnly
   @NonNull Iterable<? extends Audience> audiences();
 
+  @Override
+  default @NonNull Audience filterAudience(final @NonNull Predicate<? super Audience> filter) {
+    @Nullable List<Audience> audiences = null;
+    for(final Audience audience : this.audiences()) {
+      if(filter.test(audience)) { // todo(kashike): should we test before filtering children?
```
kashike, line 67: thoughts?

BomBardyGamer: Well I think we only want to filter children of audiences that match our predicate, so yeah, I say this is a good idea

api/src/main/java/net/kyori/adventure/audience/ForwardingAudience.java
```diff
@@ -157,6 +186,19 @@ default void openBook(final @NonNull Book book) {
       return Collections.singleton(this.audience());
     }
 
+    @Override
+    default @NonNull Audience filterAudience(final @NonNull Predicate<? super Audience> filter) {
+      final Audience audience = this.audience();
+      return filter.test(audience) // todo(kashike): should we be testing "audience" or "this"
```
BomBardyGamer, line 192: Probably `this`

api/src/main/java/net/kyori/adventure/audience/ForwardingAudience.java
zml2008: Does a forwarding audience have an identity of its own, or is a transparent wrapper around its children?

imo in the default implementation, a forwarding audience does not have any identity, so not testing is appropriate. Specific implementations of forwarding audiences may choose to add their own identity, in which case they can choose to implement testing themselves against the predicate in an overridden method.

api/src/main/java/net/kyori/adventure/audience/ForwardingAudience.java
zml2008: as above, I'd say `audience`

lucko:
To be honest, I don't see the usecase...

`foreach`: why iterate when all of the methods on the audience delegate anyway?   
`filter`: can appreciate how this *might* be useful, but what properties do you have to filter by? Audience doesn't really give you much unless you try casting etc...

api/src/main/java/net/kyori/adventure/audience/EmptyAudience.java
```diff
@@ -51,6 +53,15 @@
     return defaultValue.get();
   }
 
+  @Override
+  public @NonNull Audience filterAudience(final @NonNull Predicate<? super Audience> filter) {
+    return this;
```
kezz, line 58: Is there a reason why this method is overidden here? Even though it's empty, I'd still assume that it would be ran through the predicate.

api/src/main/java/net/kyori/adventure/audience/Audience.java
```diff
@@ -134,6 +136,34 @@
...
+  /**
+   * Filters this audience.
+   *
+   * <p>The returned {@code Audience} may be the same, or a completely different one.</p>
+   *
+   * @param filter the filter
+   * @since 4.8.0
+   */
+  default @NonNull Audience filterAudience(final @NonNull Predicate<? super Audience> filter) {
+    if(filter.test(this)) {
```
kezz, line 148: This could use the same inline if as the override in `ForwardingAudience.Single`. Alternatively, that method should use this format. Either way, it should probably be consistent.

kezz (changes requested):
Needs updating in regards to `@since` tags and style changes. My earlier comments still apply, but I like the concept - especially so with the addition of pointers.

api/src/main/java/net/kyori/adventure/audience/EmptyAudience.java
zml2008: This audience has no identity of its own -- it's like running `forEach` on an empty list.

api/src/main/java/net/kyori/adventure/audience/ForwardingAudience.java
```diff
@@ -74,6 +79,30 @@
     return defaultValue.get(); // unsupported
   }
 
+  @Override
+  default @NonNull Audience filterAudience(final @NonNull Predicate<? super Audience> filter) {
+    @Nullable List<Audience> audiences = null;
+    for(final Audience audience : this.audiences()) {
+      if(filter.test(audience)) {
+        final Audience filtered = audience.filterAudience(filter);
+        if(filtered != Audience.empty()) {
```
zml2008, line 88: why call the predicate on an empty audience at all? can this check be moved to before `filter.test(audience)`?

api/src/main/java/net/kyori/adventure/audience/EmptyAudience.java
kezz: I'm not quite sure it is, I think it's more like the equivalent of skipping nulls on a during a `forEach` on a list with nullable elements.

PR #366, opened 2021-05-13
api: audience types

kashike:


api/src/main/java/net/kyori/adventure/audience/Audience.java
```diff
@@ -78,6 +80,34 @@
...
+   */
+  Key TYPE_EMPTY = Key.key(Adventure.NAMESPACE, "empty");
+  /**
+   * A forwarded audience.
+   *
+   * @see ForwardingAudience
+   * @see ForwardingAudience.Single
+   * @since 4.8.0
+   */
+  Key TYPE_FORWARDED = Key.key(Adventure.NAMESPACE, "console");
```
Draycia, line 103:
```suggestion
  Key TYPE_FORWARDED = Key.key(Adventure.NAMESPACE, "forwarded");
```

api/src/main/java/net/kyori/adventure/audience/Audience.java
```diff
@@ -78,6 +80,41 @@
...
+   *
+   * @since 4.8.0
+   */
+  Key TYPE_CONSOLE = Key.key(Adventure.NAMESPACE, "console");
+  /**
+   * A player audience.
+   *
+   * @since 4.8.0
+   */
+  Key TYPE_PLAYER = Key.key("player");
```
kashike, line 116: note: this is in the `minecraft` namespace, while the others are in the `adventure` namespace - this is because we are using the entity key here, since we can

kezz:
Same comment as on the keys PR, I think the types should live in their own class to avoid polluting other classes with these constants.

Additionally, I'm not quite sure why the class-specific types exist. There are probably cases, say with a `WORLD` type, where an audience is both forwarding **and** a specific type. I think it makes more sense to provide a default `UNKNOWN` type and let people do instanceof checks if they want to know what audience class it uses.

One other slight issue which could exist with this is the fact that it becomes tricky to define custom types without losing functionality if your type is a subtype of a core type (say, for example, `fake_player` vs `player` in Minestom). Perhaps returning a set of types? This might also solve the issue of the types that represent the audience classes.

Some other suggestions for types could be logger/log, world, team, server or viewers.

api/src/main/java/net/kyori/adventure/audience/Audience.java
```diff
@@ -133,6 +170,14 @@
     return Audiences.COLLECTOR;
   }
 
+  /**
+   * Gets the audience type.
+   *
+   * @return the audience type
+   * @since 4.8.0
+   */
+  @NonNull Key audienceType();
```
A248, line 179: Isn't this a breaking change for Audience implementors? Audience does not have `ApiStatus.NonExtendable` on it.

kezz: I think defaulting to an `UNKNOWN` type is a good solution for this.

BomBardyGamer (approved):
Looks good to me

zml2008:
I know there was some reason you discussed this, but I can't recall it at the moment.

I'm currently uncertain on this concept -- it adds a lot of 'concreteness'

The question I asked on #365 applies here too: do forwarding audiences have their own identity? This PR would indicate that the answer is "yes". I'm not sure that's the way to go either.

Issue #372, opened 2021-05-18
Migrate remaining text-serializer-gson serializers to use `TypeAdapter`s


kashike:
The following serialisers need some work to be converted from implementing `JsonDeserializer`/`JsonSerializer` to `TypeAdapter`:
- [ ] [ComponentSerializerImpl](https://github.com/KyoriPowered/adventure/blob/master/text-serializer-gson/src/main/java/net/kyori/adventure/text/serializer/gson/ComponentSerializerImpl.java)
- [ ] [ShowEntitySerializer](https://github.com/KyoriPowered/adventure/blob/master/text-serializer-gson/src/main/java/net/kyori/adventure/text/serializer/gson/ShowEntitySerializer.java)
- [ ] [ShowItemSerializer](https://github.com/KyoriPowered/adventure/blob/master/text-serializer-gson/src/main/java/net/kyori/adventure/text/serializer/gson/ShowItemSerializer.java)
- [ ] [StyleSerializer](https://github.com/KyoriPowered/adventure/blob/master/text-serializer-gson/src/main/java/net/kyori/adventure/text/serializer/gson/StyleSerializer.java)

emilyy-dev:
Self assigned over at https://github.com/emilyy-dev/adventure/tree/feature/gson-type-adapters

### Checklist from above
* [ ] <s>[`ComponentSerializerImpl`](https://github.com/emilyy-dev/adventure/blob/feature/gson-type-adapters/text-serializer-gson/src/main/java/net/kyori/adventure/text/serializer/gson/ComponentSerializerImpl.java)</s> (see note)
* [ ] [`ShowEntitySerializer`](https://github.com/emilyy-dev/adventure/blob/feature/gson-type-adapters/text-serializer-gson/src/main/java/net/kyori/adventure/text/serializer/gson/ShowEntitySerializer.java)
* [x] [`ShowItemSerializer`](https://github.com/emilyy-dev/adventure/blob/feature/gson-type-adapters/text-serializer-gson/src/main/java/net/kyori/adventure/text/serializer/gson/ShowItemSerializer.java)
* [ ] [`StyleSerializer`](https://github.com/emilyy-dev/adventure/blob/feature/gson-type-adapters/text-serializer-gson/src/main/java/net/kyori/adventure/text/serializer/gson/StyleSerializer.java)

### Additional changes
[...]

PR #373, opened 2021-05-19
Migrate remaining text-serializer-gson serializers to use `TypeAdapter`s

emilyy-dev:
#372


text-serializer-gson/src/main/java/net/kyori/adventure/text/serializer/gson/ShowItemSerializer.java
```diff
@@ -23,71 +23,82 @@
...
 
-    json.add(ID, context.serialize(src.item()));
+    out.name(ID);
+    KeySerializer.INSTANCE.write(out, value.item());
 
-    final int count = src.count();
+    final int count = value.count();
     if(count != 1) {
-      json.addProperty(COUNT, count);
+      out.name(COUNT).value(count);
```
kashike, line 94: I think I'd prefer to split this (applies to all cases) to be consistent with cases like above where we delegate to another serialiser:
```suggestion
      out.name(COUNT);
      out.value(count);
```

kashike (changes requested):

text-serializer-gson/src/main/java/net/kyori/adventure/text/serializer/gson/ShowItemSerializer.java
emilyy-dev: https://github.com/KyoriPowered/adventure/pull/373/commits/4bc69716006f639c63a19847be2b895bd5065ab8

text-serializer-gson/src/main/java/net/kyori/adventure/text/serializer/gson/SerializerFactory.java
```diff
@@ -0,0 +1,91 @@
...
+  @Override
+  @SuppressWarnings("unchecked")
+  public <T> TypeAdapter<T> create(final Gson gson, final TypeToken<T> type) {
+    final Class<? super T> rawType = type.getRawType();
+    if(KEY_TYPE.isAssignableFrom(rawType)) {
+      return (TypeAdapter<T>) KeySerializer.INSTANCE;
+    } else if(STYLE_TYPE.isAssignableFrom(rawType)) {
+      return (TypeAdapter<T>) new StyleSerializer(this.legacyHoverSerializer, this.emitLegacyHover, gson::getAdapter);
+    } else if(CLICK_ACTION_TYPE.isAssignableFrom(rawType)) {
+      return (TypeAdapter<T>) IndexedSerializer.of("click action", ClickEvent.Action.NAMES);
```
kashike, line 72: Should be returning an `INSTANCE` here too.

text-serializer-gson/src/main/java/net/kyori/adventure/text/serializer/gson/SerializerFactory.java
```diff
@@ -0,0 +1,91 @@
...
+    } else if(HOVER_ACTION_TYPE.isAssignableFrom(rawType)) {
+      return (TypeAdapter<T>) IndexedSerializer.of("hover action", HoverEvent.Action.NAMES);
```
kashike, line 74: Should be returning an `INSTANCE` here too.

text-serializer-gson/src/main/java/net/kyori/adventure/text/serializer/gson/SerializerFactory.java
```diff
@@ -0,0 +1,91 @@
...
+    } else if(SHOW_ITEM_TYPE.isAssignableFrom(rawType)) {
+      return (TypeAdapter<T>) ShowItemSerializer.INSTANCE;
+    } else if(SHOW_ENTITY_TYPE.isAssignableFrom(rawType)) {
+      return (TypeAdapter<T>) new ShowEntitySerializer(gson::getAdapter);
+    } else if(COLOR_WRAPPER_TYPE.isAssignableFrom(rawType)) {
+      return (TypeAdapter<T>) TextColorWrapper.Serializer.INSTANCE;
+    } else if(COLOR_TYPE.isAssignableFrom(rawType)) {
+      return (TypeAdapter<T>) (this.downsampleColors ? TextColorSerializer.DOWNSAMPLE_COLOR : TextColorSerializer.INSTANCE);
+    } else if(TEXT_DECORATION_TYPE.isAssignableFrom(rawType)) {
+      return (TypeAdapter<T>) IndexedSerializer.of("text decoration", TextDecoration.NAMES);
```
kashike, line 84: Should be returning an `INSTANCE` here too.

kashike (changes requested):

text-serializer-gson/src/main/java/net/kyori/adventure/text/serializer/gson/SerializerFactory.java
emilyy-dev: I don't quite like the idea of the instances being in the same `IndexedSerializer` but I believe it's where they fit best, thoughts?
https://github.com/KyoriPowered/adventure/pull/373/commits/8e51ffd22b3c9597b51940a87e42d98c8cc1987f

kashike: I'd just make a class with a single field for each of these, maybe? `TextDecorationSerializer`, etc

text-serializer-gson/src/main/java/net/kyori/adventure/text/serializer/gson/SerializerFactory.java
```diff
@@ -0,0 +1,104 @@
...
+  SerializerFactory(final boolean downsampleColors, final @Nullable LegacyHoverEventSerializer legacyHoverSerializer, final boolean emitLegacyHover) {
+    this.downsampleColors = downsampleColors;
+    this.legacyHoverSerializer = legacyHoverSerializer;
+    this.emitLegacyHover = emitLegacyHover;
+    this.colorSerializer = downsampleColors ? TextColorSerializer.DOWNSAMPLE_COLOR : TextColorSerializer.INSTANCE;
+
+    // TODO: this is hideous
+    this.styleSerializer = new StyleSerializer(legacyHoverSerializer, emitLegacyHover, this.colorSerializer);
+    this.componentSerializer = this.styleSerializer.getComponentSerializer();
+    this.showEntitySerializer = this.styleSerializer.getShowEntitySerializer();
```
kashike, line 71: Is there a way to make it less-hideous?

zml2008 (changes requested):
Thanks for doing this (and sorry for taking so long to get to reviewing)!

I'd rather you request TypeAdapter instances via a `Gson` instance in each adapter's constructor, rather than using static fields, at least where possible. That allows doing fun tricks with swapping out one adapter for another.

Second -- I've recently tossed in some wiring for JMH benchmarking. Would you be willing to set up some basic test cases for a before/after comparison? You should be able to use `ComponentCompactionBenchmark` in the `api` submodule as a reference for that.

text-serializer-gson/src/main/java/net/kyori/adventure/text/serializer/gson/ComponentSerializerImpl.java
```diff
@@ -70,18 +69,28 @@
...
+
+  private ComponentSerializerImpl(final TypeAdapter<Style> styleSerializer) {
+    this.styleSerializer = styleSerializer;
+  }
+
   @Override
-  public Component deserialize(final JsonElement json, final Type typeOfT, final JsonDeserializationContext context) throws JsonParseException {
-    return this.deserialize0(json, context);
+  public Component read(final JsonReader in) throws IOException {
+    return this.deserialize(Streams.parse(in));
```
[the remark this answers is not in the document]
emilyy-dev: > I'd rather you request TypeAdapter instances via a `Gson` instance in each adapter's constructor, rather than using static fields, at least where possible. That allows doing fun tricks with swapping out one adapter for another.

Actually doing this would solve a couple of issues, first this one you pointed out, and second the maze-like mess of the style serializer and component serializer (impl) requiring each other leading to this:
https://github.com/KyoriPowered/adventure/blob/06ec1dfd3cff6daeb72e01b345b43268e47ee074/text-serializer-gson/src/main/java/net/kyori/adventure/text/serializer/gson/StyleSerializer.java#L92-L93

Additionally it would be an ideal/equivalent replacement for `JsonSerializationContext` / `JsonDeserializationContext` give its over-usage in some of the serializers.

That was an eye-opening suggestion :smile:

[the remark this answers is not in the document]
zml2008: Unfortunately I think you might not be able to avoid the style serializer mess because of the circular dependency there -- but you are welcome to give it a shot :p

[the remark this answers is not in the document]
emilyy-dev: I believe it should work fine? The circular dependency is not needed at the spot on construction/creation, but rather lazily.
Will definitely try and test! Ending with a SOE from some infinite recursion is not on my schedule :laughing:

zml2008 (approved):

text-serializer-gson/src/jmh/java/net/kyori/adventure/text/serializer/gson/ComponentSerializationBenchmark.java
```diff
@@ -0,0 +1,83 @@
...
+import static net.kyori.adventure.text.event.HoverEvent.showText;
+import static net.kyori.adventure.text.format.Style.style;
+import static net.kyori.adventure.text.format.TextColor.color;
+import static net.kyori.adventure.text.serializer.gson.GsonComponentSerializer.gson;
+
+@State(Scope.Benchmark)
+@BenchmarkMode(Mode.SampleTime)
+@OutputTimeUnit(TimeUnit.MICROSECONDS)
+public class ComponentSerializationBenchmark {
+
```
kashike, line 52: no empty lines before class content please

kashike (changes requested):

zml2008 (approved):
The ComponentSerializerImpl and StyleSerializer classes are definitely a touch messier than we'd like, but that's unavoidable I think. Thank you for pushing this through to the finish line!

PR #379, opened 2021-05-23
feat(api/text) #378: ability to unset decorations during component construction

kashike:


api/src/main/java/net/kyori/adventure/text/format/TextDecoration.java
```diff
@@ -77,6 +77,17 @@
...
   }
 
+  /**
+   * Creates a {@link TextDecorationAndState}.
+   *
+   * @param state the state
+   * @return a {@link TextDecorationAndState}
+   * @since 4.8.0
+   */
+  public final @NonNull TextDecorationAndState as(final @NonNull State state) {
```
zml2008, line 87: Your example of `TextDecoration.BOLD.as(TextDecoration.State.TRUE)` from the tests is annoyingly verbose imo.

Does it make sense to add some helper methods like `asTrue()` and `asFalse()` to reduce that verbosity?

zml2008 (changes requested):

api/src/main/java/net/kyori/adventure/text/format/TextDecoration.java
lucko: `TextDecoration.BOLD.setTo(true)` reads a bit better IMO.

kashike: `as(boolean)` added.

zml2008 (approved):

PR #388, opened 2021-05-29
api: Add Pointers

kezz:
This PR adds a class that holds a collection of pointers and optional values. With this class, the following goals are achieved:

### Easier implementation
With a class such as this, implementing the pointered interface will only involve building a pointers instance and returning it. 

#### Before
```java
@Override
@SuppressWarnings("unchecked") // safe casts
public @NonNull <T> Optional<T> get(@NonNull Pointer<T> pointer) {
    if(pointer.equals(NAME)) return Optional.ofNullable((T) this.name());
    if(pointer.equals(SERVER_NAME)) return Optional.of(this.getServer().name());
    return Optional.empty();
}
```

#### After
```java
private Pointers pointers = null;

@Override
public @NonNull Pointers pointers() {
    if(this.pointers == null) {
        this.pointers = Pointers.builder()
            .addPointerWithVariableValue(NAME, this::name)
            .addPointerWithFixedValue(SERVER_NAME, this.getServer().name())
            .build();
    }

    return this.pointers;
}
```

### Distinction between null and not set values
With the existing system, there is no way to tell if a value is not set or if it isn't present. 
For example, if an implementation had a nullable display name, there is no way to tell if an empty optional means that the audience has no display name or if the audience cannot have a display name.
Alternatively, new pointers might be added and standardised but not implemented in some implementations, meaning that the difference between a not set value and a null value becomes more important.
With this, the pointers class introduces a method to check if a value is set using the tri-state class introduced earlier.

### DIFF
--- a/api/src/main/java/net/kyori/adventure/pointer/Pointers.java
+++ b/api/src/main/java/net/kyori/adventure/pointer/Pointers.java
@@ -26,7 +26,6 @@ package net.kyori.adventure.pointer;
 import java.util.Optional;
 import java.util.function.Supplier;
 import net.kyori.adventure.util.Buildable;
-import net.kyori.adventure.util.TriState;
 import org.jetbrains.annotations.Contract;
 import org.jetbrains.annotations.NotNull;
 import org.jetbrains.annotations.Nullable;
@@ -39,7 +38,7 @@ import org.jetbrains.annotations.UnknownNullability;
  */
 public interface Pointers extends Buildable<Pointers, Pointers.Builder> {
   /**
-   * An empty set of pointers.
+   * An empty collection of pointers.
    *
    * @since 4.8.0
    */
@@ -70,7 +69,7 @@ public interface Pointers extends Buildable<Pointers, Pointers.Builder> {
   /**
    * Gets the value of {@code pointer}.
    *
-   * <p>If this {@code Audience} is unable to provide a value for {@code pointer}, {@code defaultValue} will be returned.</p>
+   * <p>If a value for {@code pointer} is unable to be provided, {@code defaultValue} will be returned.</p>
    *
    * @param pointer the pointer
    * @param defaultValue the default value
@@ -87,7 +86,7 @@ public interface Pointers extends Buildable<Pointers, Pointers.Builder> {
   /**
    * Gets the value of {@code pointer}.
    *
-   * <p>If this {@code Audience} is unable to provide a value for {@code pointer}, the value supplied by {@code defaultValue} will be returned.</p>
+   * <p>If a value for {@code pointer} is unable to be provided, the value supplied by {@code defaultValue} will be returned.</p>
    *
    * @param pointer the pointer
    * @param defaultValue the default value supplier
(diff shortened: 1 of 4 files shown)


api/src/main/java/net/kyori/adventure/pointer/Pointers.java
```diff
@@ -0,0 +1,179 @@
...
+  @NonNull Pointers EMPTY = PointersImpl.EMPTY;
+
+  /**
+   * Gets a new pointers builder.
+   *
+   * @return the builder
+   * @see Builder
+   * @since 4.8.0
+   */
+  @Contract(value = "-> new", pure = true)
```
kashike, line 55: Invalid annotation.

kashike (changes requested):

kashike:
Not sure how I feel about inheritance here.

lucko:
I agree, this looks good but I don't think inheritance makes sense.

kezz:
Yeah the more I think about inheritance, the harder it gets to make it all work out. I really do like the idea as it enables stuff like team or server names to be obtained really easily but without multiple parents (in the case of team/world/etc) it becomes annoying to handle.

So I've rebased and removed the inheritance, it's a lot simpler of an implementation.

api/src/main/java/net/kyori/adventure/pointer/Pointers.java
```diff
@@ -0,0 +1,159 @@
...
+
+  /**
+   * Checks if these pointers contain a value for the given pointer.
+   *
+   * @param pointer the pointer
+   * @param <T> the type
+   * @return a tri-state
+   * @since 4.8.0
+   */
+  <T> @NotNull TriState has(final @NotNull Pointer<T> pointer);
```
zml2008, line 111: Not sure if it makes sense for this to be a `TriState`.

From looking at the implementation, it seems that this returns NOT_SET if the pointer is not known to this `Pointers` instance, and TRUE/FALSE based on the nullness of the value, but I'm not sure that distinction is useful.

zml2008 (changes requested):
I like this concept, see comment for my one issue

api/src/main/java/net/kyori/adventure/pointer/Pointers.java
kashike: Perhaps a different name for this method? `supports`?

kezz: A different enum could work, `Pointer.State` with `PRESENT`, `ABSENT` and `NOT_SUPPORTED` perhaps?

kashike: Not a fan of a new enum for this.

zml2008: I think a boolean `supports()` like kashike suggested would make the most sense here, just indicating whether a Pointer is supported or not, and saying nothing about its value.

kezz: Ah I like that! So it essentially just returns the result of `containsKey()`?

kashike: Yea

zml2008 (approved):

api/src/main/java/net/kyori/adventure/pointer/Pointers.java
```diff
@@ -0,0 +1,160 @@
...
+ *
+ * @since 4.8.0
+ */
+public interface Pointers extends Buildable<Pointers, Pointers.Builder> {
+  /**
+   * An empty collection of pointers.
+   *
+   * @since 4.8.0
+   */
+  @NotNull Pointers EMPTY = PointersImpl.EMPTY;
```
kashike, line 45: Would prefer this be a method instead.

kashike (changes requested):

kashike (approved):

PR #390, opened 2021-05-29
Remove checker qual and migrate to jetbrains annotations

### LINKED ISSUE #386: Migrate from checker-qual to jetbrains annotations
`org.jetbrains:annotations` is a multi-release dependency with [`module-info`](https://github.com/JetBrains/java-annotations/blob/master/module-info/src/main/java/module-info.java) support, unlike `org.checkerframework:checker-qual`.

kezz:
Closes #386.

This also removes the compile only dependencies on checker qual, not sure if these should be left until a breaking release or not.

Some points of note:
- Some `@Nullable` annotations were added where they were missing in equals overrides.
- There is no replacement for `@MonotonicNonNull` or `@PolyNull`. Both have been replaced with `@Nullable` annotations with appropriate `@Contract` annotations for poly null methods.

### DIFF
--- a/api/src/main/java/net/kyori/adventure/audience/EmptyAudience.java
+++ b/api/src/main/java/net/kyori/adventure/audience/EmptyAudience.java
@@ -30,69 +30,72 @@ import net.kyori.adventure.identity.Identity;
 import net.kyori.adventure.inventory.Book;
 import net.kyori.adventure.pointer.Pointer;
 import net.kyori.adventure.text.ComponentLike;
-import org.checkerframework.checker.nullness.qual.NonNull;
-import org.checkerframework.checker.nullness.qual.PolyNull;
+import org.jetbrains.annotations.Contract;
+import org.jetbrains.annotations.NotNull;
+import org.jetbrains.annotations.Nullable;
+import org.jetbrains.annotations.UnknownNullability;
 
 final class EmptyAudience implements Audience {
   static final EmptyAudience INSTANCE = new EmptyAudience();
 
   @Override
-  public @NonNull <T> Optional<T> get(final @NonNull Pointer<T> pointer) {
+  public @NotNull <T> Optional<T> get(final @NotNull Pointer<T> pointer) {
     return Optional.empty();
   }
 
+  @Contract("_, null -> null; _, !null -> !null")
   @Override
-  public <T> @PolyNull T getOrDefault(final @NonNull Pointer<T> pointer, final @PolyNull T defaultValue) {
+  public <T> @Nullable T getOrDefault(final @NotNull Pointer<T> pointer, final @Nullable T defaultValue) {
     return defaultValue;
   }
 
   @Override
-  public <T> @PolyNull T getOrDefaultFrom(final @NonNull Pointer<T> pointer, final @NonNull Supplier<? extends T> defaultValue) {
+  public <T> @UnknownNullability T getOrDefaultFrom(final @NotNull Pointer<T> pointer, final @NotNull Supplier<? extends T> defaultValue) {
     return defaultValue.get();
   }
 
   @Override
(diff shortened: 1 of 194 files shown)


kashike:
`org.jetbrains.annotations.UnknownNullability`?

kezz:
> `org.jetbrains.annotations.UnknownNullability`?

I pondered that but it's not really true, we do know the nullability but there's no way to accurately express it. That being said, it is definitely the case that `@UnknownNullability` will prevent IDE warnings (IDEA doesn't bothing nagging you) but at the cost of people potentially missing errors in their own code. I think it might make more sense to leave it `@Nullable` and let people ignore this warning if they know that the supplier never returns null.

kashike:
> I think it might make more sense to leave it `@Nullable` and let people ignore this warning if they know that the supplier never returns null.

I would rather we drop the annotation - I know that having this `@Nullable` would annoy people, myself included.

kezz:
> I would rather we drop the annotation - I know that having this `@Nullable` would annoy people, myself included.

Okie dokie! I've moved it over to `@UnknownNullability` as that is the default for non-annotated types and the docs for that method state pretty much the exact reason we'd use it in these supplier cases.

api/src/main/java/net/kyori/adventure/audience/EmptyAudience.java
```diff
@@ -30,69 +30,72 @@
...
   @Override
-  public @NonNull <T> Optional<T> get(final @NonNull Pointer<T> pointer) {
+  public @NotNull <T> Optional<T> get(final @NotNull Pointer<T> pointer) {
     return Optional.empty();
   }
 
+  @Contract("_, null -> null; _, !null -> !null")
   @Override
-  public <T> @PolyNull T getOrDefault(final @NonNull Pointer<T> pointer, final @PolyNull T defaultValue) {
+  public <T> @UnknownNullability T getOrDefault(final @NotNull Pointer<T> pointer, final @Nullable T defaultValue) {
```
kezz, line 48: This annotation should be on the method below

api/src/main/java/net/kyori/adventure/audience/ForwardingAudience.java
```diff
@@ -57,55 +59,56 @@
...
   @Override
-  default <T> @NonNull Optional<T> get(final @NonNull Pointer<T> pointer) {
+  default <T> @NotNull Optional<T> get(final @NotNull Pointer<T> pointer) {
     return Optional.empty(); // unsupported
   }
 
+  @Contract("_, null -> null; _, !null -> !null")
   @Override
-  default <T> @PolyNull T getOrDefault(final @NonNull Pointer<T> pointer, final @PolyNull T defaultValue) {
+  default <T> @UnknownNullability T getOrDefault(final @NotNull Pointer<T> pointer, final @Nullable T defaultValue) {
```
kezz, line 71: Also needs to be on the line below.

api/src/main/java/net/kyori/adventure/audience/EmptyAudience.java
kashike: What?

api/src/main/java/net/kyori/adventure/audience/ForwardingAudience.java
kashike: What?

api/src/main/java/net/kyori/adventure/audience/EmptyAudience.java
kezz: It's the supplier methods that need annotating with the `@UnknownNullability` annotations. These methods are fine as the poly-nullability is covered by the `@Contract` annotations.

kashike:
Rebase required.

lucko (approved):

kashike (approved):

PR #401, opened 2021-06-08
api: Fix TextColor#lerp docs

kezz:


kezz:
hmm actually i think this does make sense and i just can't read

PR #408, opened 2021-06-22
Clean up StyleImpl

### LINKED ISSUE #392: Reduce duplication in StyleImpl
[`StyleImpl`](https://github.com/KyoriPowered/adventure/blob/master/api/src/main/java/net/kyori/adventure/text/format/StyleImpl.java) has a lot of duplication when it comes to `TextDecoroation` - it would be nice to find a way to reduce this.

MrMacor:
This closes #392 by replacing the hard-coded `TextDecoration` values in `StyleImpl` and `StyleImpl.BuilderImpl` with `TextDecoration.State[]`s. The index of each `State` is the same as the ordinal of the `TextDecoration` it represents.

### DIFF
--- a/api/src/main/java/net/kyori/adventure/text/format/Style.java
+++ b/api/src/main/java/net/kyori/adventure/text/format/Style.java
@@ -27,6 +27,7 @@ import java.util.Collections;
 import java.util.Map;
 import java.util.Set;
 import java.util.function.Consumer;
+import java.util.stream.Stream;
 import net.kyori.adventure.key.Key;
 import net.kyori.adventure.sound.Sound;
 import net.kyori.adventure.text.Component;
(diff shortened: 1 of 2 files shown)


api/src/main/java/net/kyori/adventure/text/format/Style.java
```diff
@@ -108,7 +109,7 @@
    */
   static @NotNull Style style(final @Nullable TextColor color) {
     if (color == null) return empty();
-    return new StyleImpl(null, color, TextDecoration.State.NOT_SET, TextDecoration.State.NOT_SET, TextDecoration.State.NOT_SET, TextDecoration.State.NOT_SET, TextDecoration.State.NOT_SET, null, null, null);
+    return new StyleImpl(null, color, Stream.of(TextDecoration.values()).map(decoration -> TextDecoration.State.NOT_SET).toArray(TextDecoration.State[]::new), null, null, null);
```
zml2008, line 112: this is a joke, right?

zml2008 (changes requested):

kashike:
I am not really a fan of this approach - it introduces mutable state (via modifying the arrays) an immutable object.

zml2008:
I don't think this will be something we'll be wanting to merge.

PR #410, opened 2021-06-24
api: ClickEvent setters and ClickEventSource

kezz:
This PR adds a `ClickEventSource` class in addition to setters for `ClickEvent` properties. In order to prevent the annoyance of forcing people to change `component.clickEvent(null)` to `component.clickEvent((ClickEvent) null)` to prevent unwanted errors the method for setting a click event from a source is named `clickEventFromSource`. This could be replaced to use just `ClickEventSource` methods in 5.0.0 if we fancy a small, but technically breaking, change.

The `ClickEventSource` class provides a method to set the action of the resulting click event. This will resolve the common use case of, for example, obtaining a command as a click event but with a different method (run vs. suggest).

### DIFF
--- a/api/src/main/java/net/kyori/adventure/text/Component.java
+++ b/api/src/main/java/net/kyori/adventure/text/Component.java
@@ -38,6 +38,7 @@ import java.util.regex.Pattern;
 import java.util.stream.Collector;
 import net.kyori.adventure.key.Key;
 import net.kyori.adventure.text.event.ClickEvent;
+import net.kyori.adventure.text.event.ClickEventSource;
 import net.kyori.adventure.text.event.HoverEvent;
 import net.kyori.adventure.text.event.HoverEventSource;
 import net.kyori.adventure.text.format.Style;
@@ -1838,6 +1839,18 @@ public interface Component extends ComponentBuilderApplicable, ComponentLike, Ex
   @Contract(pure = true)
   default @NotNull Component clickEvent(final @Nullable ClickEvent event) {
     return this.style(this.style().clickEvent(event));
+  }
+
+  /**
+   * Sets the click event of this component from a source.
+   *
+   * @param source the click event source
+   * @return a component
+   * @since 4.9.0
+   */
+  @Contract(pure = true)
+  default @NotNull Component clickEventFromSource(final @NotNull ClickEventSource source) {
+    return this.clickEvent(Objects.requireNonNull(source, "source").asClickEvent());
   }
 
   /**
--- a/api/src/main/java/net/kyori/adventure/text/ComponentBuilder.java
+++ b/api/src/main/java/net/kyori/adventure/text/ComponentBuilder.java
@@ -24,11 +24,13 @@
 package net.kyori.adventure.text;
 
 import java.util.List;
+import java.util.Objects;
 import java.util.Set;
 import java.util.function.Consumer;
 import java.util.function.Function;
 import net.kyori.adventure.key.Key;
 import net.kyori.adventure.text.event.ClickEvent;
+import net.kyori.adventure.text.event.ClickEventSource;
 import net.kyori.adventure.text.event.HoverEventSource;
 import net.kyori.adventure.text.format.Style;
 import net.kyori.adventure.text.format.TextColor;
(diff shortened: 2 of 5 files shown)


api/src/main/java/net/kyori/adventure/text/Component.java
```diff
@@ -1840,6 +1841,18 @@ default boolean hasDecoration(final @NotNull TextDecoration decoration) {
...
 
+  /**
+   * Sets the click event of this component from a source.
+   *
+   * @param source the click event source
+   * @return a component
+   * @since 4.9.0
+   */
+  @Contract(pure = true)
+  default @NotNull Component clickEventFromSource(final @NotNull ClickEventSource source) {
```
kashike, line 1852:
```suggestion
  default @NotNull Component clickEvent(final @NotNull ClickEventSource source) {
```

api/src/main/java/net/kyori/adventure/text/ComponentBuilder.java
```diff
@@ -297,6 +299,18 @@
...
 
+  /**
+   * Sets the click event of this component.
+   *
+   * @param source the click event source
+   * @return this builder
+   * @since 4.9.0
+   */
+  @Contract("_ -> this")
+  default @NotNull B clickEventFromSource(final @NotNull ClickEventSource source) {
```
kashike, line 310:
```suggestion
  default @NotNull B clickEvent(final @NotNull ClickEventSource source) {
```

api/src/main/java/net/kyori/adventure/text/event/ClickEventSource.java
```diff
@@ -0,0 +1,79 @@
...
+  }
+
+  /**
+   * Creates a click event derived from this object with a specific action.
+   *
+   * @param action the click event action
+   * @return a click event
+   * @since 4.9.0
+   */
+  default @NotNull ClickEvent asClickEvent(final ClickEvent.@NotNull Action action) {
```
kashike, line 64: I don't like this method.

api/src/main/java/net/kyori/adventure/text/format/Style.java
```diff
@@ -362,6 +364,17 @@ default boolean hasDecoration(final @NotNull TextDecoration decoration) {
...
   @NotNull Style clickEvent(final @Nullable ClickEvent event);
 
+  /**
+   * Sets the click event.
+   *
+   * @param source the click event source
+   * @return a style
+   * @since 4.9.0
+   */
+  default @NotNull Style clickEventFromSource(final @NotNull ClickEventSource source) {
```
kashike, line 374:
```suggestion
  default @NotNull Style clickEvent(final @NotNull ClickEventSource source) {
```

api/src/main/java/net/kyori/adventure/text/format/Style.java
```diff
@@ -713,6 +726,8 @@ static boolean hasAll(final @NotNull Set<Merge> merges) {
     @Contract("_ -> this")
     @NotNull Builder clickEvent(final @Nullable ClickEvent event);
 
+
+
```
kashike, line 730: ?

kashike (changes requested):

api/src/main/java/net/kyori/adventure/text/event/ClickEventSource.java
zml2008: yeah, me neither. I think providing a suitable `ClickEventSource` will be more difficult than for hover events, since click events are pretty string-based (and types they could use like URL or int are not controlled directly by platforms)

kashike: @kezz

kezz: Happy to remove, although I think removing it will just lead to everyone using the other method to modify it to set the action as zml is right, it's hard to know what the right action is for something like a command.

kezz:
Due to the lack of type safety around click events, in addition to the annoying need for a different methods name, I'm going to close this for now.

PR #422, opened 2021-08-15
Added Moshi component serializer

BomBardyGamer:
As the title states, this PR adds a new component serializer using the [Moshi](https://github.com/squareup/moshi) JSON library. 
This serializer borrows a bit from Gson's code, as Moshi and Gson are similar in parts with the way they function.

I have also copied all of the tests over from the Gson serializer, as I think it is fair that this serializer should be able to at least do everything that the Gson one does, and they are all passing (I ran them all multiple times).

You may also notice that, as per @zml2008 's request, I have included some JMH benchmarks that I copied from [this PR](https://github.com/KyoriPowered/adventure/pull/373). I am running the tests as I am writing this, so I am unable to produce any results yet. Feel free to run these benchmarks yourself and post your results here, as it will help in determining performance differences.

I am also looking to implement more serializers for other JSON libraries in the near future, as I believe it would be useful for people to be able to natively serialize using a library of their choice, rather than being forced to use Gson. These will likely be kept separate from the main project, however, as I think that too many choices can confuse people a bit (unless you think that it is worth getting them merged).

PR #424, opened 2021-08-22
Added common abstraction between JSON serializers

BomBardyGamer:
The idea for this stems from #422 , though this has been split off due to PR scope issues.

This adds two new modules that will act as abstraction layer modules to allow implementations to define their own JSON serializers to use whilst end users can simply use JSON (de)serialization without needing to depend on a specific library's implementation.

### DIFF
--- a/text-serializer-gson/src/main/java/net/kyori/adventure/text/serializer/gson/GsonComponentSerializer.java
+++ b/text-serializer-gson/src/main/java/net/kyori/adventure/text/serializer/gson/GsonComponentSerializer.java
@@ -125,6 +125,12 @@ public interface GsonComponentSerializer extends JsonComponentSerializer {
    * @since 4.0.0
    */
   interface Builder extends JsonComponentSerializer.Builder {
+    @Override
+    @NotNull Builder downsampleColors();
+
+    @Override
+    @NotNull Builder legacyHoverEventSerializer(final @Nullable net.kyori.adventure.text.serializer.json.LegacyHoverEventSerializer serializer);
+
     /**
      * Sets a serializer that will be used to interpret legacy hover event {@code value} payloads.
      * If the serializer is {@code null}, then only {@link net.kyori.adventure.text.event.HoverEvent.Action#SHOW_TEXT}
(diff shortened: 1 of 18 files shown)


kezz:
Can you verify that none of the changes made are breaking, and if they are, fix this PR so that it doesn't include breaking changes? See [this](https://paste.gg/p/anonymous/8b825289059f4c258a4c2d3381b265ef) [japi](https://lvc.github.io/japi-compliance-checker/) report for info.

text-serializer-gson/src/main/java/net/kyori/adventure/text/serializer/gson/GsonComponentSerializer.java
```diff
@@ -111,20 +110,21 @@ static Builder builder() {
...
    * @since 4.0.0
    */
-  interface Builder extends Buildable.Builder<GsonComponentSerializer> {
-    /**
-     * Sets that the serializer should downsample hex colors to named colors.
-     *
-     * @return this builder
-     * @since 4.0.0
-     */
-    @NotNull Builder downsampleColors();
```
kezz, line 126: I don't think we can remove these methods from this builder as this is a breaking change I _thinkkkk_

text-serializer-gson/src/main/java/net/kyori/adventure/text/serializer/gson/GsonComponentSerializer.java
```diff
@@ -111,20 +110,21 @@ static Builder builder() {
...
-
+  interface Builder extends JsonComponentSerializer.Builder {
```
kezz, line 127: I don't think we can change the type of the builder here. Methods need to return `GsonComponentSerializer` so we don't break anything. Perhaps it needs to extend _both_ interfaces for now, ensuring all the methods return the old method?

kezz (changes requested):

text-serializer-gson/src/main/java/net/kyori/adventure/text/serializer/gson/GsonComponentSerializer.java
BomBardyGamer: Yeah I actually forgot that the return types are all screwed up. I can likely fix it by using a `<B extends Builder<B>>`, since the super builder shouldn't be used standalone anyway.

BomBardyGamer: Actually, I see why I didn't do this. Let me just override all of these methods instead.

BomBardyGamer:
As the Moshi serializer pull request was rejected, and also because alternative serializers are not really a requirement, nor an investment that is really worth it for now, I am henceforth closing this pull request, as the use case it used to serve no longer exists.

If anyone wants to play around with this or use this code on their own, feel free to use it, as I am henceforth releasing this in to the public domain.

PR #429, opened 2021-08-26
 feat(api): Pointers supplier

### LINKED ISSUE #428: Per-type pointer checking
Whilst implementing the pointer system in various platforms, it became clear that the system creates a considerable amount of duplicated objects that perform identical functions.

Taking Paper's implementation as an example, each `Player` object will hold a `Pointers` object that delegates to the same set of functions, regardless of the player.

A per-type pointer registry system could alleviate this issue by registering a type with one pointer that provides data from pointers when given a specific instance of an object.

kezz:
This PR closes #428 by providing the `PointersSupplier` class which can be stored statically and resolved whenever `Audience#pointers()` is called to prevent having a bazillion hash maps lying about that all do the same thing.

When implemented, it would look something like this:

```java
 public class MyPointeredObject extends SomePointeredParent implements Pointered {
  private static final PointersSupplier<MyPointeredObject> POINTERS = PointersSupplier.builder()
    .parent(SomePointeredParent.POINTERS) // Fallback to the parent to get pointers from.
    .resolving(Identity.UUID, MyPointeredObject::getUniqueId)
    .resolving(Identity.DISPLAY_NAME, MyPointeredObject::getDisplayName)
    .build();

  @Override
  public Pointers pointers() {
    return POINTERS.view(this);
  }
}
```

### DIFF
--- a/api/src/main/java/net/kyori/adventure/pointer/PointersSupplier.java
+++ b/api/src/main/java/net/kyori/adventure/pointer/PointersSupplier.java
@@ -35,20 +35,20 @@ import org.jetbrains.annotations.Nullable;
  * each instance of a given type.
  *
  * <p>An example of how this could be implemented is as follows:</p>
- * <pre>{@code
+ * <pre>
  * public class MyPointeredObject extends SomePointeredParent implements Pointered {
- *   private static final PointersSupplier<MyPointeredObject> POINTERS = PointersSupplier.builder()
+ *   private static final PointersSupplier&lt;MyPointeredObject&gt; POINTERS = PointersSupplier.builder()
  *     .parent(SomePointeredParent.POINTERS) // Fallback to the parent to get pointers from.
  *     .resolving(Identity.UUID, MyPointeredObject::getUniqueId)
  *     .resolving(Identity.DISPLAY_NAME, MyPointeredObject::getDisplayName)
  *     .build();
  *
- *   @Override
+ *   &#64;Override
  *   public Pointers pointers() {
  *     return POINTERS.view(this);
  *   }
  * }
- * }</pre>
+ * </pre>
  *
  * @param <T> the type
  * @since 4.17.0
--- a/api/src/main/java/net/kyori/adventure/pointer/PointersSupplier.java
+++ b/api/src/main/java/net/kyori/adventure/pointer/PointersSupplier.java
@@ -1,0 +1,127 @@
+/*
+ * This file is part of adventure, licensed under the MIT License.
+ *
+ * Copyright (c) 2017-2024 KyoriPowered
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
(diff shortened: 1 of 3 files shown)


zml2008:
Looks good -- probably more useful for native implementations than `-platform` because -platform is more dynamic about which sources it adds pointers from.

One thing i'd like to see is some sort of 'inheritance' functionality. `toBuilder()` is close, but has a type parameter of `T` rather than `U extends T`, which would prevent it from being used to, for example, have a `PointersSupplier` for `ServerPlayer` extend from `CommandSource`.

kezz:
> One thing i'd like to see is some sort of 'inheritance' functionality. `toBuilder()` is close, but has a type parameter of `T` rather than `U extends T`, which would prevent it from being used to, for example, have a `PointersSupplier` for `ServerPlayer` extend from `CommandSource`.

Do you think it would make more sense for the builder to have a `fallback` or `parent` method where you pass in your instance of the parent class or it's pointers supplier object?

api/src/main/java/net/kyori/adventure/pointer/PointersSupplier.java
```diff
@@ -0,0 +1,78 @@
...
+import org.jetbrains.annotations.NotNull;
+
+/**
+ * A supplier of {@link Pointers} that allows for the implementation of pointers
+ * without creating a new pointers instance for each instance of a given type.
+ *
+ * @param <T> the type
+ * @since 4.9.0
+ */
+public interface PointersSupplier<T> extends Buildable<PointersSupplier<T>, PointersSupplier.Builder<T>> {
```
kashike: We have been phasing out the use of `Buildable`.
```suggestion
public interface PointersSupplier<T> {
```

kashike (changes requested):

zml2008:
> Do you think it would make more sense for the builder to have a fallback or parent method where you pass in your instance of the parent class or it's pointers supplier object?

@kezz I think that would be a reasonable option

kezz:
Rebased and addressed zml's comments. You can now pass a parent to the supplier builder - see the PR description for an updated example.

zml2008 (approved):

api/src/main/java/net/kyori/adventure/pointer/PointersSupplier.java
```diff
@@ -0,0 +1,127 @@
...
+ */
+public interface PointersSupplier<T> {
+  /**
+   * Gets a new pointers supplier builder.
+   *
+   * @param <T> the type
+   * @return the builder
+   * @since 4.17.0
+   */
+  static <T> @NotNull Builder<T> builder() {
```
zml2008, line 64: actually, should this instead be `pointersSupplier`?

kezz: im not sure, there's a lot of places we just use `builder` outside, any thoughts?

zml2008: as much as inconsistency is the spice of life, this would match our `Component` builder methods better. @kashike what do you think?

zml2008: ok, I think we're fine with this, I'm happy to just get this PR wrapped up

PR #433, opened 2021-08-29
text-serializer-legacy: Expose LegacyFormat in a more useful way

kezz:


kezz:
Closing this as I don't want to provide more API for horrible legacy stuff when people should be moving away from this. For reference, a `Map` works fine too ;p

PR #439, opened 2021-09-06
api: Be more strict about validating nullness of appended components

### LINKED ISSUE #427: Enforce nullability checks on component appending
At the moment, the nullability of components is not enforced when appending components, meaning that the children of a component may contain null elements.

zml2008:
Fixes #427

### DIFF
--- a/api/src/main/java/net/kyori/adventure/text/AbstractComponentBuilder.java
+++ b/api/src/main/java/net/kyori/adventure/text/AbstractComponentBuilder.java
@@ -39,6 +39,8 @@ import net.kyori.adventure.text.format.TextDecoration;
 import org.jetbrains.annotations.NotNull;
 import org.jetbrains.annotations.Nullable;
 
+import static java.util.Objects.requireNonNull;
+
 /**
  * An abstract implementation of a component builder.
  *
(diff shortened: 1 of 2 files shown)


api/src/main/java/net/kyori/adventure/text/AbstractComponentBuilder.java
```diff
@@ -74,13 +76,14 @@ protected AbstractComponentBuilder(final @NotNull C component) {
   public @NotNull B append(final @NotNull Component component) {
     if (component == Component.empty()) return (B) this;
     this.prepareChildren();
-    this.children.add(component);
+    this.children.add(requireNonNull(component, "component"));
```
kashike, line 79: Should move the `requireNonNull` above the `prepareChildren` call instead.

api/src/main/java/net/kyori/adventure/text/ScopedComponent.java
```diff
@@ -77,7 +79,7 @@
   default @NotNull C append(final @NotNull Component component) {
     if (component == Component.empty()) return (C) this;
     final List<Component> oldChildren = this.children();
-    return this.children(MonkeyBars.addOne(oldChildren, component));
+    return this.children(MonkeyBars.addOne(oldChildren, requireNonNull(component, "component")));
```
kashike, line 82: Should move the `requireNonNull` above the `oldChildren` variable instead.

kashike (changes requested):

PR #440, opened 2021-09-06
api: Don't override Pointered methods that just delegate

zml2008:
This allows for easier overriding of these methods for forwarding
audiences that have their own pointers/identity.

(needed to implement pointers on things like the `-platform` console)


kashike (approved):
If this is not technically a break in some way, then 👍.

PR #444, opened 2021-09-13
Add ComponentBuider#decorations(Map)

lynxplay:
This commit introduces a new decorations method to the component builder
that takes in a map of text decorations and their state. This method
already exists on the Component interface but was previously missing on
the respective builder.

As the style builder stores the decorations in individual fields and
hence cannot natively work with the map layout, the method was
implemented as a default method on the interface and simply delegates
the map entriess to the normal decorate methods.

The main motivation for the addition of this method is the easy
application of a preset map of decorations.

--------------------------------------------

I followed the current precedence of stating that the mutation modifies the "component" instead of the builder as that seems to be the current practice in all other decoration related methods on the interface.


api/src/main/java/net/kyori/adventure/text/ComponentBuilder.java
```diff
@@ -273,6 +274,24 @@
...
+   * <p>If a given decoration does not have a value explicitly set, the value of that particular decoration is not changed.</p>
+   *
+   * @param decorations a map containing text decorations and their respective state.
+   * @return this builder
+   * @since 4.10.0
+   */
+  @Contract("_ -> this")
+  @SuppressWarnings("unchecked")
+  default @NotNull B decorations(final @NotNull Map<TextDecoration, TextDecoration.State> decorations) {
+    for (Map.Entry<TextDecoration, TextDecoration.State> entry : decorations.entrySet()) {
```
kezz, line 289:
```suggestion
    for (final Map.Entry<TextDecoration, TextDecoration.State> entry : decorations.entrySet()) {
```

kezz (changes requested):
While you're at it, could probably do with this method in `Style.Builder` as well.

lynxplay:
> While you're at it, could probably do with this method in `Style.Builder` as well.

Got you, thanks for the suggestions :+1:

kezz (approved):
Thanks for the PR :)

Proximyst (approved):
Nice!

PR #449, opened 2021-09-21
api: Rename Title.Times.of

### LINKED ISSUE #448: Rename `Title.Times.of`
In order to be consistent with the rest of the API, this method should be named `times`.

NoahvdAa:
Fixes #448.

### DIFF
--- a/api/src/main/java/net/kyori/adventure/title/Title.java
+++ b/api/src/main/java/net/kyori/adventure/title/Title.java
@@ -122,7 +122,7 @@ public interface Title extends Examinable {
      * @since 4.0.0
      * @deprecated for removal since 4.10.0, use {@link #times()}
      */
-    @ApiStatus.ScheduledForRemoval
+    @ApiStatus.ScheduledForRemoval(inVersion = "5.0.0")
     @Deprecated
     static @NotNull Times of(final @NotNull Duration fadeIn, final @NotNull Duration stay, final @NotNull Duration fadeOut) {
       return times(fadeIn, stay, fadeOut);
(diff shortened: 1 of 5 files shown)


api/src/main/java/net/kyori/adventure/title/Title.java
```diff
@@ -120,8 +120,23 @@
      * @param fadeOut the fade-out time
      * @return times
      * @since 4.0.0
+     * @deprecated for removal since 4.10.0, use {@link #times()}
      */
+    @Deprecated
```
kezz, line 125: Should have an `@ApiStatus.ScheduledForRemoval` annotation here.

kezz (changes requested):
Thanks for the PR! This PR should however also replace all usages of the now deprecated method in the codebase.

api/src/main/java/net/kyori/adventure/title/Title.java
```diff
@@ -120,8 +120,24 @@
      * @param fadeOut the fade-out time
      * @return times
      * @since 4.0.0
+     * @deprecated for removal since 4.10.0, use {@link #times()}
      */
+    @ApiStatus.ScheduledForRemoval
```
Proximyst, line 125: Perhaps add an `inVersion` attribute?

kezz: Ooh yes I didn't even know this was a thing! Definitely add one of these with `5.0.0` as the version!

kezz (approved):
Thanks for the PR :)

zml2008 (approved):

PR #465, opened 2021-10-04
api: Add isEmpty to component

Foorcee:
This PR adds ` Component#isEmpty` to easily test if the component is empty.

Like, for example, that an empty component shouldn't be sent to the player.

### DIFF
--- a/api/src/main/java/net/kyori/adventure/text/Component.java
+++ b/api/src/main/java/net/kyori/adventure/text/Component.java
@@ -35,6 +35,7 @@ import java.util.Spliterators;
 import java.util.function.BiPredicate;
 import java.util.function.Consumer;
 import java.util.function.Function;
+import java.util.function.Predicate;
 import java.util.function.UnaryOperator;
 import java.util.regex.Pattern;
 import java.util.stream.Collector;
@@ -114,6 +115,13 @@ public interface Component extends ComponentBuilderApplicable, ComponentLike, Ex
    * @since 4.8.0
    */
   BiPredicate<? super Component, ? super Component> EQUALS_IDENTITY = (a, b) -> a == b;
+
+  /**
+   * A predicate that checks if a component is non-empty.
+   *
+   * @since 4.10.0
+   */
+  Predicate<? super Component> NOT_EMPTY = c -> c != empty();
 
   /**
    * Gets an empty component.
--- a/api/src/main/java/net/kyori/adventure/text/Component.java
+++ b/api/src/main/java/net/kyori/adventure/text/Component.java
@@ -2006,6 +2006,17 @@ public interface Component extends ComponentBuilderApplicable, ComponentLike, Ex
   }
 
   /**
+   * Tests if this {@link Component} is empty.
+   *
+   * @return {@code true} if the component is the empty component, {@code false} if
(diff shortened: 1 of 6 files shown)


zml2008:
I believe this has been discussed in the past in the lead-up to 4.0.0, and we decided against having an isEmpty method, with the plan to fix anything that did incorrectly produce an empty component that was not identity-equal to `empty()`. I think the more appropriate solution would be along the lines of that commit.

The solution used for the children processing (on the line you linked) is not ideal imo -- you're duplicating the work done when constructing the TextComponent later on. This might be a time to pull forward [this change](https://github.com/KyoriPowered/adventure/commit/a75de3acf238fcf500c46517951c271e60810353#diff-eebe19911f1223f3522daeb5e4e83ff05fbf46640a61e2b059a5cfaa0cdd6982R86) from the planned changes for 5.0.0, and do that instead.

api/src/main/java/net/kyori/adventure/text/Component.java
```diff
@@ -115,6 +116,13 @@
    */
   BiPredicate<? super Component, ? super Component> EQUALS_IDENTITY = (a, b) -> a == b;
 
+  /**
```
kezz, line 119: Should be no newline above this variable.

api/src/main/java/net/kyori/adventure/text/Component.java
```diff
@@ -2005,6 +2005,17 @@ default boolean hasStyling() {
     return !this.style().isEmpty();
   }
 
+  /**
+   * Tests if this {@link Component} is empty.
+   *
+   * @return {@code true} if the component is the empty component, {@code false} if
```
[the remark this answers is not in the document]
kezz: Yes, this is definitely correct! The unit tests look like they do a good job of ensuring this as well :)

api/src/main/java/net/kyori/adventure/text/TextComponentImpl.java
```diff
@@ -49,6 +49,13 @@
     return new TextComponentImpl(Collections.emptyList(), Style.empty(), content);
   }
 
+  static @NotNull TextComponent create(final @NotNull List<? extends ComponentLike> children, final @NotNull Style style, final @NotNull String content) {
+    if (children.isEmpty() && style.isEmpty() && content.isEmpty()) {
```
kezz, line 53: There should be a non-null check on `content` here. Also, we could check if each component in the list is empty if the list itself isn't empty? This would ensure that creating a component with empty children also returns the empty component.

kezz (changes requested):

api/src/main/java/net/kyori/adventure/text/Component.java
```diff
@@ -2005,6 +2004,17 @@ default boolean hasStyling() {
     return !this.style().isEmpty();
   }
 
+  /**
+   * Tests if this component is empty.
```
kezz, line 2008: This is kinda misleading, is the component `text("", NamedTextColor.RED)` empty? What about a translatable component that renders to the empty component?

kezz:
Going to close this for now as it appears to be inactive. If you want this reopened in the future please let us know!

PR #472, opened 2021-10-08
Add TriState toBoolean helper methods

lynxplay:
While the TriState enum already has helper methods to quickly convert a
boolean into a TriState (both nullable wrappers and their primitive
counterparts), helper methods to convert a TriState back into a boolean
representation are currently missing.

This commit introduces two methods that do exactly that, one
allowing a straight forward conversion to a Boolean as it can represent
a TriState through its nullability as well as another method that
discards the TriState#NOT_SET by accepting a supplier that will be
used to map a not set tri-state.

---------------------------------------------------------------------------------------------------

I would be **very, very** happy about suggestions/names for `TriState#toBooleanOrElseGet`. I based the current method name off of the `Optional#orElseGet` method that also uses a supplier to map the "not-present" state of an optional, but there might very well be a much better name that I am missing.

### DIFF
--- a/api/src/main/java/net/kyori/adventure/util/TriState.java
+++ b/api/src/main/java/net/kyori/adventure/util/TriState.java
@@ -23,6 +23,7 @@
  */
 package net.kyori.adventure.util;
 
+import java.util.function.BooleanSupplier;
 import org.jetbrains.annotations.NotNull;
 import org.jetbrains.annotations.Nullable;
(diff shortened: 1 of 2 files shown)


lynxplay:
> Might be worth a `toBooleanOrElse` method, but it's not massively necessary, and this is just fine regardless :)

I'll add it :)

PR #480, opened 2021-10-19
extra-kotlin: Added more extensions and enabled explicit API

BomBardyGamer:
This is a cloned branch from #338 because I deleted the fork with the original branch in it.

As mentioned in the original, this adds some more nice extensions to the extra-kotlin module, and enables Kotlin's explicit API feature.

I do plan to add a couple more extensions to this, now I've thought about it, but I will mark it as ready when I've done that.


BomBardyGamer:
Alright then, I'll save them for later.

kezz:
I think that the Kotlin version we expose was left on 1.4 to avoid any breaking changes for those who transitively depend on Kotlin through this module. @zml2008 could confirm this as they touched that last.

zml2008:
that's correct - you can bump the plugin version, but not the language version or stdlib dependency version

kezz (approved):

zml2008 (approved):

PR #481, opened 2021-10-20
fix #447 by allowing deserializeOr to return null

### LINKED ISSUE #447: GsonComponentSerializer doesn't properly handle "null" string
Tested with Adventure 4.9.1.

A Minecraft server can send a set title packet (https://wiki.vg/Protocol#Set_Title_Text) with its contents set as the String `null` and a 1.17.1 vanilla Java client will accept it with no errors. The Minecraft client (tested on 1.17.1 using a Fabric mod) deserializes the Component as literal null:

(Yarn mappings)
```java
@Mixin(ClientPlayNetworkHandler.class)
public class ClassNameLol {

    @Inject(method = "onTitle", at = @At("HEAD"))
    public void onSetTitle(TitleS2CPacket packet, CallbackInfo ci) {
        System.out.println(packet.getTitle()); // will be a literal null in this case
    }
}
```

It appears that this just clears the title in Minecraft 1.17.1 - the same thing that the clear title packet does, except only for the title and not the subtitle.

~~This behavior can be seen with the BetterRTP plugin: https://www.spigotmc.org/resources/betterrtp-random-wild-teleport.36081/ . Installing the plugin and doing `/rtp` will show the issue to any Adventure deserializer listening on the client's packets.~~ This can be replicated on Paper 1.16.5 by executing `Player#sendTitle(null, "")`. It does not manifest with Paper 1.17.1 build 259; it does manifest with Paper 1.16.5 [691 and 788] and https://ci.viaversion.com/job/ViaVersion/553/.

Adventure currently errors on this code as such:

```
com.google.gson.JsonParseException: Don't know how to turn null into a Component
	at net.kyori.adventure.text.serializer.gson.ComponentSerializerImpl.notSureHowToDeserialize(ComponentSerializerImpl.java:323)
	at net.kyori.adventure.text.serializer.gson.GsonComponentSerializerImpl.deserialize(GsonComponentSerializerImpl.java:85)
	at net.kyori.adventure.text.serializer.gson.GsonComponentSerializerImpl.deserialize(GsonComponentSerializerImpl.java:37)
	at com.github.steveice10.mc.protocol.packet.ingame.server.title.ServerSetTitleTextPacket.read(ServerSetTitleTextPacket.java:22)
```

Source code: https://github.com/GeyserMC/MCProtocolLib/blob/feature/1.17.1/src/main/java/com/github/steveice10/mc/protocol/packet/ingame/server/title/ServerSetTitleTextPacket.java#L22 which calls https://github.com/GeyserMC/MCProtocolLib/blob/feature/1.17.1/src/main/java/com/github/steveice10/mc/protocol/data/DefaultComponentSerializer.java#L8

It seems this was initially caused by https://github.com/KyoriPowered/adventure/issues/212 and its commit.

kashike:


zml2008 (approved):

kezz (approved):
Pending style fixes :p

PR #488, opened 2021-10-28
Component::applyFallbackStyle helper method

### LINKED ISSUE #486: Add component constructor to "wrap" another component
Often we are provided with a component externally and want to provide default styles but without overriding any styles on the component provided. In these cases, style merging works well. However, a more developer-friendly method would be nice.

A proposed syntax would be `Component#wrap(ComponentLike, Style)` in addition to a varargs constructor of applicables.

For example, the code for setting italic to false on a MiniMessage-provided component would go from:
```java
Component.text().append(mm).decoration(ITALIC, false).build()
```
to:
```java
Component.wrap(mm, ITALIC.as(false))
```

rymiel:
Implements and resolves #486.

**EDIT**: Method was extensively renamed, see discussion below

**Of note**: I'm not good with javadocs and the explanation is moot at best, please suggest a better wording to describe this "wrapping" feature.

Also, could there possibly be a `default` instance method, to allow for `mm.wrap(ITALIC.as(false))`, or would that be too unclear in its operation?


kashike (dismissed):

astei (approved):

api/src/main/java/net/kyori/adventure/text/Component.java
```diff
@@ -1709,6 +1710,38 @@ default void detectCycle(final @NotNull Component that) {
...
+   *
+   * <p>This method can be used to set the "default" style for a component, whilst still allowing the component to override the style.</p>
+   *
+   * @param like the component which will be wrapped
+   * @param style style to use by default
+   * @return wrapped component
+   * @since 4.10.0
+   */
+  @Contract(pure = true)
+  static @NotNull Component wrap(final @NotNull ComponentLike like, final @NotNull Style style) {
```
kezz, line 1724: The componentlike parameter should just be called `component` instead of like

api/src/main/java/net/kyori/adventure/text/Component.java
```diff
@@ -1709,6 +1710,38 @@ default void detectCycle(final @NotNull Component that) {
     return this.append(builder.build());
   }
 
+  /**
+   * Wraps the provided component in an empty component with the given style.
+   *
+   * <p>This method can be used to set the "default" style for a component, whilst still allowing the component to override the style.</p>
+   *
+   * @param like the component which will be wrapped
+   * @param style style to use by default
```
kezz, line 1719: style for the wrapper component

api/src/main/java/net/kyori/adventure/text/Component.java
```diff
@@ -1709,6 +1710,38 @@ default void detectCycle(final @NotNull Component that) {
...
+   * @return wrapped component
```
kezz, line 1720: the wrapped component

kezz (changes requested):
Final nitpicks :p

api/src/main/java/net/kyori/adventure/text/Component.java
rymiel: was just copying the others ^^

kezz (approved):

zml2008 (changes requested):
I'm not sure if this is the best solution for this sort of operation. What can this `wrap` operator do that an appropriate style merge can't do with fewer levels of nesting?

With an alternative API of something like:

```java
interface Component {
  default @NotNull Component applyUnsetStyleFrom(final @NotNull Style /* or StyleBuilderApplicable... */ other) {
    return this.style(this.style().merge(other, Merge.Strategy.IF_ABSENT_ON_TARGET));
  }
}
```

would there be any limitations compared to this PR? (well, aside from the name)

kezz:
Yes, this component wrapping isn't the ideal solution to the problem it solves. However, it is way more intuitive than other solutions. If we can think of a solid name for this apply unset method, I'd be happy to see a PR with that come through too. However, as it stands I think this PR is solid, and introduces a short-hand for a relatively common pattern.

rymiel:
Should this PR still be considered for 4.10.0?

kezz:
I've changed my mind - I think calling this method `wrap` is good and clear, but it should be performing a style merge.

Think of the case where someone heads onto webui and makes a non-italic lore item. If a developer wants to make that the default, it would be nice to have this method as a no-op in the case where the user has already made the change, for example.

rymiel:
I've made the method perform a style merge as outlined in https://github.com/KyoriPowered/adventure/pull/488#pullrequestreview-829660431, but now it theoretically no longer contains any "wrapper"s per se, so i'm confused on how to document this, although it accomplishes the same thing

zml2008:
`Component.applyFallbackStyle()` maybe?

kezz:
> `Component.applyFallbackStyle()` maybe?

Yeah I kiinda like this.

Would this be worth having in StyleSetter instead of Component?

zml2008 (approved):

PR #489, opened 2021-10-29
Add animations.

codingTheRedCat:
## About
Generally it is the way to  animate playable media. System is designed to be simple and extendable by users. Now it supports Action Bar, Title and Boss Bar animations.

## Structure Concept
The core of animating media is Animation class. It stores frames as list of them, so provides information how to step by step play animation. Animation can be played by animation player - an class implemented by platform which prepares animation for display by setting up scheduler for its frames. When animation is prepared to play the animation display is created. It is an object controlling a process of animation show e.g. stopping it.

_I can always change a little bit the structure, so if it is not ok it will not be a problem for me to correct._

## Tests
I have tested this feature on paper platform. All of the supported frame types seemed to work correctly, but extra test would ensure that.

## Implementation
My plan for implementation is to leave for this stage as little things as its possible. Only think to implement are animation players. Players are classes which have a function to display an animation. Its job is to create scheduler task for animation that will schedule its frames and put it to animation display.


api/src/main/java/net/kyori/adventure/animation/ComponentAnimationImpl.java
```diff
@@ -0,0 +1,52 @@
...
+import java.util.List;
+import net.kyori.adventure.audience.Audience;
+import net.kyori.adventure.text.Component;
+
+class ComponentAnimationImpl extends AbstractAnimation<Component, AnimationDisplay<Component>> implements ComponentAnimation {
+
+  protected ComponentAnimationImpl(final ComponentAnimation[] subAnimations) {
+    super(new ArrayList<>());
+
+    for (final ComponentAnimation animation : subAnimations) frames().addAll(animation.frames());
```
astei, line 36: This isn't going to work, since `AbstractAnimation::frames()` is going to return an unmodifiable list view.

api/src/main/java/net/kyori/adventure/animation/AbstractAnimation.java
```diff
@@ -0,0 +1,52 @@
...
+import java.util.Collections;
+import java.util.Iterator;
+import java.util.List;
+
+abstract class AbstractAnimation<F, D extends AnimationDisplay<? extends F>> implements Animation<F, D> {
+
+  private final List<F> frames;
+
+  protected AbstractAnimation(final List<F> frames) {
+    this.frames = Collections.unmodifiableList(frames);
```
astei, line 36: Defensive copy needed

api/src/main/java/net/kyori/adventure/animation/ComponentAnimation.java
```diff
@@ -0,0 +1,172 @@
...
+  /**
+   * Appends given number of empty component frames to <b>tail</b> of this instance's copy.
+   *
+   * @param amount an amount of frames to append
+   * @return the result of described operation
+   * @see Component#empty()
+   *
+   * @since 4.9.2
+   */
+  default ComponentAnimation appendEmpty(int amount) {
```
astei, line 81: Methods like these seem builder-like and you aren't providing an explicit builder for these.

api/src/main/java/net/kyori/adventure/animation/ComponentAnimation.java
```diff
@@ -0,0 +1,172 @@
...
+public interface ComponentAnimation extends Animation<Component, AnimationDisplay<Component>> {
+
+  /**
+   * Inverts (reverses the order) of frames of copy of this object.
+   *
+   * @return the result of described operation
+   *
+   * @since 4.9.2
+   */
+  default ComponentAnimation invert() {
```
astei, line 48: Why not call this `reversed()`?

api/src/main/java/net/kyori/adventure/animation/LoopableAnimation.java
```diff
@@ -0,0 +1,66 @@
...
+   *
+   * @return true - if animation is looped, else - false.
+   * @since 4.9.2
+   */
+  default boolean looped() {
+    return this.loopIndex() != -1;
+  }
+
+  /**
+   * Returns index after which animation loops.
```
astei, line 47: The magic sentinel value `-1` isn't documented at all here.

api/src/main/java/net/kyori/adventure/animation/AbstractLoopableAnimation.java
```diff
@@ -0,0 +1,62 @@
...
+import java.util.List;
+import net.kyori.adventure.animation.util.LoopedListIterator;
+
+/*
+ * An abstraction of loopable animation.
+ *
+ * @param <F> type of media used as frames
+ * @param <D> type of generated displays used for animation show
+ * @param <T> type of extending interface for chaining
+ * @since 4.9.2
```
astei, line 36: This is clearly a semver-minor change. It would be introduced in Adventure 4.10.0.

api/src/main/java/net/kyori/adventure/animation/Animation.java
```diff
@@ -0,0 +1,276 @@
...
+   * @param subtitleAnimation an animation representing subtitle part
+   * @return new title animation from parameters
+   * @since 4.9.2
+   */
+  static TitleAnimation title(ComponentAnimation titleAnimation, ComponentAnimation subtitleAnimation) {
+    return Animation.title(titleAnimation, 0, subtitleAnimation, Title.DEFAULT_TIMES);
+  }
+
+  /**
+   * Returns a list of frames contained by this animation.
```
astei, line 251: Is this list mutable or immutable? Generally, mutable lists are frowned upon by the Adventure API unless it is absolutely unavoidable (such as builders or boss bar objects).

api/src/main/java/net/kyori/adventure/animation/Animation.java
```diff
@@ -0,0 +1,276 @@
...
+   *
+   * @param audience the viewer of the display
+   * @param schedulerTask task for scheduling frames
+   * @return created display
+   * @since 4.9.2
+   */
+  D createDisplay(Audience audience, AnimationTask schedulerTask);
+
+  /**
+   * Returns iterator that specifies the way of getting frames in correct order for this animation.
```
astei, line 269: Is the iterator mutable?

api/src/main/java/net/kyori/adventure/animation/AnimationTask.java
```diff
@@ -0,0 +1,59 @@
...
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+ * SOFTWARE.
+ */
+package net.kyori.adventure.animation;
+
+import java.time.Duration;
+
+/**
+ * Represent platform scheduler task that arranges in time display of particular animation frames.
```
astei, line 29: This interface veers quite a bit into being mutable state and it doesn't seem to fit in with other components in Adventure.

api/src/main/java/net/kyori/adventure/animation/util/LoopedListIterator.java
```diff
@@ -0,0 +1,91 @@
...
+import java.util.Iterator;
+import java.util.List;
+
+/**
+ * A list iterator, that loops after specified index.
+ *
+ * @param <E> type of the iterated elements
+ * @since 4.9.2
+ */
+public final class LoopedListIterator<E> implements Iterator<E> {
```
astei, line 35: `CircularListIterator`

api/src/main/java/net/kyori/adventure/animation/util/LoopedListIterator.java
```diff
@@ -0,0 +1,91 @@
...
+
+  private LoopedListIterator(final List<E> list, final int loopIndex) {
+    this.list = list;
+    this.loopIndex = loopIndex;
+    if (loopIndex >= list.size() || loopIndex < 0)
+      throw new IllegalArgumentException("Loop index must be positive number less than list size");
+  }
+
+  /**
+   * This iterator is looped, so it always has next element.
```
astei, line 64: Yes, but there is also no guard to see if the given list has been mutated in the meanwhile, or even a check to guarantee that the list even has items, let alone the index item being looped from.

api/src/main/java/net/kyori/adventure/animation/ComponentAnimation.java
```diff
@@ -0,0 +1,172 @@
...
+import net.kyori.adventure.text.Component;
+
+/**
+ * Represents an animation in which frames are components.
+ *
+ * <p>Component animation does not create displays, but it can create other animations with component as frames.</p>
+ *
+ * @since 4.9.2
+ */
+public interface ComponentAnimation extends Animation<Component, AnimationDisplay<Component>> {
```
astei, line 39: Calling this an "animation" sounds like a misnomer to me. This appears to me more like a list of components. I would consider calling this `ComponentAnimationFrames` or `ComponentAnimationFrameList`.

astei (changes requested):
I have some fairly serious concerns about this API being included in Adventure proper - IMO, the Adventure API is intended to be a provide a presentation layer, and animations do not belong in the core API. I would strongly suggest splitting the animation API into a separate module called something like `extra-animation`.

I also have some fairly major code quality and API design concerns.

astei:
One last thing - there is a complete lack of unit tests here. They will have to be added before this PR could be merged.

kezz (changes requested):
Thanks for the PR!

Some general points to address:
- Move to an `extra-animation` module, as discussed previously.
- Style clean up (erroneous newlines mainly - see checkstyle and the style of other Adventure classes/interfaces).
- Unit tests to ensure everything works correctly.
- Nullability annotations are missing.
- Since tags are incorrect (although this can be fixed later, not a priority at this point).
- Examples of how the system works, documentation/etc.
- Proper, robust nullability validation (to ensure exceptions are thrown as soon as possible, especially important in a scheduler system so that exceptions don't have to be handled by any implementation).

However, I'd personally much prefer if you mocked up some examples of how such a system would be used, showcasing the design and usability of the API first, and discussed this with us in Discord so we could share ideas/etc. 

I also believe that the second most important part of implementing a good animation API would will be an easy to implement API. Some idea of how exactly animations would be implemented by a native platform would be ideal, as it would help to be able to abstract out common logic, provide helper methods, standard executor interop, etc.

codingTheRedCat:
Thanks for review I will tru to corvette some features tomorrow.

api/src/main/java/net/kyori/adventure/animation/ComponentAnimation.java
codingTheRedCat: What about `ComponentFramesContainer`?

api/src/main/java/net/kyori/adventure/animation/util/LoopedListIterator.java
codingTheRedCat: Its not really circular. It may have a linear beginning.

astei: No, the iterator is clearly _circular_ in nature. We define _circular_ as a list or iterator whose end points back to the beginning of the list. Suppose the list `L` looks like this:

```
+-----+       +-----+       +-----+
|  1  |  ==>  |  2  |  ==>  |  3  |
+-----+       +-----+       +-----+
```

We define an iterator as a function `I(L) => (N, F(N))` where `N` is the current node to visit and `F(N)` is a function that returns the next node in the list as directed by `I(L)`. Suppose that we use this iterator `I_A(L)` to over over this list multiple times. It is clear that `I_A(L)` wraps another `I(L)`, in fact we can define its behavior like so:

image

The only question is how to define `F_A(F(N))`, and it can be defined as a recursive function:

image

We thus conclude this iterator is circular by definition.

api/src/main/java/net/kyori/adventure/animation/ComponentAnimation.java
astei: That would be a good name!

api/src/main/java/net/kyori/adventure/animation/AnimationTask.java
codingTheRedCat: Its true, but I have no idea how to provide a level of abstraction for animation display scheduling and control different way.

ghost:
I have changed the structure of animation api. I also added 101 tests.

KingOfSquares:
You seem to have changed formatting in unrelated files, might want to revert that.

ghost:
Ok, that is my new version of animation system. I provided documentation in markdown file (I can rewrite it to other format of documentation later. You can tell me what you thing about this and i will try to correct mistakes.

ghost:
> You seem to have changed formatting in unrelated files, might want to revert that.

I think its due my fork sync with orginal. Idk why this occured. I will try to fix that.

ghost:
It's fixed now. Only related changes are committed.

zml2008:
Thank you for your time on this pull request. We think the latest changes are some very good improvements, and I'm sure this API will be interesting for a lot of people.

Unfortunately, the team doesn't currently have the bandwidth to take on the extra maintenance burden needed to finish integrating and properly support this API in the long term, so we will have to close this PR.

We think this animation API would be a great choice to develop further as a standalone library though. If you choose to do so, we would also encourage you to contribute it to the [Community Libraries](https://docs.adventure.kyori.net/community-libraries.html) page in our documentation once you feel it's ready, so that others can hear about it. If you run across any smaller additions or changes that need to be made to `adventure-api` while working on your project, we'll be happy to consider those contributions as well -- just open an issue or ask in our Discord.

PR #490, opened 2021-10-31
api: add TextDecoration#withState and add missing null pointer check

### LINKED ISSUE #431: Provide an alternative for `TextDecoration#as(boolean)`
On Kotlin, `as` is a protected keyword which means you can't do `TextDecoration.ITALIC.as(false)` without doing the ugly ``TextDecoration.ITALIC.`as`(false)``.

This could either exist in the API or in the Kotlin module, although preferably the former.

astei:
Fixes #431. While we're at it, I noticed there was no null check for the state in the `TextDecorationAndStateImpl` constructor so I fixed that too.

### DIFF
--- a/api/src/main/java/net/kyori/adventure/text/format/TextDecoration.java
+++ b/api/src/main/java/net/kyori/adventure/text/format/TextDecoration.java
@@ -107,7 +107,7 @@ public enum TextDecoration implements StyleBuilderApplicable, TextFormat {
    *
    * @param state the state
    * @return a {@link TextDecorationAndState}
-   * @since 4.8.0
+   * @since 4.10.0
    */
   public final @NotNull TextDecorationAndState withState(final boolean state) {
     return this.as(State.byBoolean(state));
@@ -118,7 +118,7 @@ public enum TextDecoration implements StyleBuilderApplicable, TextFormat {
    *
    * @param state the state
    * @return a {@link TextDecorationAndState}
-   * @since 4.8.0
+   * @since 4.10.0
    */
   public final @NotNull TextDecorationAndState withState(final @NotNull State state) {
     return this.as(state);
--- a/api/src/main/java/net/kyori/adventure/text/format/TextDecoration.java
+++ b/api/src/main/java/net/kyori/adventure/text/format/TextDecoration.java
@@ -81,7 +81,7 @@ public enum TextDecoration implements StyleBuilderApplicable, TextFormat {
   }
 
   /**
-   * Creates a {@link TextDecorationAndState}.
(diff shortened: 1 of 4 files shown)


api/src/main/java/net/kyori/adventure/text/format/TextDecoration.java
```diff
@@ -99,6 +99,28 @@
     return new TextDecorationAndStateImpl(this, state);
   }
 
+  /**
+   * An alias for {@link #as(boolean)}.
+   *
+   * @param state the state
+   * @return a {@link TextDecorationAndState}
+   * @since 4.8.0
```
kashike, line 107:
```suggestion
   * @since 4.10.0
```

api/src/main/java/net/kyori/adventure/text/format/TextDecoration.java
```diff
@@ -99,6 +99,28 @@
...
+  public final @NotNull TextDecorationAndState withState(final boolean state) {
+    return this.as(State.byBoolean(state));
+  }
+
+  /**
+   * An alias for {@link #as(State)}.
+   *
+   * @param state the state
+   * @return a {@link TextDecorationAndState}
+   * @since 4.8.0
```
kashike, line 118:
```suggestion
   * @since 4.10.0
```

kashike (changes requested):

kezz (approved):
I wonder if we should deprecate the `as` method because `withState` is much clearer? Either way I'm happy though.

api/src/main/java/net/kyori/adventure/text/format/TextDecoration.java
```diff
@@ -99,6 +99,28 @@
...
+  /**
+   * An alias for {@link #as(State)}.
+   *
+   * @param state the state
+   * @return a {@link TextDecorationAndState}
+   * @since 4.10.0
+   */
+  public final @NotNull TextDecorationAndState withState(final @NotNull State state) {
+    return this.as(state);
+  }
```
zml2008, line 122: would you be able to add `withState(TriState)` (and `as(TriState)`) overloads here? kinda scope creep but merging #495 sparked the thought

zml2008 (changes requested):

api/src/main/java/net/kyori/adventure/text/format/TextDecoration.java
kezz: Do we need an `as(TriState)` overload? As it's new API we don't need another method here that does the same thing. Although I'm also for deprecating the existing `as` method, so that might change if we decide not to.

zml2008 (approved):

kashike (approved):

PR #495, opened 2021-11-03
Add Tristate method to TextDecoration.State

KingOfSquares:
Felt like a naturally missing piece

### DIFF
--- a/api/src/main/java/net/kyori/adventure/text/format/TextDecoration.java
+++ b/api/src/main/java/net/kyori/adventure/text/format/TextDecoration.java
@@ -29,6 +29,8 @@ import net.kyori.adventure.util.TriState;
 import org.jetbrains.annotations.NotNull;
 import org.jetbrains.annotations.Nullable;
 
+import static java.util.Objects.requireNonNull;
+
 /**
  * An enumeration of decorations which may be applied to a {@link Component}.
  *
@@ -176,6 +178,7 @@ public enum TextDecoration implements StyleBuilderApplicable, TextFormat {
      * @since 4.10.0
      */
     public static @NotNull State byTriState(final @NotNull TriState flag) {
+      requireNonNull(flag);
       return byBoolean(flag.toBoolean());
     }
   }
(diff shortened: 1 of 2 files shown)


api/src/main/java/net/kyori/adventure/text/format/TextDecoration.java
```diff
@@ -166,5 +167,16 @@ public String toString() {
...
+
+    /**
+     * Gets a state from a {@link net.kyori.adventure.util.TriState}.
+     *
+     * @param flag the tristate
+     * @return the state
+     * @since 4.10.0
+     */
+    public static @NotNull State byTriState(final @NotNull TriState flag) {
+      return byBoolean(flag.toBoolean());
```
kezz, line 179: Should have a require not null check on the flag

kezz (changes requested):

api/src/main/java/net/kyori/adventure/text/format/TextDecoration.java
```diff
@@ -166,5 +169,17 @@ public String toString() {
...
+    /**
+     * Gets a state from a {@link net.kyori.adventure.util.TriState}.
+     *
+     * @param flag the tristate
+     * @return the state
+     * @since 4.10.0
+     */
+    public static @NotNull State byTriState(final @NotNull TriState flag) {
+      requireNonNull(flag);
+      return byBoolean(flag.toBoolean());
```
kashike, line 182: Think I'd prefer a `switch` here.

kashike (changes requested):

kezz (approved):

BomBardyGamer:
Wouldn't it also be useful to add a `asTriState` conversion method, so you can convert a state to its tri state representation?

kezz:
> Wouldn't it also be useful to add a `asTriState` conversion method, so you can convert a state to its tri state representation?

I don't think so. This comes back to the reason they aren't the same in the first place. A `TriState` will never have its values change but a `TextDecoration.State` might.

kashike (approved):

zml2008 (approved):

PR #498, opened 2021-11-05
Fix unquoted string being parsed as double

kennytv:


nbt/src/main/java/net/kyori/adventure/nbt/TagStringReader.java
```diff
@@ -288,10 +288,12 @@ private BinaryTag scalar() {
       try {
         return IntBinaryTag.of(Integer.parseInt(built));
       } catch (final NumberFormatException ex) {
-        try {
-          return DoubleBinaryTag.of(Double.parseDouble(built));
-        } catch (final NumberFormatException ex2) {
-          // ignore
+        if (built.indexOf('.') != -1) {
```
kashike, line 291: Perhaps add a comment here explaining why.

kashike (changes requested):

kashike (approved):

PR #501, opened 2021-11-21
Abstract out style methods to StyleWriter/Reader

### LINKED ISSUE #491: Abstract out common "style" logic
There are three objects which can be styles; `Style`, `Component` and `ComponentBuilder`. Each of these three classes contain different methods for setting elements of the style and duplicated implementation details.

If style logic could be abstracted into a `Styleable`(?) interface, the API could be unified and the implementation of some methods could be shared. Additionally, this would allow for better abstraction, aiding developers who are manipulating the style of objects.

KingOfSquares:
Closes #491

Also solves some other minor inconsistencies like:
- Adds font methods to `Component`
- Adds the `#decorate(Set<TextDecoration> decorations, boolean flag)` everywhere (Not just `ComponentBuilder`)
- Adds `@ApiStatus.NonExtendable` to `ComponentBuilder` (in addition to the new interfaces)

Should these get their own PR? Excluding the second one

### DIFF
--- a/api/src/main/java/net/kyori/adventure/text/format/StyleReader.java
+++ b/api/src/main/java/net/kyori/adventure/text/format/StyleReader.java
@@ -41,7 +41,6 @@ import org.jetbrains.annotations.Unmodifiable;
  */
 @ApiStatus.NonExtendable
 public interface StyleReader {
-  @ApiStatus.Internal TextDecoration[] DECORATIONS = TextDecoration.values();
 
   /**
    * Gets the font.
@@ -92,8 +91,8 @@ public interface StyleReader {
   @SuppressWarnings("Duplicates")
   default @Unmodifiable @NotNull Map<TextDecoration, TextDecoration.State> decorations() {
     final Map<TextDecoration, TextDecoration.State> decorations = new EnumMap<>(TextDecoration.class);
-    for (int i = 0, length = DECORATIONS.length; i < length; i++) {
-      final TextDecoration decoration = DECORATIONS[i];
+    for (int i = 0, length = StyleImpl.DECORATIONS.length; i < length; i++) {
+      final TextDecoration decoration = StyleImpl.DECORATIONS[i];
       final TextDecoration.State value = this.decoration(decoration);
       decorations.put(decoration, value);
     }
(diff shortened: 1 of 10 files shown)


api/src/main/java/net/kyori/adventure/text/format/StyleWriter.java
```diff
@@ -0,0 +1,178 @@
...
+import net.kyori.adventure.text.event.HoverEventSource;
+import org.jetbrains.annotations.ApiStatus;
+import org.jetbrains.annotations.NotNull;
+import org.jetbrains.annotations.Nullable;
+
+/**
+ * Writes style properties to an object.
+ *
+ * @param <T> The type implementing this interface, e.g. {@link net.kyori.adventure.text.Component}
+ * @see net.kyori.adventure.text.format.Style
```
kashike, line 42:
```suggestion
 * @see Style
```

api/src/main/java/net/kyori/adventure/text/format/StyleWriter.java
```diff
@@ -0,0 +1,178 @@
...
+import net.kyori.adventure.text.event.ClickEvent;
+import net.kyori.adventure.text.event.HoverEventSource;
+import org.jetbrains.annotations.ApiStatus;
+import org.jetbrains.annotations.NotNull;
+import org.jetbrains.annotations.Nullable;
+
+/**
+ * Writes style properties to an object.
+ *
+ * @param <T> The type implementing this interface, e.g. {@link net.kyori.adventure.text.Component}
```
kashike, line 41:
```suggestion
 * @param <T> the type implementing this interface, e.g. {@link Component}
```

api/src/main/java/net/kyori/adventure/text/format/StyleReader.java
```diff
@@ -0,0 +1,126 @@
...
+
+/**
+ * Reads style properties from an object.
+ *
+ * @see net.kyori.adventure.text.format.Style
+ * @since 4.10.0
+ */
+@ApiStatus.NonExtendable
+public interface StyleReader {
+  @ApiStatus.Internal TextDecoration[] DECORATIONS = TextDecoration.values();
```
kashike, line 44: Access this from a package-private class instead.

kashike (changes requested):

zml2008 (changes requested):
The general concept/impl looks good here. However, in `adventure-nbt` we use a `Setter` suffix -- see `ListTagSetter` and `CompoundTagSetter`. Do we want to change these names to `StyleGetter` and `StyleSetter` to match?

zml2008 (approved):

kashike (approved):

PR #503, opened 2021-11-24
feat(pointer): add pointers() method to Pointers collection

Silthus:
It is now possible to query a Pointers collection for all contained Pointer.
This can be useful to create dynamic replacements for the pointer of a Pointered object.

### DIFF
--- a/api/src/main/java/net/kyori/adventure/pointer/Pointers.java
+++ b/api/src/main/java/net/kyori/adventure/pointer/Pointers.java
@@ -23,13 +23,18 @@
  */
 package net.kyori.adventure.pointer;
 
+import java.util.Collections;
+import java.util.HashSet;
+import java.util.Objects;
 import java.util.Optional;
+import java.util.Set;
 import java.util.function.Supplier;
 import net.kyori.adventure.util.Buildable;
 import org.jetbrains.annotations.Contract;
 import org.jetbrains.annotations.NotNull;
 import org.jetbrains.annotations.Nullable;
 import org.jetbrains.annotations.UnknownNullability;
+import org.jetbrains.annotations.Unmodifiable;
 
 /**
  * A collection of {@link Pointer pointers}.
@@ -58,6 +63,36 @@ public interface Pointers extends Buildable<Pointers, Pointers.Builder> {
   @Contract(pure = true)
   static @NotNull Builder builder() {
     return new PointersImpl.BuilderImpl();
+  }
+
+  /**
+   * Gets all {@code pointer} in this collection which have the given value type.
+   *
+   * @param valueType the type class of the pointer
+   * @param <V> the type of the pointer
+   * @return an unmodifiable set of pointers filtered by the given type
+   * @since 4.10.0
+   */
+  @SuppressWarnings("unchecked")
+  default <V> @NotNull @Unmodifiable Set<Pointer<V>> pointers(final @NotNull Class<V> valueType) {
+    Objects.requireNonNull(valueType, "valueType");
+    final Set<Pointer<V>> filteredPointers = new HashSet<>();
+    for (final Pointer<?> pointer : this.pointers()) {
+      if (valueType.isAssignableFrom(pointer.type()))
+        filteredPointers.add((Pointer<V>) pointer);
+    }
+    return Collections.unmodifiableSet(filteredPointers);
+  }
+
+  /**
+   * Gets all {@code pointer} in this pointer collection.
+   *
+   * @return an unmodifiable set of pointers in this pointer collection
+   * @throws UnsupportedOperationException if the implementing class does not support querying for pointers
+   * @since 4.10.0
+   */
+  default @NotNull @Unmodifiable Set<Pointer<?>> pointers() {
+    throw new UnsupportedOperationException("The underlying Pointers implementation does not support querying for pointers.");
   }
 
   /**
--- a/api/src/main/java/net/kyori/adventure/pointer/PointersImpl.java
+++ b/api/src/main/java/net/kyori/adventure/pointer/PointersImpl.java
@@ -23,16 +23,23 @@
  */
 package net.kyori.adventure.pointer;
 
+import java.util.Collections;
 import java.util.HashMap;
 import java.util.Map;
 import java.util.Objects;
 import java.util.Optional;
+import java.util.Set;
 import java.util.function.Supplier;
 import org.jetbrains.annotations.NotNull;
 import org.jetbrains.annotations.Nullable;
 
 final class PointersImpl implements Pointers {
   static final Pointers EMPTY = new Pointers() {
+    @Override
+    public @NotNull Set<Pointer<?>> pointers() {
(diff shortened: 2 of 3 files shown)


api/src/main/java/net/kyori/adventure/pointer/Pointers.java
```diff
@@ -60,6 +64,28 @@
...
+      .map(pointer -> (Pointer<V>) pointer)
+      .collect(Collectors.toSet()));
+  }
+
+  /**
+   * Gets all {@code pointer} in this pointer collection.
+   *
+   * @return an unmodifiable set of pointers in this pointer collection
+   */
+  @NotNull Set<Pointer<?>> pointers();
```
Silthus, line 87: Not sure if this addition should create a breaking change or if this should be made default and return an empty set.
This however could make the behaviour of the method unpredictable if not overwritten.

api/src/main/java/net/kyori/adventure/pointer/PointersImpl.java
```diff
@@ -23,16 +23,24 @@
...
+import java.util.Set;
 import java.util.function.Supplier;
 import org.jetbrains.annotations.NotNull;
 import org.jetbrains.annotations.Nullable;
 
 final class PointersImpl implements Pointers {
   static final Pointers EMPTY = new Pointers() {
+    @Override
+    public @NotNull Set<Pointer<?>> pointers() {
+      return Collections.unmodifiableSet(new HashSet<>());
```
kezz, line 41: I'm pretty sure there's an empty set method on `Collections`?

api/src/test/java/net/kyori/adventure/pointer/PointersTest.java
```diff
@@ -23,10 +23,18 @@
  */
 package net.kyori.adventure.pointer;
 
+import java.util.Set;
 import java.util.function.Supplier;
+
```
kezz, line 28: Don't forget to run checkstyle :p

api/src/main/java/net/kyori/adventure/pointer/Pointers.java
kezz: Hmm yes, annoyingly we need to return something here. Trying to work out if we could make this interface non-extendable. Not quite sure...

api/src/main/java/net/kyori/adventure/pointer/Pointers.java
```diff
@@ -60,6 +63,30 @@
...
+   * Gets all {@code pointer} in this collection which have the given value type.
+   *
+   * @param valueType the type class of the pointer
+   * @param <V> the type of the pointer
+   * @return an unmodifiable set of pointers filtered by the given type
+   * @since 4.10.0
+   */
+  @SuppressWarnings("unchecked")
+  default <V> @NotNull Set<Pointer<V>> pointers(final @NotNull Class<V> valueType) {
+    return Collections.unmodifiableSet(this.pointers().stream()
```
kezz, line 76: We probably don't need to use streams here, we can construct it without these.

api/src/main/java/net/kyori/adventure/pointer/Pointers.java
```diff
@@ -60,6 +63,30 @@
...
+  /**
+   * Gets all {@code pointer} in this collection which have the given value type.
+   *
+   * @param valueType the type class of the pointer
+   * @param <V> the type of the pointer
+   * @return an unmodifiable set of pointers filtered by the given type
+   * @since 4.10.0
+   */
+  @SuppressWarnings("unchecked")
+  default <V> @NotNull Set<Pointer<V>> pointers(final @NotNull Class<V> valueType) {
```
kezz, line 75: Is there an unmodifiable annotation that could fit here or am I imagining that?

kezz, line 75: We need to do a non-null check on this parameter.

kezz (changes requested):
Looks good to me! I think this is a good improvement for sure.

api/src/main/java/net/kyori/adventure/pointer/Pointers.java
Silthus: Why not use streams? Is there a specific reason behind it? I find it a lot more readable.

Silthus:
Here is an example how this can be used:

```java
public interface PointeredResolvable extends Resolvable, Pointered {

    @Override
    @ApiStatus.NonExtendable
    default /*final*/ @NotNull TemplateResolver resolver() {
        return resolverWithPrefix("");
    }

    default /*final*/ @NotNull TemplateResolver resolverWithPrefix(String prefix) {
        return TemplateResolver.combining(
                new StringPointerTemplateResolver(this, getPointerMap(String.class, prefix)),
                new ComponentPointerTemplateResolver(this, getPointerMap(Component.class, prefix))
        );
    }

    private <T> @NotNull Map<String, Pointer<T>> getPointerMap(Class<T> pointerType, String prefix) {
        return pointers().pointers(pointerType).stream()
                .collect(toMap(
                        stringPointer -> prefix + stringPointer.key().value(),
                        stringPointer -> stringPointer
                ));
    }
}
 ```

**Comment by kezz** [comment]
Tbh, I reckon that would be a really good fit for inclusion into MiniMessage after this is merged!

File: api/src/main/java/net/kyori/adventure/pointer/Pointers.java (hunk 4 above)
Comment (electronicboy) [reply api/src/main/java/net/kyori/adventure/pointer/Pointers.java:79]: Streams have overheads, here and there, it's not an issue, but, if we're expecting stuff to be accessing it a lot it can create many needlessly hot areas

File: api/src/main/java/net/kyori/adventure/pointer/Pointers.java [hunk 6]
```diff
@@ -60,6 +65,36 @@
...
+
+  /**
+   * Gets all {@code pointer} in this pointer collection.
+   *
+   * @return an unmodifiable set of pointers in this pointer collection
+   * @throws UnsupportedOperationException if the implementing class does not support querying for pointers
+   * @since 4.10.0
+   */
+  default @NotNull @Unmodifiable Set<Pointer<?>> pointers() {
+    throw new UnsupportedOperationException("The underlying Pointers implementation does not support querying for pointers.");
```
kezz, line 95: Could we not theoretically make this default to `pointers(Object.class)` so we don't have to throw a UOE here?

kezz (changes requested):

api/src/main/java/net/kyori/adventure/pointer/Pointers.java
Silthus: But that would be a endless loop as the `pointers(Class)` requires this method.
We could return an empty set but I thought that would be unpredictable functionality. And the Unsupported Operation only affects custom implementations.

kezz: Ugh yeah... Man I wish we just made this class internal :')

PR #506, opened 2021-11-27
Use switch statement instead of if-else in NamedTextColor#ofExact

Leonardo-DGS:


kashike:
Why?

kezz (approved):

Leonardo-DGS:
> Why?

In this cases is more suited the switch construct, and it's less verbose.

api/src/main/java/net/kyori/adventure/text/format/NamedTextColor.java
```diff
@@ -171,23 +171,25 @@
...
+      case GRAY_VALUE: return GRAY;
+      case DARK_GRAY_VALUE: return DARK_GRAY;
+      case BLUE_VALUE: return BLUE;
+      case GREEN_VALUE: return GREEN;
+      case AQUA_VALUE: return AQUA;
+      case RED_VALUE: return RED;
+      case LIGHT_PURPLE_VALUE: return LIGHT_PURPLE;
+      case YELLOW_VALUE: return YELLOW;
+      case WHITE_VALUE: return WHITE;
+      default: return null;
```
Proximyst, line 191: This shouldn't be included: javac will warn on any non-mapped values if there is no default branch. 😄

rymiel: This isn't an exhaustive switch, it can't be because the argument is an int

Leonardo-DGS: > 

Indeed _there is_ a default branch.

kashike (approved):

PR #617, opened 2021-11-29
api: Introduce Component.textOfChildren as a replacement for TextComponent.ofChildren

kashike:
This indirectly reverts commit 9ffba5b718bd396e6d397c6543c1982d9e631b08.


jpenilla (approved):

kezz (changes requested):
This doesn't make sense. Why not add/undeprecate one of the methods on `Component`? Why is `TextComponent` the only component that gets a constructor method in the interface?

kezz (approved):
Oh didn't spot this was actually on `Compoment` :p

PR #623, opened 2021-12-07
text-minimessage: Reduce the API surface of the parse context a little bit

zml2008:
Can more of these methods be removed? The only truly essential one is imo `parse(String)`, but I'm not sure what sort of usecases others might have

### DIFF
--- a/text-minimessage/src/main/java/net/kyori/adventure/text/minimessage/Context.java
+++ b/text-minimessage/src/main/java/net/kyori/adventure/text/minimessage/Context.java
@@ -24,6 +24,7 @@
 package net.kyori.adventure.text.minimessage;
 
 import net.kyori.adventure.text.Component;
+import org.jetbrains.annotations.ApiStatus;
 
 /**
  * Parser context for use within transformations.
@@ -32,6 +33,7 @@ import net.kyori.adventure.text.Component;
  *
  * @since 4.10.0
  */
+@ApiStatus.NonExtendable
 public interface Context {
   /**
    * Returns original message as provided to the parser.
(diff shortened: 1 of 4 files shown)


kashike:
It's definitely easier to add methods as-required, versus deprecating and removing methods later on.

text-minimessage/src/main/java/net/kyori/adventure/text/minimessage/Context.java
```diff
@@ -23,226 +23,23 @@
...
-   * Returns the root element.
-   *
-   * @return root
-   * @since 4.10.0
-   */
-  public ElementNode tokens() {
-    return this.root;
-  }
-
+public interface Context {
```
kezz, line 35: Should we annotate this with a non-extendible notice?

kezz (changes requested):

text-minimessage/src/main/java/net/kyori/adventure/text/minimessage/Context.java
```diff
@@ -23,226 +23,25 @@
...
-  /**
-   * Returns callback ran at the end of parsing which could be used to compact the output.
-   *
-   * @return Post-processing function
-   * @since 4.10.0
-   */
-  public UnaryOperator<Component> postProcessor() {
-    return this.postProcessor;
-  }
+  String originalMessage();
```
kezz, line 44:
```suggestion
  @NotNull String originalMessage();
```

text-minimessage/src/main/java/net/kyori/adventure/text/minimessage/Context.java
```diff
@@ -251,7 +50,5 @@ public MiniMessage miniMessage() {
    * @return the parsed message
    * @since 4.10.0
    */
-  public Component parse(final String message) {
-    return this.miniMessage.deserialize(message, this.placeholderResolver);
-  }
+  Component parse(String message);
```
kezz, line 53:
```suggestion
  @NotNull Component parse(final @NotNull String message);
```

kezz (changes requested):

kezz (approved):

PR #632, opened 2021-12-11
feature: Placeholder cleanup

### LINKED ISSUE #517: Add a distinction between "plain" and "minimessage" string placeholders

### LINKED ISSUE #518: Remove the key parameter from templates

kezz:
This PR provides a cleanup of the placeholder system and closes #518 and closes #517.

The following changes are made:
- Remove the hacky placeholder resolver methods in favour of forcing people to not use weird hacky methods.
- Swap placeholder resolvers to return a new class, `Replacement` instead, thus removing the need for a key.
- Change the map placeholder resolver constructor to use this new replacement class.
- Remove the `LazyComponentPlaceholder` as it is no longer needed due to the resolver system being lazy, if desired.
- Remove `PlaceholderResolver#canResolve` to allow easier implementation of `PlaceholderResolver` due to it being a functional interface.

Essentially, this is a move to a placeholder system that is safe at compile time. This means that (unless you ignore `@ApiStatus.NonExtendable`) you cannot construct any placeholder resolver that will fail at runtime (for example, if you have an integer replacement).

### DIFF
--- a/text-minimessage/src/main/java/net/kyori/adventure/text/minimessage/placeholder/Placeholder.java
+++ b/text-minimessage/src/main/java/net/kyori/adventure/text/minimessage/placeholder/Placeholder.java
@@ -79,9 +79,7 @@ public interface Placeholder<T> extends Replacement<T> {
   /**
    * Get the key for this placeholder.
    *
-   * <p>
-   *   The key will always be lowercase.
-   * </p>
+   * <p>The key will always be lowercase.</p>
    *
    * @return the key
    * @since 4.10.0
--- a/text-minimessage/src/main/java/net/kyori/adventure/text/minimessage/placeholder/Placeholder.java
+++ b/text-minimessage/src/main/java/net/kyori/adventure/text/minimessage/placeholder/Placeholder.java
@@ -74,21 +74,6 @@ public interface Placeholder extends Examinable {
   }
 
   /**
-   * Constructs a placeholder that gets replaced with a component lazily.
-   *
-   * @param key the placeholder
-   * @param value the supplier that supplies the component to replace the key with
-   * @return the constructed placeholder
-   * @since 4.10.0
-   */
-  static @NotNull Placeholder placeholder(final @NotNull String key, final @NotNull Supplier<? extends ComponentLike> value) {
-    return new LazyComponentPlaceholder(
-        requireNonNull(key, "key"),
-        requireNonNull(value, "value")
-    );
-  }
-
-  /**
    * Get the key for this placeholder.
    *
    * @return the key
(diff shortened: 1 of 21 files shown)


text-minimessage/src/main/java/net/kyori/adventure/text/minimessage/placeholder/Placeholder.java
```diff
@@ -24,195 +24,67 @@
...
     );
   }
 
   /**
    * Get the key for this placeholder.
    *
+   * <p>
+   *   The key will always be lowercase.
+   * </p>
+   *
```
zml2008, line 85:
```suggestion
   * <p>The key will always be lowercase.</p>
   *
```

text-minimessage/src/main/java/net/kyori/adventure/text/minimessage/placeholder/PlaceholderResolver.java
```diff
@@ -40,88 +38,22 @@
...
 public interface PlaceholderResolver {
   /**
-   * Constructs a placeholder resolver from key-value pairs or {@link Placeholder} instances.
+   * Constructs a placeholder resolver from a map.
    *
-   * <p>The {@code pairs} arguments must be a string key followed by a string or {@link ComponentLike} value or a {@link Placeholder}.</p>
+   * <p>
+   *   The provided map is used as the backing for the returned placeholder resolver. This means
+   *   that changes to the map will be reflected in the placeholder resolver.
+   * </p>
```
zml2008, line 49:
```suggestion
   * <p>The provided map is used as the backing for the returned placeholder resolver. This means
   *   that changes to the map will be reflected in the placeholder resolver.</p>
```

text-minimessage/src/main/java/net/kyori/adventure/text/minimessage/placeholder/PlaceholderResolver.java
```diff
@@ -40,88 +38,22 @@
...
    *
-   * @param objects the objects
+   * @param map the map
    * @return the placeholder resolver
    * @since 4.10.0
    */
-  static @NotNull PlaceholderResolver resolving(final @NotNull Object @NotNull ... objects) {
```
zml2008, line 53: I liked having this around -- it's a handy shortcut for placeholders.

I see that you moved it into the test classes below as well. If it's useful enough that you want to keep it available for the tests, why is it not useful enough to be in API?

text-minimessage/src/main/java/net/kyori/adventure/text/minimessage/placeholder/PlaceholderResolver.java
```diff
@@ -194,29 +126,20 @@
   /**
    * Constructs a placeholder resolver capable of dynamically resolving placeholders.
    *
-   * <p>The {@code resolver} function must return instances of String, {@link ComponentLike} or {@link Placeholder}.
-   * The resolver can return {@code null} to indicate it cannot resolve a placeholder.</p>
+   * <p>
+   *   The resolver can return {@code null} to indicate it cannot resolve a placeholder.
+   *   Once a string to replacement mapping has been created, it will be cached to avoid
+   *   the cost of recreating the replacement.
+   * </p>
```
zml2008, line 133:
```suggestion
   * <p>The resolver can return {@code null} to indicate it cannot resolve a placeholder.
   *   Once a string to replacement mapping has been created, it will be cached to avoid
   *   the cost of recreating the replacement.</p>
```

text-minimessage/src/main/java/net/kyori/adventure/text/minimessage/placeholder/PlaceholderResolver.java
```diff
@@ -228,20 +151,17 @@
...
-   */
-  boolean canResolve(final @NotNull String key);
-
-  /**
-   * Returns a placeholder from a given key, if any exist.
+   * <p>
+   *   This method might be called multiple times during each parse attempt. This is due to the
+   *   fact that it is used in places to check if a tag is a placeholder or not. Therefore, you
+   *   should prefer using fixed or cached replacements instead of dynamic construction.
+   * </p>
```
zml2008, line 160:
```suggestion
   * <p>This method might be called multiple times during each parse attempt. This is due to the
   *   fact that it is used in places to check if a tag is a placeholder or not. Therefore, you
   *   should prefer using fixed or cached replacements instead of dynamic construction.</p>
```

zml2008 (changes requested):

zml2008 (approved):

broccolai (approved):

PR #634, opened 2021-12-12
Added NBT Kotlin extensions

BomBardyGamer:
See #340

### DIFF
--- a/nbt-extra-kotlin/src/main/kotlin/net/kyori/adventure/nbt/extra/kotlin/BinaryTagConverters.kt
+++ b/nbt-extra-kotlin/src/main/kotlin/net/kyori/adventure/nbt/extra/kotlin/BinaryTagConverters.kt
@@ -11,22 +11,32 @@ import net.kyori.adventure.nbt.LongBinaryTag
 import net.kyori.adventure.nbt.ShortBinaryTag
 import net.kyori.adventure.nbt.StringBinaryTag
 
+@JvmSynthetic
 public fun Byte.toBinaryTag(): ByteBinaryTag = ByteBinaryTag.of(this)
 
+@JvmSynthetic
 public fun Short.toBinaryTag(): ShortBinaryTag = ShortBinaryTag.of(this)
 
+@JvmSynthetic
 public fun Int.toBinaryTag(): IntBinaryTag = IntBinaryTag.of(this)
 
+@JvmSynthetic
 public fun Long.toBinaryTag(): LongBinaryTag = LongBinaryTag.of(this)
 
+@JvmSynthetic
 public fun Float.toBinaryTag(): FloatBinaryTag = FloatBinaryTag.of(this)
 
+@JvmSynthetic
 public fun Double.toBinaryTag(): DoubleBinaryTag = DoubleBinaryTag.of(this)
 
+@JvmSynthetic
 public fun ByteArray.toBinaryTag(): ByteArrayBinaryTag = ByteArrayBinaryTag.of(*this)
 
+@JvmSynthetic
 public fun IntArray.toBinaryTag(): IntArrayBinaryTag = IntArrayBinaryTag.of(*this)
 
+@JvmSynthetic
 public fun LongArray.toBinaryTag(): LongArrayBinaryTag = LongArrayBinaryTag.of(*this)
 
+@JvmSynthetic
 public fun String.toBinaryTag(): StringBinaryTag = StringBinaryTag.of(this)
--- a/nbt-extra-kotlin/src/main/kotlin/net/kyori/adventure/nbt/extra/kotlin/BinaryTagConverters.kt
+++ b/nbt-extra-kotlin/src/main/kotlin/net/kyori/adventure/nbt/extra/kotlin/BinaryTagConverters.kt
@@ -1,3 +1,26 @@
+/*
+ * This file is part of adventure, licensed under the MIT License.
+ *
+ * Copyright (c) 2017-2021 KyoriPowered
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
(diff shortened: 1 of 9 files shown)


nbt-extra-kotlin/src/main/kotlin/net/kyori/adventure/nbt/extra/kotlin/BinaryTagConverters.kt
```diff
@@ -0,0 +1,32 @@
...
+
+public fun Float.toBinaryTag(): FloatBinaryTag = FloatBinaryTag.of(this)
+
+public fun Double.toBinaryTag(): DoubleBinaryTag = DoubleBinaryTag.of(this)
+
+public fun ByteArray.toBinaryTag(): ByteArrayBinaryTag = ByteArrayBinaryTag.of(*this)
+
+public fun IntArray.toBinaryTag(): IntArrayBinaryTag = IntArrayBinaryTag.of(*this)
+
+public fun LongArray.toBinaryTag(): LongArrayBinaryTag = LongArrayBinaryTag.of(*this)
```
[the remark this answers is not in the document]
BomBardyGamer: For non-bots, I am aware of such a flaw, and there is nothing at all that can be done about it,except for some hacks like creating a class in that package to access constructors that take arrays, and refactoring all of the implementation classes to take arrays rather than varargs, which is something that is not being done.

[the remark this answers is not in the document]
TomMD: @BomBardyGamer Thank you for the feedback. In this case you can ignore the warning by committing a `.lift/config.toml` file with `ignoreRules = [ "SpreadOperator" ]`.  We're on the lookout for how to improve the results and these sorts of feedback are valuable.

nbt-extra-kotlin/src/main/kotlin/net/kyori/adventure/nbt/extra/kotlin/ListBinaryTagConverters.kt
```diff
@@ -0,0 +1,47 @@
...
+import net.kyori.adventure.nbt.IntBinaryTag
+import net.kyori.adventure.nbt.ListBinaryTag
+import net.kyori.adventure.nbt.LongArrayBinaryTag
+import net.kyori.adventure.nbt.LongBinaryTag
+import net.kyori.adventure.nbt.ShortBinaryTag
+import net.kyori.adventure.nbt.StringBinaryTag
+
+public fun <T : BinaryTag> List<T>.toBinaryTag(): ListBinaryTag = ListBinaryTag.from(this)
+
+@JvmName("toBinaryTagByte")
```
kashike, line 19:
```suggestion
@JvmName("toByteBinaryTag")
```

nbt-extra-kotlin/src/main/kotlin/net/kyori/adventure/nbt/extra/kotlin/ListBinaryTagConverters.kt
```diff
@@ -0,0 +1,47 @@
...
+public fun List<Byte>.toBinaryTag(): ListBinaryTag = ListBinaryTag.of(BinaryTagTypes.BYTE, map(ByteBinaryTag::of))
+
+@JvmName("toBinaryTagShort")
```
kashike, line 22:
```suggestion
@JvmName("toShortBinaryTag")
```
... etc (applies to all cases like this)

kashike (changes requested):

nbt-extra-kotlin/src/main/kotlin/net/kyori/adventure/nbt/extra/kotlin/ListBinaryTagConverters.kt
BomBardyGamer: I can change these, though I think that it doesn't really matter what they're called tbh, since it's only designed to make it readable in Java, which we couldn't care less about. I am tempted to go through and do what I did with all of my extensions in another project, which is make them `@JvmSynthetic` so they don't show up in Java at all. I will go ahead and rename these, but I'll also add the synthetics, since I think it's a good idea this stuff is hidden from Java anyway, so that people don't go and accidentally use it where they shouldn't.

BomBardyGamer:
Looking back at this pull request, I think that most of it is likely unnecessary or provides no special benefit over the alternatives that it replaces. I know that the idea that seems to be emphasised by a lot of the Kotlin community is to be idiomatic, but a lot of this doesn't make sense.

Here are the issues I have with this:
1. The `toBinaryTag` extension functions aren't necessary, and the scope and imports from the static functions are better. Also, it is clearer to see that `ByteBinaryTag.of(3)` produces a new binary tag, over `3.toBinaryTag()`. In addition, extension functions on number values like that aren't the nicest to work with imo, especially when expressions are involved, such as `(3 + 7).toBinaryTag()`, where those brackets are required.
2. The `set` extensions:
  * These imply that the tags are mutable, which they are not.
  * The way in which they are used, the property access syntax (`tag[name] = otherTag`), doesn't work with immutable values where `set` returns the new value, as there is no way to get the resulting tag from that, since `=` is not an expression.
3. Converting lists to binary tags with the `toBinaryTag()` extensions is a bit clunky. Especially since the current syntax also doesn't support more complex, multi-dimensional lists, of which would be more complicated to implement and make this not just clunky but ridiculous.
4. The `plus` functions are meant to be used in syntatic sugar, but:
  * They are only present for the lists
  * They only have real usage in either combining two list tags with `+`, or some whacky nonsense like what I have seen before, something like:
    ```kotlin
    // DSL way
    list("name") {
        +someTag
        +someOtherTag
    }

    // Even worse whack using the builder pattern
    (ListBinaryTag.builder() + someTag + someOtherTag).build()
    ```

For these reasons, I have no interest in working on this pull request, and I don't think it should be merged in its current state. If others see usefulness in some of the stuff in this, such as the DSL functions, then feel free to open a separate pull request, and feel free to use my code that I am henceforth releasing in to the public domain.

PR #641, opened 2021-12-14
api: Use initial style when replacement-rendering a hover event

### LINKED ISSUE #638: TextReplacementConfig being weird
The following code:
```java
var replacement = TextReplacementConfig.builder().match("hover").replacement(
	Component.text()
		.content("yeeet")
		.hoverEvent(Component.text("hover!"))
		.build()
).build();
var component1 = Component.text("hover").replaceText(replacement); // text the same as match above
var component2 = Component.text("hover ").replaceText(replacement); // there's space at the end now
player.sendMessage(component1);
player.sendMessage(component2);
```
Produces this result:
image
image
For some reason, TextReplacementConfig gets called twice (?) in the first case which wrongly replaces text inside hover.

Using Paper 1.17.1 build 399 and adventure-text-minimessage 4.10.0-SNAPSHOT.

rymiel:
Fixes #638

This matters if the component is entirely replaced due to an exact match (whereas a non-exact match creates an empty parent with no style)

I'm not 100% sure this won't cause more issues in the future, but i tried to follow how children are implemented with oldChildren

### DIFF
--- a/api/src/main/java/net/kyori/adventure/text/TextReplacementRenderer.java
+++ b/api/src/main/java/net/kyori/adventure/text/TextReplacementRenderer.java
@@ -52,6 +52,7 @@ final class TextReplacementRenderer implements ComponentRenderer<TextReplacement
 
     final List<Component> oldChildren = component.children();
     final int oldChildrenSize = oldChildren.size();
+    final Style oldStyle = component.style();
     List<Component> children = null;
     Component modified = component;
     // replace the component itself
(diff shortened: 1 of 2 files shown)


kezz (approved):

kashike (approved):

zml2008:
how does this handle cases where both the original component and the replacement result have different hover events?

I think the most sensible behaviour would be to use the replacement result's hover event and ignore the incoming one - could you add a test to ensure that's what happens?

zml2008:
the one from the pre-replacement component, sorry

zml2008 (approved):

PR #645, opened 2021-12-16
feat(key): expose a Comparator for Key

kashike:


kezz (approved):

key/src/main/java/net/kyori/adventure/key/Key.java
```diff
@@ -61,6 +62,14 @@
...
    */
   String MINECRAFT_NAMESPACE = "minecraft";
+  /**
+   * The comparator.
+   *
+   * <p>The {@link #value() value} is compared first, followed by the {@link #namespace() namespace}.</p>
+   *
+   * @since 4.10.0
+   */
+  Comparator<? super Key> COMPARATOR = Comparator.comparing(Key::value).thenComparing(Key::namespace);
```
zml2008, line 72: i'd rather use a factory method for this with the actual instance in `KeyImpl`, just in case we need to move away from a constant field for any reason in the future.

zml2008 (changes requested):

zml2008 (approved):

Issue #453, opened 2021-09-26
Create a `Font` interface for use as a `StyleBuilderApplicable`


kezz:
This would allow for component construction like `Component.text("Hi!", NamedTextColor.RED, Font.font(Key.key("minecraft:alt")))` and provide a way people could easily store/generate easily usable fonts.

PR #648, opened 2021-12-19
feat(api): #453 Introduce Font

kashike:


kezz (approved):

KingOfSquares:
Maybe expand a little more than just `Font.` in the Java doc? Something that distinguishes between the representation of the Minecraft font(the class) and its resource location(the key) could be something to expand on. I know the Key is regarded _as_ the font all the other places in the api, so maybe just something about the class not being the same as the contained Key…?

KingOfSquares:
Lighting speed 😎👍

zml2008 (changes requested):
Maybe add an overload of `StyleSetter.font` that accepts instances of `Font`?

lucko (approved):

kezz (approved):

kezz (changes requested):
Actually, the ambiguous overload is a bit of a deal breaker for this - we don't want to force developers to change code/recompile if they are removing a font.

zml2008:
hmm, how awful would it be to bring this in with just `Font`, and not the added method on `StyleSetter`?

PR #650, opened 2021-12-21
feat(api): Internal configuration

kashike:


api/src/main/java/net/kyori/adventure/util/internal/AdventureConfig.java
```diff
@@ -0,0 +1,111 @@
...
+ * @since 4.10.0
+ */
+@ApiStatus.Internal
+public final class AdventureConfig {
+  /**
+   * The directory in which our configuration file lives.
+   *
+   * @since 4.10.0
+   */
+  public static final String FILESYSTEM_DIRECTORY_NAME = "config";
```
kezz, line 43: Is there a reason for these being separate to the file name? Would we ever have a situation where we'd have more than one?

api/src/main/java/net/kyori/adventure/util/internal/AdventureConfigImpl.java
```diff
@@ -0,0 +1,52 @@
...
+package net.kyori.adventure.util.internal;
+
+import java.io.IOException;
+import java.io.InputStream;
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.nio.file.Paths;
+import java.util.Properties;
+
+final class AdventureConfigImpl {
```
kezz, line 33: I don't think this class needs to exist - it's all internal anyway.

api/src/main/java/net/kyori/adventure/util/internal/AdventureConfig.java
kezz, line 43: Might be nicer to have these in an enum that can be loaded directly, so `AdventureConfig.THING.getBoolean()`. Or maybe even a typed interface so we can just `get()` without having to worry about what the type is.

kezz (changes requested):

api/src/main/java/net/kyori/adventure/util/internal/AdventureProperties.java
```diff
@@ -0,0 +1,118 @@
...
+   * Gets a boolean value.
+   *
+   * @param option the option
+   * @param defaultValue the default value
+   * @return the boolean value
+   * @since 4.10.0
+   */
+  public static boolean booleanValueOf(final @NotNull Option<Boolean> option, final boolean defaultValue) {
+    return Boolean.TRUE.equals(valueOf(option, defaultValue));
+  }
```
KingOfSquares, line 83: What about just having a `get()` or `value()` like kezz talked about in another comment in `Option` instead?

That way 
```java
AdventureProperties.booleanValueOf(AdventureProperties.TEXT_WARN_WHEN_LEGACY_FORMATTING_DETECTED, false);
```
could become
```java
AdventureProperties.TEXT_WARN_WHEN_LEGACY_FORMATTING_DETECTED.value();
```

The default value could be decided when each `Option` is parsed.
```java
public static final Option<Boolean> TEXT_WARN_WHEN_LEGACY_FORMATTING_DETECTED = option("text.warnWhenLegacyFormattingDetected", Boolean::parseBoolean, false);
```

api/src/main/java/net/kyori/adventure/util/internal/AdventureProperties.java
```diff
@@ -0,0 +1,118 @@
...
+   * @param option the option
+   * @param defaultValue the default value
+   * @return the boolean value
+   * @since 4.10.0
+   */
+  public static <T> @Nullable T valueOf(final @NotNull Option<T> option, final @Nullable T defaultValue) {
+    final String key = option.name();
+    final String property = String.join(".", "net", "kyori", "adventure", key);
+    final String value = System.getProperty(property, AdventurePropertiesImpl.PROPERTIES.getProperty(key));
+    return value != null ? ((AdventurePropertiesImpl.OptionImpl<T>) option).parser.apply(value) : defaultValue;
```
KingOfSquares, line 97: Also, I agree with kezz about this internal utility not needing interface+impl. That way casting like this could be avoided.

kashike: It may be an internal utility, but it is still technically something visible. I'd rather keep the internals in their own class.

api/src/main/java/net/kyori/adventure/internal/properties/AdventurePropertiesImpl.java
```diff
@@ -0,0 +1,99 @@
...
+import org.jetbrains.annotations.NotNull;
+import org.jetbrains.annotations.Nullable;
+
+final class AdventurePropertiesImpl {
+  private static final String FILESYSTEM_DIRECTORY_NAME = "config";
+  private static final String FILESYSTEM_FILE_NAME = "adventure.properties";
+  private static final Properties PROPERTIES = new Properties();
+
+  static {
+    final Path path = Paths.get(FILESYSTEM_DIRECTORY_NAME, FILESYSTEM_FILE_NAME);
```
zml2008, line 42: Is it reasonable to have a system property to specify a specific file to read from, rather than just this specific location?

zml2008 (changes requested):

zml2008 (approved):
lgtm assuming checks pass

api/src/main/java/net/kyori/adventure/internal/package-info.java
```diff
@@ -22,7 +22,7 @@
  * SOFTWARE.
  */
 /**
- * Utilities that are not considered part of the Public API.
+ * Internals.
```
kezz: Why the change here? Former was way clearer.

kezz (approved):

PR #653, opened 2021-12-22
feat(api): Add possibility of using Component#join with a root style

### LINKED ISSUE #505: Add a parent component/style to `JoinConfiguration`
It would be nice to be able to apply a root style/component to the join system so that, in the example below, I wouldn't have to declare the green colouring twice as it could just be the default.

```java
join(
  noSeparators(),
  List.of(
      text("You received ", NamedTextColor.GREEN),
      text(amount, NamedTextColor.GOLD),
      space(),
      prize,
      text(".", NamedTextColor.GREEN)
  )
)
```

KingOfSquares:
Closes #505

### DIFF
--- a/api/src/main/java/net/kyori/adventure/text/JoinConfigurationImpl.java
+++ b/api/src/main/java/net/kyori/adventure/text/JoinConfigurationImpl.java
@@ -134,7 +134,8 @@ final class JoinConfigurationImpl implements JoinConfiguration {
       ExaminableProperty.of("lastSeparator", this.lastSeparator),
       ExaminableProperty.of("lastSeparatorIfSerial", this.lastSeparatorIfSerial),
       ExaminableProperty.of("convertor", this.convertor),
-      ExaminableProperty.of("predicate", this.predicate)
+      ExaminableProperty.of("predicate", this.predicate),
+      ExaminableProperty.of("rootStyle", this.rootStyle)
     );
   }
 
--- a/api/src/main/java/net/kyori/adventure/text/JoinConfigurationImpl.java
+++ b/api/src/main/java/net/kyori/adventure/text/JoinConfigurationImpl.java
@@ -29,6 +29,7 @@ import java.util.function.Function;
 import java.util.function.Predicate;
 import java.util.stream.Stream;
 import net.kyori.adventure.internal.Internals;
+import net.kyori.adventure.text.format.Style;
 import net.kyori.examination.ExaminableProperty;
 import org.jetbrains.annotations.Contract;
 import org.jetbrains.annotations.NotNull;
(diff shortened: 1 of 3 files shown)


api/src/main/java/net/kyori/adventure/text/Component.java
```diff
@@ -218,7 +218,24 @@
...
+   * @param components the components
+   * @param rootStyle the style of the parent component containing all the joined components(e.g. if this is styled RED all joined components that does not have a color becomes RED)
+   * @return the resulting component
+   * @see JoinConfiguration#noSeparators()
+   * @see JoinConfiguration#separator(ComponentLike)
+   * @see JoinConfiguration#separators(ComponentLike, ComponentLike)
+   * @since 4.10.0
+   */
+  @Contract(pure = true)
+  static @NotNull Component join(final @NotNull JoinConfiguration config, final @NotNull Iterable<? extends ComponentLike> components, final @NotNull Style rootStyle) {
```
kezz, line 237: It would be better if this was just added to the `JoinConfiguration`

kezz (changes requested):

api/src/main/java/net/kyori/adventure/text/JoinConfigurationImpl.java
```diff
@@ -153,13 +164,21 @@ public String toString() {
     int componentsSeen = 0;
 
     if (!it.hasNext()) {
-      return singleElementJoin(config, component);
+      return singleElementJoin(
+        config,
+        hasRootStyle ?
```
kezz, line 169: I think we should handle this in the `singleElementJoin` method. Otherwise, the root style won't apply to the prefix/suffix/etc (which is the case for multi-element joining).

api/src/main/java/net/kyori/adventure/text/JoinConfigurationImpl.java
```diff
@@ -144,6 +153,8 @@ public String toString() {
     final Component suffix = config.suffix();
     final Function<ComponentLike, Component> convertor = config.convertor();
     final Predicate<ComponentLike> predicate = config.predicate();
+    final Style rootStyle = config.rootStyle();
+    final boolean hasRootStyle = rootStyle != Style.empty();
 
     if (!it.hasNext()) {
       return singleElementJoin(config, null);
```
kezz, line 160: Another reason for handling the style in the `singleElementJoin` method is so that this case can have the root style applied, which it isn't at the moment.

kezz (changes requested):
Lots of docs nitpicks here and a minor change for the single element join method. This is a very good change though!

kezz (changes requested):
Style needs to be added to the toString output

kezz (approved):
Needs since tags updating, but otherwise LGTM!

zml2008 (approved):

PR #657, opened 2021-12-29
Introduce a method by which to create raw replacements.

mbax:
Moreso meant to start conversation, because I bet there's a better stage at which to escape, using the specific MiniMessage object and not just the default one. 

Presently, only a minimessage-parsed string replacement is offered. Manually doing escapeTokens all the time gets tiring, so I hope to introduce a built-into-adventure method by which to do this. Otherwise, I'll just be creating my own simple helper method to automate.

To quote @kezz 
> maybe we can do a pass right at the end once all the tokens are made

image

### DIFF
--- a/text-minimessage/src/main/java/net/kyori/adventure/text/minimessage/placeholder/Placeholder.java
+++ b/text-minimessage/src/main/java/net/kyori/adventure/text/minimessage/placeholder/Placeholder.java
@@ -56,6 +56,18 @@ public interface Placeholder<T> extends Replacement<T> {
   }
 
   /**
+   * Creates a placeholder that inserts a raw string, ignoring any MiniMessage tags present.
+   *
+   * @param key the key
+   * @param value the replacement
+   * @return the placeholder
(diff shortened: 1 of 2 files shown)


kezz:
Random thoughts: if we hold a way of identifying a raw replacement then we can do the escaping as and when the placeholders are resolved, thus solving the problem of the escaping not happening on the currently active MM instance.

zml2008:
hm, why not return a `Placeholder<Component>`, and this method just delegates to `Placeholder.component(name, Component.text(raw))`?

kezz:
> hm, why not return a `Placeholder<Component>`, and this method just delegates to `Placeholder.component(name, Component.text(raw))`?

Yup, this is the solution we should be using for this! With a Javadoc that is something like `Creates a placeholder that inserts a raw string, ignoring any MiniMessage tags present in the string.`

zml2008 (approved):

kashike (approved):

rymiel (approved):

PR #658, opened 2021-12-31
text-minimessage: add deserialization method returning parsed tree

rymiel:
Related to KyoriPowered/adventure-webui#68.

Introduces a method returning the `ElementNode` tree, right before it would be turned into a component.

As this is mostly a debug feature with not much actual practical use, I'm wondering if it should somehow be more hidden in the API...

### DIFF
--- a/text-minimessage/src/main/java/net/kyori/adventure/text/minimessage/MiniMessage.java
+++ b/text-minimessage/src/main/java/net/kyori/adventure/text/minimessage/MiniMessage.java
@@ -159,6 +159,16 @@ public interface MiniMessage extends ComponentSerializer<Component, Component, S
   }
 
   /**
+   * Deserializes a string into a tree of parsed elements,
+   * This is intended for inspecting the output of the parser for debugging purposes.
+   *
+   * @param input the input string
+   * @return the root of the resulting tree
+   * @since 4.10.0
+   */
+  @NotNull Node deserializeToTree(@NotNull String input);
+
+  /**
    * Deserializes a string into a tree of parsed elements, with a tag resolver to parse tags of the form {@code <key>}.
    * This is intended for inspecting the output of the parser for debugging purposes.
    *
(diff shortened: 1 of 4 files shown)


text-minimessage/src/main/java/net/kyori/adventure/text/minimessage/MiniMessage.java
```diff
@@ -127,6 +128,18 @@ default Component parse(final @NotNull String input) {
...
+   * Deserializes a string into a tree of parsed elements, with a placeholder resolver to parse placeholders of the form {@code <key>}.
+   *
+   * <p>Placeholders will be resolved from this resolver before the resolver provided in the builder is used.</p>
+   *
+   * @param input the input string
+   * @param placeholderResolver the placeholder resolver
+   * @return the root of the resulting tree
+   * @since 4.10.0
+   */
+  @NotNull ElementNode deserializeToTree(final @NotNull String input, final @NotNull PlaceholderResolver placeholderResolver);
```
zml2008, line 141: do we want to make the `ElementNode` class as an implementation detail, and expose a new `MiniMessageNode` interface, that for now just has a `toString()` method?

That would keep things buttoned down for an initial release, and allow us to expand the API more thoughtfully at a future time. Other uses of `ElementNode`, like in `Transformation`s, could also be replaced with this `MiniMessageNode`.

rymiel: What is the node used for in Transformations?

zml2008: it's used for `Modifying` type transformations, in both cases to inspect the tree to get its length for gradient purposes.

rymiel: would you not consider Nodes as they are currently as API then? those `Modifying` transformations seem to be checking for very specific classes or is that fine because it's internal?

rymiel:
Rebased for #672

text-minimessage/src/main/java/net/kyori/adventure/text/minimessage/MiniMessage.java
```diff
@@ -157,6 +158,19 @@
...
+   * This is intended for inspecting the output of the parser for debugging purposes.
+   *
+   * <p>Tags will be resolved from the resolver parameter before the resolver provided in the builder is used.</p>
+   *
+   * @param input the input string
+   * @param tagResolver the tag resolver for any additional tags to handle
+   * @return the root of the resulting tree
+   * @since 4.10.0
+   */
+  @NotNull Node deserializeToTree(final @NotNull String input, final @NotNull TagResolver tagResolver);
```
zml2008, line 182: Can you add an overload that doesn't require a `TagResolver`?

text-minimessage/src/main/java/net/kyori/adventure/text/minimessage/MiniMessageParser.java
```diff
@@ -116,7 +117,7 @@ private void processTokens(final @NotNull StringBuilder sb, final @NotNull Strin
     }
   }
 
-  @NotNull Component parseFormat(final @NotNull String richMessage, final @NotNull ContextImpl context) {
+  @NotNull Node parseToTree(final @NotNull String richMessage, final @NotNull ContextImpl context) {
```
zml2008: This can just return an `ElementNode` -- it's not public API, and that removes the cast below.

zml2008 (changes requested):

text-minimessage/src/main/java/net/kyori/adventure/text/minimessage/MiniMessageParser.java
rymiel: Sorry, forgot how classes and interfaces work

zml2008 (approved):

Issue #515, opened 2021-11-02
Warn when input contains legacy formatting codes


kezz:
Perhaps even just throwing an exception would be nice

PR #662, opened 2022-01-11
Implemented #515.

LinoxGH:
Implemented the nagging feature from API to MiniMessage. But because API wasn't a part of MiniMessage I put down a small method and used that.

### DIFF
--- a/text-minimessage/src/main/java/net/kyori/adventure/text/minimessage/MiniMessageParser.java
+++ b/text-minimessage/src/main/java/net/kyori/adventure/text/minimessage/MiniMessageParser.java
@@ -90,7 +90,7 @@ final class MiniMessageParser {
 
   private void processTokens(final @NotNull StringBuilder sb, final @NotNull String richMessage, final @NotNull ContextImpl context, final BiConsumer<Token, StringBuilder> tagHandler) {
     final TagResolver combinedResolver = TagResolver.resolver(this.tagResolver, context.extraTags());
-    final List<Token> root = TokenParser.tokenize(richMessage, context.strict());
+    final List<Token> root = TokenParser.tokenize(richMessage);
     for (final Token token : root) {
       switch (token.type()) {
         case TEXT:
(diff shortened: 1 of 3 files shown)


text-minimessage/src/main/java/net/kyori/adventure/text/minimessage/MiniMessageParser.java
```diff
@@ -247,4 +250,10 @@ private Component handleModifying(final Modifying modTransformation, final Compo
   private String sanitizePlaceholderName(final String name) {
     return name.toLowerCase(Locale.ROOT);
   }
+
+  private void tryToNag(final String input, final boolean strict) {
```
zml2008, line 254: Maybe just call this in `TokenParser`? I believe that would capture all code paths at once.

zml2008 (changes requested):

text-minimessage/src/main/java/net/kyori/adventure/text/minimessage/MiniMessageParser.java
kezz: Yeah, perhaps doing it as late as possible is the better solution?

text-minimessage/src/main/java/net/kyori/adventure/text/minimessage/parser/TokenParser.java
```diff
@@ -503,6 +503,8 @@ private static ElementNode buildTree(
       throw new ParsingException(sb.toString(), message, errorTokens);
     }
 
+    tryToNag(message, strict);
```
zml2008, line 506: Why are you running this so late? Plus, buildTree isn't used for the strip and escape tokens methods.

zml2008 (changes requested):

zml2008 (approved):

zml2008:
I've rebased this and tweaked the implementation to be fully within the parser -- would appreciate another set of eyes over it though.

Issue #516, opened 2021-11-28
More color aliases


rymiel:
I know this has been brought up a few times before in discussion but I couldn't see a formal issue tracking this.

The idea is to support a wider range of names for colors, as a "color alias" system already exists in the code. Perhaps it could even support [the whole set of CSS color](https://www.w3.org/TR/css-color-3/#svg-color).

PR #664, opened 2022-01-12
CSS Colors and Color Alias Methods

Majekdor:
Related to #516, this PR adds a new Transformation Type, `TransformationType.CSS_COLOR`, that allows for all standard CSS colors to be parsed with MiniMessage. If both `TransformationType.CSS_COLOR` and `TransformationType.COLOR` are included and a color is specified that is included in both (ex. aqua), then the standard named text color is used. You can specify to use the CSS color like so: `<css:aqua>`.

This PR also adds two additions methods to TransformationType that allow the user to create their own ColorTransformation from a map of strings to text colors.

### DIFF
--- a/text-minimessage/src/main/java/net/kyori/adventure/text/minimessage/transformation/TransformationType.java
+++ b/text-minimessage/src/main/java/net/kyori/adventure/text/minimessage/transformation/TransformationType.java
@@ -32,6 +32,7 @@ import java.util.function.Predicate;
 import java.util.stream.Collectors;
 import java.util.stream.Stream;
 import net.kyori.adventure.text.format.TextDecoration;
+import net.kyori.adventure.text.minimessage.transformation.inbuild.CSSColorTransformation;
 import net.kyori.adventure.text.minimessage.transformation.inbuild.ClickTransformation;
 import net.kyori.adventure.text.minimessage.transformation.inbuild.ColorTransformation;
 import net.kyori.adventure.text.minimessage.transformation.inbuild.DecorationTransformation;
@@ -56,6 +57,10 @@ public final class TransformationType<T extends Transformation> {
   public static final TransformationType<?> COLOR = transformationType(
     ColorTransformation::canParse,
     ColorTransformation::create
+  );
+  public static final TransformationType<?> CSS_COLOR = transformationType(
+    CSSColorTransformation::canParse,
+    CSSColorTransformation::create
   );
   public static final TransformationType<?> DECORATION = transformationType(
     acceptingNames(
--- a/text-minimessage/src/main/java/net/kyori/adventure/text/minimessage/transformation/TransformationType.java
+++ b/text-minimessage/src/main/java/net/kyori/adventure/text/minimessage/transformation/TransformationType.java
@@ -27,11 +27,14 @@ import java.util.Arrays;
 import java.util.Collection;
 import java.util.HashSet;
 import java.util.Locale;
+import java.util.Map;
 import java.util.Set;
 import java.util.function.Predicate;
 import java.util.stream.Collectors;
 import java.util.stream.Stream;
+import net.kyori.adventure.text.format.TextColor;
 import net.kyori.adventure.text.format.TextDecoration;
+import net.kyori.adventure.text.minimessage.parser.ParsingException;
 import net.kyori.adventure.text.minimessage.transformation.inbuild.CSSColorTransformation;
 import net.kyori.adventure.text.minimessage.transformation.inbuild.ClickTransformation;
 import net.kyori.adventure.text.minimessage.transformation.inbuild.ColorTransformation;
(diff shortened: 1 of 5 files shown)


text-minimessage/src/main/java/net/kyori/adventure/text/minimessage/transformation/inbuild/CSSColorTransformation.java
```diff
@@ -0,0 +1,263 @@
...
+ * A transformation applying a single css color.
+ *
+ * @since 4.10.0
+ */
+public final class CSSColorTransformation extends Transformation {
+
+  public static final Map<String, TextColor> CSS_COLORS = new HashMap<>();
+
+  static {
+    CSS_COLORS.put("aliceblue", TextColor.color(0xf0f8ff));
```
MiniDigger, line 50: a bit meh that java has no constants for this, only javafx has Color.web and stuff :/

text-minimessage/src/main/java/net/kyori/adventure/text/minimessage/transformation/TransformationType.java
```diff
@@ -182,4 +190,58 @@ private TransformationType(final Predicate<String> canParse, final Transformatio
...
   }
+
+  /**
+   * Create a new color transformation type for parsing the supplied color map.
+   *
+   * @param aliases map of color names to their {@link TextColor} values
+   * @return a new color transformation
+   * @since 4.10.0
+   */
+  public static TransformationType<ColorTransformation> color(final Map<String, TextColor> aliases) {
```
MiniDigger, line 201: I think these  methods should be in color transformation and the ctor there private

text-minimessage/src/main/java/net/kyori/adventure/text/minimessage/transformation/inbuild/CSSColorTransformation.java
Majekdor: I looked everywhere thinking they'd surely be somewhere. Funny story, I used javafx's Color and System.out.println to make those 140 lines xP

kezz:
Blocking on #672, will need to be rewritten to use the new tag system after merge.

Majekdor:
Added the new CSSColorTagResolver in a new package I called "nonstandard". Happy to change that package name or move it somewhere else but I saw mention of the possibility of tags being added that wouldn't be "standard" so I stuck it there.

text-minimessage/src/main/java/net/kyori/adventure/text/minimessage/tag/nonstandard/CSSColorTagResolver.java
```diff
@@ -0,0 +1,222 @@
...
+ * A transformation applying a single css text color.
+ *
+ * @since 4.10.0
+ */
+@ApiStatus.Internal
+public class CSSColorTagResolver implements TagResolver {
+  public static final Map<String, TextColor> CSS_COLORS = new HashMap<>();
+
+  static {
+    CSS_COLORS.put("aliceblue", TextColor.color(0xf0f8ff));
```
zml2008, line 125: If these are built from constants in `java.awt.Color`, can we auto-generate those at build time (even if it's into a `properties` file that gets read), rather than having a manually maintained list?

text-minimessage/src/main/java/net/kyori/adventure/text/minimessage/tag/nonstandard/CSSColorTagResolver.java
```diff
@@ -0,0 +1,222 @@
...
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+ * SOFTWARE.
+ */
+package net.kyori.adventure.text.minimessage.tag.nonstandard;
```
zml2008: I'd rather have this in `tag.standard`, exposed in `StandardTags`, but not included in `StandardTags`

text-minimessage/src/main/java/net/kyori/adventure/text/minimessage/tag/nonstandard/CSSColorTagResolver.java
```diff
@@ -0,0 +1,222 @@
...
+import org.jetbrains.annotations.NotNull;
+import org.jetbrains.annotations.Nullable;
+
+/**
+ * A transformation applying a single css text color.
+ *
+ * @since 4.10.0
+ */
+@ApiStatus.Internal
+public class CSSColorTagResolver implements TagResolver {
```
zml2008: I'd kinda like to expose API for users to build their own color tags with any set of aliases -- maybe `StandardTags.color[Like?](Map<String, TextColor> aliases)`, or `StandardTags.color[Like?](String qualifier, Map<String, TextColor> aliases)`? That would get rid of a separate tag implementation class for just the CSS colours.

The tag implementation classes shouldn't be part of the API either way, there should be a factory method for them if this class has to stick around.

zml2008 (changes requested):

zml2008:
This tag will also want a companion pr to https://github.com/KyoriPowered/adventure-docs.

text-minimessage/src/main/java/net/kyori/adventure/text/minimessage/tag/nonstandard/CSSColorTagResolver.java
Majekdor: They were built from `javafx.scene.paint.Color` in a very hacky manner. From what I've found there's no reasonable way to use another list.

text-minimessage/src/main/java/net/kyori/adventure/text/minimessage/tag/nonstandard/CSSColorTagResolver.java
Majekdor: I like the idea of these methods and have them implemented. My one question is regarding the second method with a qualifier. Using the css colors as an example, if a user adds that resolver to the MiniMessage instance and uses the `<aqua>` tag, the aqua value from css instead of Minecraft will be used. 

Should the method with the qualifier prioritize Minecraft color codes if the qualifier isn't present? If it doesn't I'm not entirely sure why we'd need a qualifier for colors. 

It can be made to work properly if the user adds the standard color resolver after the css color resolver, but that seems unintuitive since the standard color resolver is already included.

Majekdor: Hmmm nevermind on this. I think this way may be the best. Even if css colors are added after then Minecraft colors can still be accessed with `<c:name>` and if no qualifier is given it will default to last resolver added.

text-minimessage/src/main/java/net/kyori/adventure/text/minimessage/tag/nonstandard/CSSColorTagResolver.java
kezz: Hmm, I'm not sure about this - I feel like that package should be kept for tags included in the default parser - I'd much prefer an `extra` package alongside `ExtraTags` perhaps?

Majekdor:
The map of css colors should probably move. I'm just not sure where to.

Issue #656, opened 2021-12-29
Placeholder resolver builder


kezz:
In addition to the static constructor pattern, it would be nice to provide a robust builder

PR #665, opened 2022-01-13
text-minimessage: Add a builder for PlaceholderResolvers

zml2008:
Closes GH-656

Should we keep the static factory methods after this builder is added? do some of them make sense to get rid of?

I've noticed some of the static factory methods use live views of provided collections in the returned resolvers. I think this is generally the wrong choice, since it could result in live state being modified in unexpected ways, and it is trivial to implement a live `PlaceholderResolver` if that behavior truly desired. Does anyone have a defense of that live collection behavior?


kezz (approved):

gepron1x:
Cool! Would you mind adding convenience placeholder methods that takes primitives and such? E.g builder.placeholder(key, int), builder.placeholder(key, component), placeholder(key, string)

zml2008:
@gepron1x that can definitely be considered after the next stage of refactoring -- I had them in initial work on the builder, but chose to remove them to match the factory methods for PlaceholderResolvers.

PR #669, opened 2022-01-22
Title Builder

### LINKED ISSUE #668: Add replaceText to Title
as the title, it is suggested to add `replaceText` to the title
```
  @Contract(pure = true)
  @NotNull Component replaceText(final @NotNull TextReplacementConfig config);
```
allow it to replace text in both title and subtitle

thiccaxe:
Closes #668 by adding a more versatile builder for both Title and Titles.Times.


thiccaxe:
Would having builders for other immutable objects be of use? No need to make a special case for Title...

zml2008:
What sort of objects were you thinking about? It's probably not worth the implementation overhead for many of the smaller or lesser-used interfaces in the API, just those with specific use cases or with more deeply nested object graphs

api/src/main/java/net/kyori/adventure/title/Title.java
```diff
@@ -164,5 +199,109 @@
...
+    interface Builder extends Buildable.Builder<Title.Times> {
+      /**
+       * Set the time the title will fade-in.
+       *
+       * @param fadeIn the fade-in time
+       * @return this
+       * @since 4.10.0
+       */
+      @Contract("_ -> this")
+      @NotNull Builder fadeIn(final @NotNull Duration fadeIn);
```
zml2008, line 231: Maybe add in `fadeInSeconds` overloads here?

@kashike thoughts on adding tick-based convenience overloads here? or do we not want to depend on a fixed tick rate?

api/src/main/java/net/kyori/adventure/title/Title.java
```diff
@@ -164,5 +199,109 @@
...
+
+    /**
+     * Set the times.
+     *
+     * @param times the times
+     * @return this
+     * @since 4.10.0
+     */
+    @Contract("_ -> this")
+    @NotNull Builder times(final @Nullable Times times);
```
zml2008, line 297: How about a `times(Duration, Duration, Duration)` overload here?

api/src/main/java/net/kyori/adventure/title/Title.java
```diff
@@ -164,5 +199,109 @@
...
+
+    /**
+     * Builds.
+     *
+     * @return a new title
```
zml2008, line 302:
```suggestion
     * @return a new title
     * @since 4.10.0
```

api/src/main/java/net/kyori/adventure/title/Title.java
```diff
@@ -164,5 +199,109 @@
...
+       * @return this
+       * @since 4.10.0
+       */
+      @Contract("_ -> this")
+      @NotNull Builder fadeOut(final @NotNull Duration fadeOut);
+
+      /**
+       * Builds.
+       *
+       * @return a new times
```
zml2008, line 256:
```suggestion
       * @return a new times
       * @since 4.10.0
```

api/src/main/java/net/kyori/adventure/title/Title.java
```diff
@@ -106,12 +118,25 @@
    */
   <T> @UnknownNullability T part(final @NotNull TitlePart<T> part);
 
+  /**
+   * Create a new builder initialized with the attributes of this title.
+   *
+   * @return the builder
```
zml2008, line 124:
```suggestion
   * @return the builder
   * @since 4.10.0
```

zml2008 (changes requested):

thiccaxe:
> What sort of objects were you thinking about? It's probably not worth the implementation overhead for many of the smaller or lesser-used interfaces in the API, just those with specific use cases or with more deeply nested object graphs

I was thinking specifically of the Sound class, but then again I suppose most of your sounds are probably constant.

api/src/main/java/net/kyori/adventure/title/Title.java
kashike: Would prefer no tick-based methods.

thiccaxe: > fadeInSeconds
Sounds like something the Duration class handles fairly comprehensively to me 🤷‍♂️

kezz:
Tbh, I'm not sure I like the times builder, or at least, I'm not sure I like it being separate from the main builder so to speak. Kinda feels unnecessary to have a builder for three durations?

thiccaxe:
`Builder fadeIn(Duration, Duration, Duration)` ?

zml2008:
> I was thinking specifically of the Sound class, but then again I suppose most of your sounds are probably constant.

I think a `Sound` builder could be worthwhile, the `Sound` interface has quite a few parameters.

> `Builder fadeIn(Duration, Duration, Duration)` ?

You already have a `times()` method on the Title builder, that should be fine? just get rid of the `Times` builder really.

kezz:
Going to close this for now as it appears to be inactive. If you want this reopened in the future please let us know!

PR #671, opened 2022-02-04
text-minimessage: Use a Consumer rather than Appendable for debug output

### LINKED ISSUE #667: Remove `Appendable` usage
Should just be replaced with a nullable (or even non-nullable with a default dummy impl) `Consumer<String>`.

zml2008:
Fixes GH-667

### DIFF
--- a/text-minimessage/src/main/java/net/kyori/adventure/text/minimessage/MiniMessage.java
+++ b/text-minimessage/src/main/java/net/kyori/adventure/text/minimessage/MiniMessage.java
@@ -185,7 +185,7 @@ public interface MiniMessage extends ComponentSerializer<Component, Component, S
      * @return this builder
      * @since 4.10.0
      */
-    @NotNull Builder debug(final @Nullable Appendable debugOutput);
+    @NotNull Builder debug(final @Nullable Consumer<String> debugOutput);
 
     /**
      * If in lenient mode, MiniMessage will output helpful messages.
--- a/text-minimessage/src/main/java/net/kyori/adventure/text/minimessage/MiniMessage.java
+++ b/text-minimessage/src/main/java/net/kyori/adventure/text/minimessage/MiniMessage.java
@@ -181,6 +181,9 @@ public interface MiniMessage extends ComponentSerializer<Component, Component, S
      *
      * <p>Debug output includes detailed information about the parsing process to help debug parser behavior.</p>
      *
+     * <p>The consumer will receive line fragments terminated by {@code LF}, not complete lines.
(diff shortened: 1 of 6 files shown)


text-minimessage/src/main/java/net/kyori/adventure/text/minimessage/MiniMessageParser.java
```diff
@@ -120,45 +120,47 @@ private void processTokens(final @NotNull StringBuilder sb, final @NotNull Strin
...
-    final Appendable debug = context.debugOutput();
+    final Consumer<String> debug = context.debugOutput();
     if (debug != null) {
-      try {
-        debug.append("Beginning parsing message ").append(richMessage).append('\n');
-      } catch (final IOException ignored) {
-      }
+      debug.accept("Beginning parsing message ");
+      debug.accept(richMessage);
+      debug.accept("\n");
```
kashike, line 127: Should we combine this into one `accept` call instead?

zml2008: ideally we avoid string concat here, since that would produce an unnecessary StringBuilder

kezz (approved):

text-minimessage/src/main/java/net/kyori/adventure/text/minimessage/MiniMessage.java
```diff
@@ -185,7 +185,7 @@ static Builder builder() {
      * @return this builder
      * @since 4.10.0
      */
-    @NotNull Builder debug(final @Nullable Appendable debugOutput);
+    @NotNull Builder debug(final @Nullable Consumer<String> debugOutput);
```
kashike, line 191: Documentation should mention how this is used - I suspect some people (like myself at first) will assume this is an *entire* message per-string.

kashike (changes requested):

gepron1x:
May i ask, why do you use null if debug is disabled? Why not just empty consumer?

gepron1x:
Also, consider adding convenience method to builder that accepts Appendable and sneaky throws IOException?

kezz:
> May i ask, why do you use null if debug is disabled? Why not just empty consumer?

There are some non-zero cost calculations/string formatting that takes place that doesn't need to happen if nobody is caring about the logging.

> Also, consider adding convenience method to builder that accepts Appendable and sneaky throws IOException?

I'm not massively interested in adding a convenience method for `Appendable`, it's a bit of a rubbish class to use and considering how easy it is to wrap in a consumer I don't think we really need to add/encourage use of such a method.

kashike (approved):

Issue #523, opened 2021-10-23
Add a way to prevent templates impacting remaining content


kezz:
Something like `<my_template/>` which would resolve the template whilst preventing any tags inserted from bleeding into remaining strings.

Essentially shorthand for `<span><my_template></span>`, considering KyoriPowered/adventure-text-minimessage#178.

zml2008:
This should be easy with #672 -- it just depends on the value of `Tag#allowsChildren()`. We could incorporate the same logic into the parser in order to support `/` for self-closing tags.

PR #672, opened 2022-02-07
minimessage: Unify transformations and placeholders

### LINKED ISSUE #522: Convert templates to “compound” tags
Building on the ideas of KyoriPowered/adventure-text-minimessage#178 & KyoriPowered/adventure-text-minimessage#179 , this will greatly enhance the flexibility of templates.

new syntax for templates:
`<template>text with "bleeded" styles of template</template>regular text`

In essence, a template is no longer a string that is replaced, but a full tag.
Unlike a simple `<red>`, however, a template can contain its own styles, text, click events, translations, or even templates inside it.

i.e.: if `<template>` was defined as `<red><bold>Prefix</bold>`,
The first example would be equivalent to writing out:

`<red><bold>Prefix</bold>text with "bleeded" styles of template</red> regular text`

As previously stated, this greatly expands the flexibility of templates and improves minimesaage overall

zml2008:
This replaces both with a new Tag system, which exposes both
argumentless insertions and argument-handling transformations in one
registry, with one type hierarchy.

I hope to open the door to piggybacking on these tag resolvers to handle serialization as well at some future point.

Still to do:

- [x] examinability of built-in `Tag` implementations, to ease debugging
- [x] correct the message passed to exceptions created in `Context`
- [ ] names?
- [x] API-ification of ElementNode (related to #658)
- ~~[ ] common impl for rainbow and gradient tags~~ happening later, not an API-visible change
- ~~[ ] for rainbow and gradient length calculations, a way to distinguish between pre-calculated and lazily computed `Inserting` types (does this make sense to be restricted to just placeholders even?)~~ delayed
- [x] `ArgumentQueue`/other helpers for parsing tag arguments (something providing API along the lines `args.pop()` that auto-throws an exception if the required argument isn't present, other helpers like `Tag.Argument#asDouble(): OptionalDouble`)

# Overview

The changes in this PR aim to unify two very similar APIs to reduce the number of concepts developers have to deal with. In doing so, I've managed to reduce some layers of abstraction, and provide the full capabilities of each system for every use case. This has resulted in some shuffling around:

Placeholders | Transformations | New `Tag`s
------------ | ----------------- | --------------
`PlaceholderResolver` | `TransformationRegistry` | `TagResolver`
`Placeholder` | `TransformationFactory` | `TagResolver`
`Replacement` | `Transformation` | `Tag`

As you can see, the concepts of a factory/placeholder and resolver/registry have been merged -- it is possible to consider a `Placeholder` or `Factory` as a simple case of a resolver, which only contains one element.

# Types of `Tag`s

## `Inserting`

These tags are fairly straightforward: they represent a literal `Component`. The vast majority of `Tag` implementations will want to be `Inserting` tags. Inserting tags may also optionally be self-closing -- by default, this is only true for tags created by resolvers from the `Placeholder` class, so that placeholders are self-contained.

## `Modifying`

`Modiying` tags have the ability to perform transformations on the entire component tree of their children after it has been created. This is used for the `<rainbow>` and `<gradient>` tags, but can be applied to any other more complex transformations.

## `PreProcess` (or `minimessage`)

These tags have a value of a MiniMessage string. They are unique in that they're applied before any sort of parsing begins, so can be fragments rather than entire strings.

# Examples

The implementations of the built-in tags for representing components provide a starting point for examples of how this system works. However, to address some more simple cases, and provide examples of a custom tag with richer functionality:

**A simple case with placeholder-equivalent tags:**

```java
    final String input = "[<prefix><player><suffix>] >> <message>";
    final TagResolver placeholders = TagResolver.combining(
      Placeholder.parsed("prefix", "<red>Admin </red><blue>"),
      Placeholder.component("player", text("kashike")),
      Placeholder.parsed("suffix", "</blue>"),
      Placeholder.component("message", text("Hello!"))
    );
    
    final Component parsed = MiniMessage.miniMessage().deserialize(input, placeholders);
```

**A fancier case, with custom logic**

This creates a custom tag that performs all the styling needed to create a standard hyperlink using an HTML-like `a` tag:

```java
  Component aTagExample() {
    final String input = "Hello, <a:https://kyori.net>click me!</a> but not me!";
    final MiniMessage extendedInstance = MiniMessage.builder()
      .tags(b -> b.resolver(TagResolver.resolver("a", MiniMessageTest::createA)))
      .build();

    return extendedInstance.deserialize(input);
  }

  static Tag createA(final ArgumentQueue args, final Context ctx) {
    final String link = args.popOr("The <a> tag requires exactly one argument, the link to open").value();

    return Tag.styling(
      NamedTextColor.BLUE,
      TextDecoration.UNDERLINED,
      ClickEvent.openUrl(link),
      HoverEvent.showText(Component.text("Open " + link))
    );
  }
```


text-minimessage/src/main/java/net/kyori/adventure/text/minimessage/MiniMessageImpl.java
```diff
@@ -66,8 +65,8 @@
   }
 
   @Override
-  public @NotNull Component deserialize(final @NotNull String input, final @NotNull PlaceholderResolver placeholderResolver) {
-    return this.parser.parseFormat(input, this.newContext(input, requireNonNull(placeholderResolver, "placeholderResolver")));
+  public @NotNull Component deserialize(final @NotNull String input, final @NotNull TagResolver tagResolver) {
+    return this.parser.parseFormat(input, this.newContext(input, requireNonNull(tagResolver, "placeholderResolver")));
```
rymiel: lil missed rename?

text-minimessage/src/main/java/net/kyori/adventure/text/minimessage/MiniMessageParser.java
```diff
@@ -189,39 +196,46 @@ private void processTokens(final @NotNull StringBuilder sb, final @NotNull Strin
...
+        final Modifying modTransformation = (Modifying) tag;
 
         // first walk the tree
-        final LinkedList<ElementNode> toVisit = new LinkedList<>(node.children());
+        final LinkedList<ElementNode> toVisit = new LinkedList<>(node.unsafeChildren());
         while (!toVisit.isEmpty()) {
           final ElementNode curr = toVisit.removeFirst();
           modTransformation.visit(curr);
-          toVisit.addAll(0, curr.children());
+          toVisit.addAll(0, curr.unsafeChildren());
```
rymiel, line 217: Why do these need to use **unsafe** children? I tested with regular children and couldn't see any difference?

text-minimessage/src/main/java/net/kyori/adventure/text/minimessage/MiniMessageParser.java
```diff
@@ -119,66 +117,75 @@ private void processTokens(final @NotNull StringBuilder sb, final @NotNull Strin
...
-      return this.registry.exists(sanitized) || (includePlaceholders && combinedResolver.resolve(name) != null);
+    final Predicate<String> tagNameChecker = name -> {
+      final String sanitized = TokenParser.TagProvider.sanitizePlaceholderName(name);
+      return combinedResolver.has(sanitized);
     };
 
-    final ElementNode root = TokenParser.parse(transformationFactory, tagNameChecker, combinedResolver, richMessage, context.strict());
+    final String preProcessed = TokenParser.resolvePreProcessTags(richMessage, transformationFactory);
+    context.message(preProcessed);
+    // Then, once MiniMessage placeholders have
```
rymiel: Seems like an incomplete comment...?

rymiel (commented):
Couldn't review because github was broken but i noticed some stuff while testing it locally

text-minimessage/src/main/java/net/kyori/adventure/text/minimessage/MiniMessageParser.java
zml2008: no point in constructing an unmodifiable wrapper collection for every call while processing the transformation, when this is internal use we can control

zml2008:
I think I've covered all the essential API elements in this PR now. I'd love to get this in by the end of the upcoming weekend, so please get your feedback in.

text-minimessage/src/main/java/net/kyori/adventure/text/minimessage/Context.java
```diff
@@ -37,19 +39,48 @@
 @ApiStatus.NonExtendable
 public interface Context {
   /**
-   * Returns original message as provided to the parser.
+   * Parses a MiniMessage using all the settings of this context, including placeholders.
```
kezz: Probably don't need to specify placeholders here anymore right?

```suggestion
   * Parses a MiniMessage string using all the settings of this context.
```

text-minimessage/src/main/java/net/kyori/adventure/text/minimessage/ContextImpl.java
```diff
@@ -95,6 +105,30 @@ public boolean strict() {
 
   @Override
   public @NotNull Component parse(final @NotNull String message) {
-    return this.miniMessage.deserialize(requireNonNull(message, "message"), this.placeholderResolver);
+    return this.miniMessage.deserialize(requireNonNull(message, "message"), this.tagResolver);
+  }
+
+  @Override
+  public @NotNull ParsingException newError(@NotNull final String message) {
+    return new ParsingExceptionImpl(message, this.message, null, new Token[0]);
```
kezz: Nitpick, but we could cache this zero length array?

text-minimessage/src/main/java/net/kyori/adventure/text/minimessage/MiniMessage.java
```diff
@@ -73,18 +72,18 @@
    * <p>Useful for untrusted input.</p>
    *
    * @param input the input message, with tokens
-   * @param placeholders the placeholder resolver adding known placeholders
+   * @param tags a tag resolver to provide any additional tags that can be understood, combined with the overall tag resolver
    * @return the output, with escaped tokens
    * @since 4.10.0
    */
-  @NotNull String escapeTokens(final @NotNull String input, final @NotNull PlaceholderResolver placeholders);
+  @NotNull String escapeTokens(final @NotNull String input, final @NotNull TagResolver tags);
```
kezz: So, what if these methods took varargs tag resolvers, thus kinda restoring the more simple use case of `miniMessage.deserialize("<red><prefix><name><suffix>", prefixPlaceholder, namePlaceholder, suffixPlaceholder)`, a la 4.1.0 and kin?

text-minimessage/src/main/java/net/kyori/adventure/text/minimessage/MiniMessage.java
```diff
@@ -98,23 +97,23 @@
...
   /**
-   * Deserializes a string into a component, with a placeholder resolver to parse placeholders of the form {@code <key>}.
+   * Deserializes a string into a component, with a tag resolver to parse tags of the form {@code <key>}.
    *
-   * <p>Placeholders will be resolved from this resolver before the resolver provided in the builder is used.</p>
+   * <p>Tags will be resolved from the resolver parameter before the resolver provided in the builder is used.</p>
    *
    * @param input the input string
-   * @param placeholderResolver the placeholder resolver
+   * @param tagResolver the tag resolver for any additional tags to handle.
```
kezz: This is much nicer than the docs for the same element of the `stripTokens` and `escapeTokens` methods.

```suggestion
   * @param tagResolver the tag resolver for any additional tags to handle
```

text-minimessage/src/main/java/net/kyori/adventure/text/minimessage/MiniMessage.java
```diff
@@ -134,41 +133,28 @@ static Builder builder() {
   interface Builder extends Buildable.Builder<MiniMessage> {
 
     /**
-     * Uses the supplied transformation registry.
+     * Set the known tags to the provided tag resolver.
```
kezz, line 178: Perhaps we mention that this is the "base" tag resolver somehow?

text-minimessage/src/main/java/net/kyori/adventure/text/minimessage/Context.java
```diff
@@ -37,19 +39,48 @@
...
-  @NotNull Component parse(final @NotNull String message);
+  @NotNull ParsingException newError(
+    final @NotNull String message,
+    final @NotNull ArgumentQueue tags
+  );
+
+  /**
+   * Create a new parsing exception without reference to a specific location.
+   *
+   * @param message the detail message
```
MiniDigger: its not clear if this is the error message or the input minimessage, that should be clarified in the javadocs (for all error methods here)

text-minimessage/src/main/java/net/kyori/adventure/text/minimessage/MiniMessageParser.java
```diff
@@ -105,8 +103,8 @@ private void processTokens(final @NotNull StringBuilder sb, final @NotNull Strin
             sb.append(richMessage, token.startIndex(), token.endIndex());
             continue;
           }
-          final String sanitized = this.sanitizePlaceholderName(token.childTokens().get(0).get(richMessage).toString());
-          if (this.registry.exists(sanitized, combinedResolver) || combinedResolver.resolve(sanitized) != null) {
+          final String sanitized = TokenParser.TagProvider.sanitizePlaceholderName(token.childTokens().get(0).get(richMessage).toString());
```
MiniDigger, line 106: sanitizePlaceholderName here should be renamed too I guess?

text-minimessage/src/main/java/net/kyori/adventure/text/minimessage/ParsingException.java
```diff
@@ -0,0 +1,62 @@
...
+public abstract class ParsingException extends RuntimeException {
+  private static final long serialVersionUID = 4502774670340827070L;
+
+  /**
+   * Create a new parsing exception with only a message.
+   *
+   * @param message the message
+   * @since 4.10.0
+   */
+  protected ParsingException(final String message) {
```
MiniDigger, line 40: here too, is message the origionalText or an error message?

text-minimessage/src/main/java/net/kyori/adventure/text/minimessage/tag/builtin/BuiltInTags.java
```diff
@@ -0,0 +1,186 @@
...
+ *
+ * @since 4.10.0
+ */
+public final class BuiltInTags {
+  private BuiltInTags() {
+  }
+
+  private static final TagResolver DECORATION = Stream.concat(TextDecoration.NAMES.keyToValue().entrySet().stream(), DecorationTag.DECORATION_ALIASES.entrySet().stream())
+    .flatMap(entry -> {
+      return Stream.of(
```
MiniDigger: not 100% sure if I like having tags declared inline, having every single one as a class gave a nice overview and would keep this class as a dum registry

text-minimessage/src/main/java/net/kyori/adventure/text/minimessage/tag/builtin/BuiltInTags.java
```diff
@@ -0,0 +1,186 @@
...
+  private static final TagResolver GRADIENT = TagResolver.resolver(GradientTag.GRADIENT, GradientTag::create);
+  private static final TagResolver RAINBOW = TagResolver.resolver(RainbowTag.RAINBOW, RainbowTag::create);
+
+  /**
+   * Get a resolver for the {@value HoverTag#HOVER} tag.
+   *
+   * @return a resolver for the {@value HoverTag#HOVER} tag
+   * @since 4.10.0
+   */
+  public static TagResolver hoverEvent() {
```
MiniDigger: is there a reason to use methods over just sticking with the constants?

text-minimessage/src/test/java/net/kyori/adventure/text/minimessage/MiniMessageParserTest.java
```diff
@@ -222,79 +220,79 @@ void testAllColorAliases() {
...
     final String expected = "\\<yellow><test> random \\<bold>stranger\\</bold>\\<click:run_command:test command>\\<underlined>\\<red>click here <notToken>\\</click>\\<blue> to \\<bold>FEEL\\</underlined> it";
-    assertEquals(expected, this.PARSER.escapeTokens(input));
+    assertEquals(expected, PARSER.escapeTokens(input));
   }
 
   @Test
   void testEscapePlaceholders() {
     final String input = "Hello, <red><name>!";
     final String expected = "Hello, \\<red>\\<name>!";
-    assertEquals(expected, this.PARSER.escapeTokens(input, PlaceholderResolver.placeholders(Placeholder.miniMessage("name", "you"))));
```
MiniDigger, line 290: just noticed this here, its not 100% clear, especially when you statically import it, what the methods in Placeholder do.
miniMessagePlaceholder would be kinda long tho and redundant when you do Placeholder.miniMessage, so idk

text-minimessage/src/main/java/net/kyori/adventure/text/minimessage/MiniMessage.java
```diff
@@ -134,41 +133,28 @@ static Builder builder() {
...
+     * @param tags the tag resolver to use
      * @return this builder
      * @since 4.10.0
      */
-    @NotNull Builder transformations(final @NotNull TransformationRegistry transformationRegistry);
+    @NotNull Builder tags(final @NotNull TagResolver tags);
 
     /**
-     * Modify the set transformation registry.
+     * Add to the set of known tags this MiniMessage instance can use.
```
kezz, line 190: Nitpick again, but it's slightly confusing to me that one `tags` method sets the tags but the other one adds too the tags. I know it's pretty clear from reading the javadocs, but perhaps this could be named `editTags`?

text-minimessage/src/main/java/net/kyori/adventure/text/minimessage/MiniMessageParser.java
```diff
@@ -119,66 +117,75 @@ private void processTokens(final @NotNull StringBuilder sb, final @NotNull Strin
...
+            debug.accept(name);
             debug.accept("'\n");
           } else {
             debug.accept("Successfully matched node '");
-            debug.accept(node.name());
-            debug.accept("' to transformation ");
-            debug.accept(transformation.examinableName());
+            debug.accept(name);
+            debug.accept("' to tag ");
+            debug.accept(transformation instanceof Examinable ? ((Examinable) transformation).examinableName() : transformation.getClass().getName());
```
kezz, line 151: Hmm - could we not require transformations to be examinable perhaps? Not a big issue..

text-minimessage/src/main/java/net/kyori/adventure/text/minimessage/MiniMessageParser.java
```diff
@@ -119,66 +117,75 @@ private void processTokens(final @NotNull StringBuilder sb, final @NotNull Strin
...
             debug.accept("\n");
           }
 
           return transformation;
         } catch (final ParsingException e) {
-          if (e.tokens().length == 0) {
-            e.tokens(new Token[]{node.token()});
+          if (token != null && e instanceof ParsingExceptionImpl) {
```
kezz, line 157: The null check isn't necessary here right?

text-minimessage/src/main/java/net/kyori/adventure/text/minimessage/ParsingException.java
```diff
@@ -0,0 +1,62 @@
...
+ * SOFTWARE.
+ */
+package net.kyori.adventure.text.minimessage;
+
+/**
+ * An exception thrown when an error occurs while parsing a MiniMessage string.
+ *
+ * @since 4.10.0
+ */
+public abstract class ParsingException extends RuntimeException {
```
kezz, line 31: Are end-users allowed to implement this class?

text-minimessage/src/main/java/net/kyori/adventure/text/minimessage/parser/TokenParser.java
```diff
@@ -368,41 +363,28 @@ private static ElementNode buildTree(
...
               } else {
                 // This is a recognized tag, goes in the tree
-                tagNode.transformation(transformation);
+                tagNode.tag(transformation);
                 node.addChild(tagNode);
-                if (!(transformation instanceof Inserting)) {
-                  // this tag has children
-                  node = tagNode;
+                if (!(transformation instanceof Inserting) || ((Inserting) transformation).allowsChildren()) {
+                  node = tagNode; // TODO: self-terminating tags (i.e. <tag/>) don't set this, so they don't have children
```
kezz, line 387: Just noting this TODO in case we forget.

text-minimessage/src/main/java/net/kyori/adventure/text/minimessage/parser/TokenParser.java
```diff
@@ -672,4 +610,64 @@ public static String unescape(final String text, final int startIndex, final int
...
     return sb.toString();
   }
+
+  /**
+   * Normalizing provider for tag information.
+   *
+   * @since 4.10.0
+   */
+  @ApiStatus.Internal
+  public interface TagProvider {
```
kezz, line 620: Bit of a lack of nulability annotations in this class - I know it's only internal so it doesn't really matter, but worth throwing them in just for ease of contributing.

text-minimessage/src/main/java/net/kyori/adventure/text/minimessage/parser/match/StringResolvingMatchedTokenConsumer.java
```diff
@@ -68,26 +68,16 @@ public void accept(final int start, final int end, final @NotNull TokenType toke
...
-          // we only care about string placeholders!
-          if (value instanceof String) {
-            this.builder.append((String) value);
-            return;
-          }
-        }
-
-        // if we get here, the placeholder wasn't found or was null
-        this.builder.append(match);
+      if (replacement instanceof PreProcess) {
```
kezz, line 74: This is _so_ much nicer than the previous method. Means you can lazy-load expensive replacements rather than having to provide them every time.

MiniDigger (commented):
very minor stuff. 
I was a bit confused by ClickTag not implementing anything and ColorTagResolver being a tag resolver. I also not sure if I like that basically everything is static in the tags classes. Wouldn't they be tag resolvers too, just really simple ones? that would clean up the buildintags class a bit where I got a bit confused too (I guess the comment there can be discared now that I looked at the tags, lol)
like, maybe SimpleTag that is a tag resolver itself?

but overall, outstanding work, I really like how it turned out so far!

text-minimessage/src/main/java/net/kyori/adventure/text/minimessage/MiniMessageParser.java
MiniDigger: (looking back at it, I saw that the concept of placeholders still exist)

text-minimessage/src/main/java/net/kyori/adventure/text/minimessage/tag/Inserting.java
```diff
@@ -21,41 +21,35 @@
...
+   * Return the component this tag produces.
+   *
+   * <p>It may be desirable to store the result of this method.</p>
    *
-   * @return the transformed component
+   * @return the component this tag produces
    * @since 4.10.0
    */
-  public abstract Component apply();
+  Component value();
```
kezz: Do we need a nullability annotation on this?

text-minimessage/src/main/java/net/kyori/adventure/text/minimessage/tag/Inserting.java
```diff
@@ -21,41 +21,35 @@
...
-
-  protected Transformation() {
-  }
-
+public /* non-sealed */ interface Inserting extends Tag {
   /**
-   * Return a transformed {@code component} based on the applied parameters.
+   * Return the component this tag produces.
+   *
+   * <p>It may be desirable to store the result of this method.</p>
```
kezz: Is this actually true anymore?

text-minimessage/src/main/java/net/kyori/adventure/text/minimessage/tag/Inserting.java
```diff
@@ -21,41 +21,35 @@
...
- *
- * @see TransformationRegistry to access and register available transformations
  * @since 4.10.0
  */
-public abstract class Transformation implements Examinable {
-
-  protected Transformation() {
-  }
-
+public /* non-sealed */ interface Inserting extends Tag {
```
kezz, line 34: Are people allowed to implement this?

text-minimessage/src/main/java/net/kyori/adventure/text/minimessage/tag/builtin/BuiltInTags.java
```diff
@@ -0,0 +1,186 @@
...
+  }
+
+  /**
+   * Get a resolver that handles all built-in tags.
+   *
+   * @return the resolver for built-in tags
+   * @since 4.10.0
+   */
+  public static TagResolver all() {
+    return TagResolver.builder()
```
kezz: Is there a reason this returns a new `TagResolver` every time?

text-minimessage/src/main/java/net/kyori/adventure/text/minimessage/tag/builtin/ClickTag.java
```diff
@@ -0,0 +1,56 @@
...
+ */
+@ApiStatus.Internal
+public final class ClickTag {
+  public static final String CLICK = "click";
+
+  private ClickTag() {
+  }
+
+  static Tag create(final ArgumentQueue args, final Context ctx) throws ParsingException {
+    final String actionName = args.popOr(() -> "A click tag requires an action of one of " + ClickEvent.Action.NAMES.keys()).lowerValue();
```
kezz: Does the set toString implementation actually produce a useful result here?

text-minimessage/src/main/java/net/kyori/adventure/text/minimessage/tag/builtin/GradientTag.java
```diff
@@ -21,28 +21,31 @@
  * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
  * SOFTWARE.
  */
-package net.kyori.adventure.text.minimessage.transformation.inbuild;
+package net.kyori.adventure.text.minimessage.tag.builtin;
```
kezz: I think we should rename this `standard` rather than `builtin` - I can see a case where we might want to include extra tags that aren't included in the standard tag resolver but are still shipped with MiniMessage (see the colour alias issue for example).

text-minimessage/src/main/java/net/kyori/adventure/text/minimessage/tag/resolver/ArgumentQueue.java
```diff
@@ -0,0 +1,93 @@
...
+import org.jetbrains.annotations.NotNull;
+import org.jetbrains.annotations.Nullable;
+
+/**
+ * A queue of {@link Tag} arguments.
+ *
+ * @since 4.10.0
+ */
+@ApiStatus.NonExtendable
+public interface ArgumentQueue {
```
kezz, line 38: Is there any particular reason this isn't an actual queue, just with some additional popOr methods?

text-minimessage/src/main/java/net/kyori/adventure/text/minimessage/tag/resolver/TagResolver.java
```diff
@@ -0,0 +1,403 @@
...
+
+  /**
+   * Create a tag resolver that only responds to certain tag names, and whose value does not depend on that name.
+   *
+   * @param names the names to respond to
+   * @param handler the tag handler, may throw {@link ParsingException} if provided arguments are in an invalid format
+   * @return a resolver that creates tags using the provided handler
+   * @since 4.10.0
+   */
+  static @NotNull TagResolver resolver(final @NotNull Set<String> names, final @NotNull BiFunction<ArgumentQueue, Context, Tag> handler) {
```
kezz, line 121: Would a `Predicate<String>` version of this method be useful?

kezz (changes requested):
Incredibly solid bit of work here - I really don't have much to comment at all beyond saying good job! I've left a few minor nitpicky comments, mainly documentation changes - but these aren't particularly blocking as I'm sure we'll clean up the actual docs upon the release of 4.10.0 which should really help users get used to this change.

Well done!

text-minimessage/src/main/java/net/kyori/adventure/text/minimessage/tag/builtin/ClickTag.java
rymiel: Yes, according to the test: https://github.com/KyoriPowered/adventure/pull/672/files/236d8d15218292b781d9a14a5337d095a4f5c462#diff-9f1c821034fa93ae64573f4cb6c7150977dc024b82c3095a74db17d05c57df12R389
https://github.com/KyoriPowered/adventure/blob/236d8d15218292b781d9a14a5337d095a4f5c462/text-minimessage/src/test/java/net/kyori/adventure/text/minimessage/MiniMessageTest.java#L389

kezz: Neat!

text-minimessage/src/main/java/net/kyori/adventure/text/minimessage/tag/Inserting.java
MiniDigger: yes, they should be, it should be API, for people to build their own tags easily

kezz: Well, there are static constructor methods on `Tag` that create instances of these tags already. Although I suppose this could work for lazy insertions? Although maybe we should just add a supplier insertion?

text-minimessage/src/main/java/net/kyori/adventure/text/minimessage/MiniMessageParser.java
zml2008: I'd rather not -- it would invite questions from users implementing tags on how to use `examination`, or potential incorrect usage if they don't override `toString()` themselves.

text-minimessage/src/main/java/net/kyori/adventure/text/minimessage/MiniMessageParser.java
zml2008: The `token` passed to this method is potentially null (in the case of string substitutions/preprocess placeholder handling), so we don't want to set the exception's tokens to an array with a null element. The null check could be at a more inner scope, but there isn't really any functionality difference so I kept everything within one `if` statement.

text-minimessage/src/main/java/net/kyori/adventure/text/minimessage/ParsingException.java
zml2008: I don't think there's any reason they couldn't?

text-minimessage/src/main/java/net/kyori/adventure/text/minimessage/tag/resolver/ArgumentQueue.java
zml2008: I didn't want to expose the mutation methods to end users. We also don't want to actually `pop` elements from the queue, just advance a pointer for the current element.

text-minimessage/src/main/java/net/kyori/adventure/text/minimessage/tag/resolver/TagResolver.java
zml2008: It might be *useful*, but one of my plans for later on is to be able to consolidate tag resolvers created using this method into a map lookup, rather than the iteration that `SequentialTagResolver` does currently, see #677. A `Predicate` method would prevent us from knowing all names ahead of time.

Since users can implement their own `TagResolver`s, I don't think it's necessary to add a Predicate method at the moment. I also doubt how useful it would be in this specific case, since the factory method here is specifically *not* passed the tag name. If it does turn into something people want in the future, that API could always be added at a later date.

text-minimessage/src/main/java/net/kyori/adventure/text/minimessage/tag/Inserting.java
zml2008: I'd intended for end-users to be allowed to implement their own versions of `Inserting`/`PreProcess`/`Modifying` for any more complex logic, with the factory methods existing as shorthand for simple cases.

text-minimessage/src/main/java/net/kyori/adventure/text/minimessage/tag/builtin/BuiltInTags.java
zml2008: Better encapsulation, and lets us change the layout of tags down the line and provide the same API surface for tags that are provided with MiniMessage but not part of the standard collection/created by default. Specifically I'm thinking of the PR to add a color tag that supports CSS colour names -- we could have `BuiltInTags.color()` for the default flag, that just returns a field, and `BuiltInTags.color(Map<String, TextColor> aliases)` that creates a new `TagResolver` each time. Users wouldn't have to think about both field-tags and method-tags.

text-minimessage/src/main/java/net/kyori/adventure/text/minimessage/MiniMessageParser.java
kezz: Well, I moreso meant that the `instanceof` already handles the null check

text-minimessage/src/main/java/net/kyori/adventure/text/minimessage/tag/resolver/TagResolver.java
kezz: Sounds good to me!

text-minimessage/src/main/java/net/kyori/adventure/text/minimessage/MiniMessageParser.java
zml2008: `instanceof` is checking `e` though, not `token`?

kezz: Oh yeah, sorry about that :')

text-minimessage/src/main/java/net/kyori/adventure/text/minimessage/tag/builtin/BuiltInTags.java
zml2008: hm, an alternative could be, for each tag:

```java
final class ClickTag {
  static final String CLICK = "click";
  static final TagResolver RESOLVER =TagResolver.resolver("click", ClickTag::create);
}
```

which would break things up more?

I do think the structure of tag classes will change some once we fix up serialization though -- I want serialization to be associated with each tag resolver, rather than one god class. Since serialization fixes are still part of my goals for 4.10.0, are you fine with pushing some of that discussion off until I have a serialization PR ready to go?

text-minimessage/src/main/java/net/kyori/adventure/text/minimessage/parser/TokenParser.java
zml2008: this is for syntax extensions associated with #523, only relevant once actual work begins on that issue.

text-minimessage/src/main/java/net/kyori/adventure/text/minimessage/tag/builtin/BuiltInTags.java
MiniDigger: yep, fine to move that out of this PR, its big enough as is. 
totally agree with having serialization and deserialization in the same tag class, that way it should be organized. 

(originally, minimessage was just one class, moving out to transformations (new tags) has been a huge improvement, its just that the serializer didn't get that treatment yet)

text-minimessage/src/main/java/net/kyori/adventure/text/minimessage/Context.java
```diff
@@ -24,32 +24,63 @@
...
    *
-   * @param message the message to parse
-   * @return the parsed message
+   * @param message a detail message describing the error
+   * @param tags the tag parts which caused the error
+   * @return the new parsing exception
    * @since 4.10.0
    */
-  @NotNull Component parse(final @NotNull String message);
+  @NotNull ParsingException newError(
```
kezz: `newException` perhaps?

text-minimessage/src/main/java/net/kyori/adventure/text/minimessage/tag/Inserting.java
```diff
@@ -21,37 +21,34 @@
...
-  }
+  @NotNull Component value();
 
-  @Override
-  String valueName() {
-    return "PlaceholderNode";
+  /**
+   * Get whether or not this tag allows children.
+   *
+   * <p>If children are not allowed, this tag will be auto-closing, and should not be closed explicitly.</p>
```
kezz: What happens if you do close them? Does it error or get ignored it stripped?

text-minimessage/src/main/java/net/kyori/adventure/text/minimessage/tag/Placeholder.java
```diff
@@ -21,74 +21,74 @@
...
-    return component(key, Component.text(value));
+  public static TagResolver.@NotNull Single unparsed(final @NotNull String key, final @NotNull String value) {
+    requireNonNull(value, "value");
+    return Placeholder.component(key, Component.text(value));
   }
 
   /**
    * Creates a replacement that inserts a component.
    *
+   * <p>This replacement is auto-closing, so its style will not influence the style of following components.</p>
```
kezz: Might it be worth adding a version that allows styles to bleed - would be useful for prefixes with role colours for example.

text-minimessage/src/main/java/net/kyori/adventure/text/minimessage/tag/resolver/TagResolver.java
```diff
@@ -0,0 +1,403 @@
...
+/**
+ * A collection of known tags.
+ *
+ * <p>A resolver can handle anywhere from a single tag, to a dynamically generated set of tags, returning a tag based on the provided name and arguments.</p>
+ *
+ * @see StandardTags
+ * @see Placeholder
+ * @since 4.10.0
+ */
+public interface TagResolver {
```
kezz, line 52: Perhaps not for this PR, but it would be nice if you could turn a resolver into a builder. Or perhaps an `andThen` default method?

text-minimessage/src/main/java/net/kyori/adventure/text/minimessage/tag/standard/FontTag.java
```diff
@@ -0,0 +1,56 @@
...
+  public static final String FONT = "font";
+
+  private FontTag() {
+  }
+
+  static Tag create(final ArgumentQueue args, final Context ctx) throws ParsingException {
+    final Key font;
+    final @Subst("empty") String valueOrNamespace = args.popOr("A font tag must have either arguments of either <value> or <namespace:value>").value();
+    if (!args.hasNext()) {
+      font = Key.key(valueOrNamespace);
```
kezz: If the argument contains bad characters, this exception should be wrapped in a parse exception, similar to the hover events

text-minimessage/src/main/java/net/kyori/adventure/text/minimessage/tag/standard/package-info.java
```diff
@@ -22,6 +22,8 @@
  * SOFTWARE.
  */
 /**
- * Placeholders.
+ * Built-in tags shipped with MiniMessage.
```
kezz, line 25: Should note these are standard tags that standard, included by default.

kezz (changes requested):

text-minimessage/src/main/java/net/kyori/adventure/text/minimessage/tag/resolver/TagResolver.java
zml2008: Could you open an issue for that?

text-minimessage/src/main/java/net/kyori/adventure/text/minimessage/tag/Placeholder.java
rymiel: Wouldn't that require significant restructuring? Pretty sure this comment is just clarifying what was the case before, right?

zml2008: This clarifies the behavior before, which is determined by `Inserting#allowsChildren` -- in this case it's `false`, but when creating a `Tag` with `Tag.inserting` the tag *does* allow children. I'm just not sure whether to have that be a boolean parameter or two separate methods

kezz (approved):
🎉

PR #685, opened 2022-02-14
Reorganize tests a bit, closes #679

### LINKED ISSUE #679: Organize MiniMessage Tests better
currently its a mess, minimessage parser test and minimessage test are kinda the same class, I always pretty much roll a dice about what to put where. together they are over 2000 lines of code, impossible to read.

I propose splitting up the tests by Tag, so every tag has a test for serializing and deserializing (or should that be two separate classes too?), then have one class that tests the escaping and stripping and stuff and maybe the debug functionality and one class that tests composed tags, adding custom tags, disabling some tags etc.

that should keep the tests more maintainable and also make it easy to stop where more tests are needed (do we track coverage somewhere?)

MiniDigger:


zml2008:
@MiniDigger would you be able to patch this up enough to be mergable? then I can base further serializer work on it :)

Issue #696, opened 2022-02-23
Namespaced `TagResolver` API


zml2008:
For situations where users may want to have multiple sources contribute to a `TagResolver`, it may make sense to have identify tags by `Key` rather than just a plain value.

This would have a default namespace, for example `adventure`, and other namespaces could be registered.

The API would look something like:

```java
interface TagResolver {
  interface NamespacedBuilder {
    self defaultNamespace(String);
    self namespaceResolver(String namespace, TagResolver resolver);
    self tag(Key tag, Tag);
    self tag(Key tag, BiFunction<ArgumentQueue, Context, Tag> factory);
    TagResolver build();
  }
}
```

This would allow platforms to collect custom tags that could safely be made available globally.

zml2008:
Due to the potential for ambiguity between namespaces and tags in the default namespace, as well as some potential additional discussion around how this could handle player-specific metadata, I think it's best to target this for 4.11.0.

PR #697, opened 2022-02-23
text-minimessage: Add a service for platforms to customize MiniMessage instances

zml2008:
I expect this to be helpful if #696 is added particularly, though other use cases may come up over time.


kezz (approved):
Does this also need a docs update that mentions implementations can customise the MiniMessage instance if desired?

zml2008:
I don't think so -- every other serializer has services, and all are marked as `ApiStatus.Internal`.

(though, thinking about that more, maybe it makes sense to add a `@PlatformApi` annotation to mark things intended to be used by native implementations, but not by end users?)

PR #701, opened 2022-02-23
Allow parsing with custom Tag Resolvers

### LINKED ISSUE #705: Rename `Context#parse` to `deserialize`
Should be consistent with the main serializer.

Joo200:
I want to create a new List tag to format lists.
My minimessage syntax is: `<list:'[entry format]':'[separator]>`

The entry format will have some placeholder tags so I modify the output of the list.
E.g.: `My unique list is: <list:'<red><name></red>':', '> - that's it.`

In java I would call:
```java
    miniMessage.deserialize(format, new ListResolver(Placeholder.parsed("name", "Joo200"), Placeholder.parsed("name", "MiniDigger")));
```

this should result in the output `My unique list is: Joo200, MiniDigger - that's it.`

Unfortunatly I'm unable to pass the Placeholder for "name" as TagResolver:
```java
    @Override
    public @Nullable Tag resolve(@NotNull String name, @NotNull ArgumentQueue arguments, @NotNull Context ctx) throws ParsingException {
        if (!this.has(name)) return null;

        String format = arguments.popOr("expected list format").value();
        String separator = arguments.popOr("expected separator").value();
        return Tag.inserting(resolvers.stream().map(r -> ctx.parse(format /*, r (this doesn't work)*/))
             .collect(Component.toComponent(ctx.parse(separator))));
    }
```

With this PR it should be possible to pass the Context dynamic Tag Resolvers. This would be really handy.

### DIFF
--- a/text-minimessage/src/main/java/net/kyori/adventure/text/minimessage/Context.java
+++ b/text-minimessage/src/main/java/net/kyori/adventure/text/minimessage/Context.java
@@ -43,7 +43,7 @@ public interface Context {
    * Parses a MiniMessage string using all the settings of this context.
    *
    * @param message the message to parse
-   * @param resolvers additional tag resolvers
+   * @param resolvers additional tag resolvers, added to all other resolvers in this parse, but taking priority in the event of a name overlap
    * @return the parsed message
    * @since 4.10.0
    */
--- a/text-minimessage/src/main/java/net/kyori/adventure/text/minimessage/Context.java
+++ b/text-minimessage/src/main/java/net/kyori/adventure/text/minimessage/Context.java
@@ -25,6 +25,7 @@ package net.kyori.adventure.text.minimessage;
 
 import net.kyori.adventure.text.Component;
 import net.kyori.adventure.text.minimessage.tag.resolver.ArgumentQueue;
+import net.kyori.adventure.text.minimessage.tag.resolver.TagResolver;
 import org.jetbrains.annotations.ApiStatus;
 import org.jetbrains.annotations.NotNull;
 import org.jetbrains.annotations.Nullable;
@@ -42,10 +43,11 @@ public interface Context {
    * Parses a MiniMessage string using all the settings of this context.
    *
    * @param message the message to parse
+   * @param resolvers additional tag resolvers
    * @return the parsed message
    * @since 4.10.0
    */
-  @NotNull Component parse(final @NotNull String message);
+  @NotNull Component parse(final @NotNull String message, final @NotNull TagResolver@NotNull... resolvers);
 
   /**
    * Create a new parsing exception.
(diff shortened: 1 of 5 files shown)


zml2008:
Could you add a test for this behavior, and rebase? otherwise, looks good

text-minimessage/src/main/java/net/kyori/adventure/text/minimessage/Context.java
```diff
@@ -42,10 +43,11 @@
    * Parses a MiniMessage string using all the settings of this context.
    *
    * @param message the message to parse
+   * @param resolvers additional tag resolvers
```
zml2008:
```suggestion
   * @param resolvers additional tag resolvers, added to all other resolvers in this parse, but taking priority in the event of a name overlap
```

just to make it a bit more clear how this interacts with existing choices

text-minimessage/src/main/java/net/kyori/adventure/text/minimessage/ContextImpl.java
```diff
@@ -106,8 +106,9 @@ void message(final @NotNull String message) {
   }
 
   @Override
-  public @NotNull Component parse(final @NotNull String message) {
-    return this.miniMessage.deserialize(requireNonNull(message, "message"), this.tagResolver);
+  public @NotNull Component parse(final @NotNull String message, final @NotNull TagResolver@NotNull... resolvers) {
+    return this.miniMessage.deserialize(requireNonNull(message, "message"),
+      TagResolver.resolver(this.tagResolver, TagResolver.resolver(resolvers)));
```
zml2008:
```suggestion
      TagResolver.builder().resolver(this.tagResolver).resolvers(resolvers).build());
```

zml2008 (changes requested):

text-minimessage/src/main/java/net/kyori/adventure/text/minimessage/Context.java
```diff
@@ -42,10 +43,11 @@
...
    * @return the parsed message
    * @since 4.10.0
    */
-  @NotNull Component parse(final @NotNull String message);
+  @NotNull Component parse(final @NotNull String message, final @NotNull TagResolver@NotNull... resolvers);
```
kezz: This should be a set of methods that matches the corresponding ones in the main serializer interface.

text-minimessage/src/main/java/net/kyori/adventure/text/minimessage/ContextImpl.java
kezz: Should be checking the provided array isn't null here.

kezz (changes requested):

text-minimessage/src/main/java/net/kyori/adventure/text/minimessage/Context.java
Joo200: I'm not too familiar with such api designs in java.
Why do you need overloaded methods without varargs here?

```java
ctx.parse("<red>Test");
ctx.parse("<red>bla<place>", Placeholder.parse("place", "bla"));
ctx.parse("<red>bla<array><of><places>", ...);
```
are all covered by the varargs TagResolver method. Or did I miss smth?

Joo200:
Modified and resolves https://github.com/KyoriPowered/adventure/issues/705 now

zml2008 (approved):

kezz (approved):

Issue #702, opened 2022-02-24
Refine the serialization API


zml2008:
The serialization system introduced in 4.10.0 is an initial design designed to get the basics of the new system out the door.

This is a tracking issue for feedback on this initial design, and larger proposals for improving that API. Any changes discussed here would be targeted for after the initial 4.10.0 release.

zml2008:
As it stands, here are some pet peeves about what I've done so far:

- The way SerializiableResolver.claimStyle and .claimComponent have relatively different layouts -- Style has an additional holder object returned, and are both part of `SerializationRegistry.Single`. `.Single` should probably be split up into two interfaces, one for style and one for component claim shorthands. If a resolver needed both, it could override `handle` directly.
- SerializableRegistry is an entirely separate interface that our public TagResolver isn't attached to -- would it make sense to unify those somehow?
- Style element claims are relatively stringly typed -- we should have some strictly defined collection of style claim keys.
- How do we provide tag-level configurability over which output formats to prefer? A hex color could be `<color:#abcdef>`, `<c:#abcdef>`, or `<#abcdef>`. It would be nice to configure this per-tag type.
- Can tags declare that they should be closed with arguments? so a `<color:blue>` is closed as `</color:blue>`

more to come, maybe?

zml2008:
- Clarify some of the language around self-closing/unclosable tags, to better distinguish between the `<tag/>` shorthand syntax, and cases where `<tag>` doesn't allow any children, so cannot be separately closed.

PR #703, opened 2022-02-24
Rework the MiniMessage serializer to consolidate behavior into resolvers

### LINKED ISSUE #513: Serializer not serializing decoration correctly
When serializing this component
`Component.text().append(Component.text("[", NamedTextColor.AQUA, TextDecoration.BOLD))
            .append(Component.text("Ultra", NamedTextColor.WHITE, TextDecoration.BOLD, TextDecoration.ITALIC))
            .append(Component.text("God", NamedTextColor.AQUA))
            .append(Component.text("++]").decorate(TextDecoration.BOLD))
            .append(Component.text().color(NamedTextColor.AQUA))
            .build()`

it returns this minimessage
`<aqua><bold>[</aqua><white><italic>Ultra</italic></bold></white><aqua>God<bold>++]</bold>`

which shows up like this in-game when being deserialized by minimessage
image

### LINKED ISSUE #526: Builder `strict(true)` method has no effect when serialising
Used latest build 4.2.0-SNAPSHOT before posting an issue.
Sample code:
```java
var minimessageSerializer = MiniMessage.builder()
        .strict(true)
        .build();

var legacy = "§eText";
var component = LegacyComponentSerializer.legacySection().deserialize(legacy);
var minimessage = minimessageSerializer.serialize(component);

System.out.println(minimessage);
```

Output:
`<yellow>Text`

Output expected:
`<yellow>Text</yellow>`

Json:
`{"color":"yellow","text":"Text"}`

zml2008:
Currently, this is only exposed internally, but after development in
future releases this may begin to be exposed more widely.

Fixes #513
Fixes #526

I've currently got the core of the system here, and all reproducible tags exposed to the serializer. Many tests fail due to differences in escaping and tag order. Much of this is subjective, and some will need reconciling with the quirks of the parser to ensure round-trippability of the whole system.

I would like review on this to be focused on correctness and style of the generated output, and API exposed on `MiniMessage` instances. API design review for the serialization system itself can be provided, but beyond small issues or complete blockers, I would like to delay discussion on the structure of the serialization API until after 4.10.0 has been released. All of the serialization system is internal API that is still subject to change, and it will likely change as we refine it later on. Design discussion is welcome in #702.


text-minimessage/src/main/java/net/kyori/adventure/text/minimessage/serializer/ArgumentQuotingPreference.java
```diff
@@ -0,0 +1,37 @@
...
+
+/**
+ * The preferred style of argument quoting when serializing a MiniMessage string.
+ *
+ * <p>This only applies to tag arguments, not the tag name itself.</p>
+ *
+ * @since 4.10.0
+ */
+public enum ArgumentQuotingPreference {
+  UNQUOTED,
```
kezz: Should perhaps provide some docs for each method here, at least a `@since` tag for now.

text-minimessage/src/main/java/net/kyori/adventure/text/minimessage/serializer/SerializableResolver.java
```diff
@@ -0,0 +1,186 @@
...
+
+  /**
+   * Check a key.
+   *
+   * <p>Temporary copy of {@code SingleResolver#checkKey(String)} until we have a better idea.</p>
+   *
+   * @param key the key
+   * @since 4.10.0
+   */
+  static void checkKey(final @NotNull String key) {
```
kezz, line 120: Annotate this/the other method with `@ApiStatus.Experimental` for now?

text-minimessage/src/main/java/net/kyori/adventure/text/minimessage/serializer/package-info.java
```diff
@@ -0,0 +1,32 @@
...
+ * SOFTWARE.
+ */
+/**
+ * Serializer for the MiniMessage language.
+ *
+ * <p>This is currently internal API, likely to change at any time. It may or may not ever become stable.</p>
+ *
+ * @since 4.10.0
+ */
+@org.jetbrains.annotations.ApiStatus.Internal
```
kezz, line 31: `@ApiStatus.Experimental` might be a better annotation for this?

text-minimessage/src/test/java/net/kyori/adventure/text/minimessage/SerializerCollectorTest.java
```diff
@@ -0,0 +1,98 @@
...
+
+  @Test
+  void testEscapesInText() {
+    final String output = this.serializeToString(c -> c.text("hello <world>"));
+    assertEquals("hello \\<world>", output);
+  }
+
+  @Test
+  void testEscapesInTag() {
+    final String output = this.serializeToString(c -> c.tag("mini_>message").text("hi").pop());
```
kezz, line 50: Should we allow this? Don't we only allow a-zA-Z as tag names?

text-minimessage/src/test/java/net/kyori/adventure/text/minimessage/tag/standard/DecorationTagTest.java
```diff
@@ -0,0 +1,33 @@
...
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+ * SOFTWARE.
+ */
+package net.kyori.adventure.text.minimessage.tag.standard;
+
+import org.junit.jupiter.api.Test;
+
+public class DecorationTagTest {
+  @Test
+  void testCompleteness() {
```
kezz, line 42: very complete :p

text-minimessage/src/test/java/net/kyori/adventure/text/minimessage/MiniMessageSerializerTest.java
```diff
@@ -161,7 +171,7 @@ void testContinuedClick() {
 
   @Test
   void testKeyBind() {
-    final String expected = "Press <key:key.jump> to jump!";
+    final String expected = "Press <key:key.jump></key> to jump!";
```
kezz: Is this required now? Or just done for explicitness?

kezz (changes requested):
I think this is a really nice API - makes a great addition to MiniMessage :)

text-minimessage/src/test/java/net/kyori/adventure/text/minimessage/MiniMessageSerializerTest.java
MiniDigger: thats just what the new serializer produces, since this tag is just inserting a component, the old behavior was "nicer" I guess?

text-minimessage/src/test/java/net/kyori/adventure/text/minimessage/SerializerCollectorTest.java
zml2008: It doesn't look like we actually restrict the tag name at all -- but we also don't support any escapes within unquoted tag content, if any are needed the part has to be quoted.

text-minimessage/src/test/java/net/kyori/adventure/text/minimessage/MiniMessageSerializerTest.java
zml2008: It's the difference between " to jump" being a child of the keybind component, and "to jump" being a sibling of the keybind component -- self-closing tags will make this component layout nicer.

text-minimessage/src/main/java/net/kyori/adventure/text/minimessage/serializer/package-info.java
zml2008: I'm not sure that coveys how much others should not use this API. There are a few things that bug me about this API that probably won't be changed before merge, so it will definitely be breaking at some point in the future.

Plus, the vast majority of external users won't have any way to implement their own serializers at this point -- the set of things that can be serialized is somewhat fixed based on the Vanilla component types and style attributes. Custom tags would be difficult to reliably rediscover based on the component tree.

text-minimessage/src/main/java/net/kyori/adventure/text/minimessage/serializer/SerializableResolver.java
zml2008: The whole package is marked internal anyways, I don't see the point.

text-minimessage/src/main/java/net/kyori/adventure/text/minimessage/serializer/ArgumentQuotingPreference.java
zml2008: tbh I don't know if this class will even stay around, I need to figure out what we want for output customization.

MiniDigger (approved):

kezz (approved):

PR #707, opened 2022-02-26
text-minimessage: Allow tags to be closed in one tag

### LINKED ISSUE #523: Add a way to prevent templates impacting remaining content

### LINKED ISSUE #704: Support self-closing tags in the parser
To be able to distinguish between components with no children and components who have all following content as a child, we should be able to have `<tag/>` be equivalent to `<tag></tag>` in the parser and serializer.

This would fully close #523.

zml2008:
This simplifies the language for tags with no children.

### DIFF
--- a/text-minimessage/src/main/java/net/kyori/adventure/text/minimessage/MiniMessageParser.java
+++ b/text-minimessage/src/main/java/net/kyori/adventure/text/minimessage/MiniMessageParser.java
@@ -99,6 +99,7 @@ final class MiniMessageParser {
           break;
         case OPEN_TAG:
         case CLOSE_TAG:
+        case OPEN_CLOSE_TAG:
           // extract tag name
           if (token.childTokens().isEmpty()) {
             sb.append(richMessage, token.startIndex(), token.endIndex());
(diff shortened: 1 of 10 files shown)


kezz (approved):

rymiel:
This doesn't implement #523, just adds the tokenization capabilities, right?
Is emitting lang and key tags as self-closing something we want? Maybe makes it imply those have children taking #523 into context

rymiel:
disregard the first part of that comment, `testIgnorableSelfClosable` addresses that

zml2008:
To support vanilla components properly, I made all of the standard tags able to have children, since all component types can have children normally. The forced-no children attribute on Inserting tags is really just for custom things -- so there should probably be an attribute on the serializer to generate those sort of tags without the closing slash.

The only situation where that's relevant in MM itself is probably the newline tag though, since that's theoretically serializable (though, due to compaction it often won't end up as a separate component).

rymiel (approved):
Today i learned those components can have children in vanilla, sounds messy

PR #709, opened 2022-02-26
text-minimessage: Add Placeholder to format Numbers and TemporalAccessors

### LINKED ISSUE #759: `ChoiceFormat` tag

Joo200:
It might be a good idea to have a decent way of formatting Numbers and Dates.
Currently I have to use `miniMessage.deserialize("Number is <no>", Placeholder.unparsed("no", String.valueOf(doubleValue)));`

Would be nice if it could be smth like `miniMessage.deserialize("Number is <no:'de-DE':'#,00'>, Formatter.formatNumber("no", doubleValue));` where `#,00` is the DecimalFormat used to convert the number and `de-DE` is the locale tag.

Same thing for Formats for Date, Time and both combined. `(miniMessage.deserialize("Date is <date:'YYYY-MM-DD'>", Placeholder.formatDate("date", myDateAsInstant)))`

This PR should add those things to Placeholders. It works already fine.
However I want to add fallback values to formatDate and formatNumber. Better display a "bad" formatted message than nothing.

Possible things:

- `<number:'locale':'format'>`
- `<number:'format'>`
- `<number:'locale'>`
- `<date:'pattern'>`

Missing stuff:

- [x] Pull Request for Docs
- [x] Fallback Value for the DecimalFormat
- [x] Fallback Value for the DateFormatter
- [x] Better determination whether a string is a locale or the format for the DecimalFormat

### DIFF
--- a/text-minimessage/src/main/java/net/kyori/adventure/text/minimessage/tag/Tag.java
+++ b/text-minimessage/src/main/java/net/kyori/adventure/text/minimessage/tag/Tag.java
@@ -25,7 +25,6 @@ package net.kyori.adventure.text.minimessage.tag;
 
 import java.util.Arrays;
 import java.util.Locale;
-import java.util.Optional;
 import java.util.OptionalDouble;
 import java.util.OptionalInt;
 import java.util.function.Consumer;
@@ -215,22 +214,6 @@ public /* sealed */ interface Tag /* permits Inserting, Modifying, ParserDirecti
         return OptionalDouble.of(Double.parseDouble(this.value()));
       } catch (final NumberFormatException ex) {
         return OptionalDouble.empty();
-      }
-    }
-
-    /**
-     * Try to return this argument as a {@code char}.
-     *
-     * <p>The optional will only be present if the value is a valid char.</p>
-     *
-     * @return an optional providing the value of this argument as an integer
-     * @since 4.10.0
-     */
-    default @NotNull Optional<Character> asChar() {
-      if (this.value().length() == 1) {
-        return Optional.of(this.value().charAt(0));
-      } else {
-        return Optional.empty();
       }
     }
   }
--- a/text-minimessage/src/main/java/net/kyori/adventure/text/minimessage/tag/Tag.java
+++ b/text-minimessage/src/main/java/net/kyori/adventure/text/minimessage/tag/Tag.java
@@ -25,6 +25,7 @@ package net.kyori.adventure.text.minimessage.tag;
 
 import java.util.Arrays;
 import java.util.Locale;
+import java.util.Optional;
 import java.util.OptionalDouble;
 import java.util.OptionalInt;
 import java.util.function.Consumer;
@@ -214,6 +215,22 @@ public /* sealed */ interface Tag /* permits Inserting, Modifying, ParserDirecti
         return OptionalDouble.of(Double.parseDouble(this.value()));
       } catch (final NumberFormatException ex) {
         return OptionalDouble.empty();
+      }
+    }
+
+    /**
+     * Try to return this argument as a {@code char}.
+     *
+     * <p>The optional will only be present if the value is a valid char.</p>
+     *
+     * @return an optional providing the value of this argument as an integer
+     * @since 4.10.0
+     */
+    default @NotNull Optional<Character> asChar() {
+      if (this.value().length() == 1) {
+        return Optional.of(this.value().charAt(0));
+      } else {
+        return Optional.empty();
       }
     }
   }
--- a/text-minimessage/src/main/java/net/kyori/adventure/text/minimessage/tag/resolver/Formatter.java
+++ b/text-minimessage/src/main/java/net/kyori/adventure/text/minimessage/tag/resolver/Formatter.java
@@ -47,9 +47,9 @@ public final class Formatter {
   /**
    * Creates a replacement that inserts a number as a component. The component will be formatted by the provided DecimalFormat.
    *
-   * <p>This tag expects a format as attribute. Refer to {@link DecimalFormat} for usable patterns.</p>
+   * <p>This tag accepts a locale, a format pattern, both or nothing as arguments. The locale has to be provided as first argument.</p>
    *
-   * <p>You can specify the decimal and grouping separator by adding those as another attribute.</p>
+   * <p>Refer to {@link Locale} for usable locale tags. Refer to {@link DecimalFormat} for usable patterns.</p>
    *
    * <p>This replacement is auto-closing, so its style will not influence the style of following components.</p>
    *
@@ -111,27 +111,7 @@ public final class Formatter {
    * @return the placeholder
    * @since 4.11.0
    */
-  public static @NotNull TagResolver choice(final @NotNull String key, final long number) {
-    return TagResolver.resolver(key, ((argumentQueue, context) -> {
-      final String format = argumentQueue.popOr("Format expected.").value();
-      final ChoiceFormat choiceFormat = new ChoiceFormat(format);
-      return Tag.inserting(context.deserialize(choiceFormat.format(number)));
-    }));
-  }
-
-  /**
-   * Creates a replacement that inserts a choice formatted text. The component will be formatted by the provided ChoiceFormat.
-   *
-   * <p>This tag expectes a format as attribute. Refer to {@link ChoiceFormat} for usable patterns.</p>
-   *
-   * <p>This replacement is auto-closing, so its style will not influence the style of following components.</p>
-   *
-   * @param key the key
-   * @param number the number
-   * @return the placeholder
-   * @since 4.11.0
-   */
-  public static @NotNull TagResolver choice(final @NotNull String key, final double number) {
+  public static @NotNull TagResolver choice(final @NotNull String key, final Number number) {
     return TagResolver.resolver(key, ((argumentQueue, context) -> {
       final String format = argumentQueue.popOr("Format expected.").value();
       final ChoiceFormat choiceFormat = new ChoiceFormat(format);
--- a/text-minimessage/src/main/java/net/kyori/adventure/text/minimessage/tag/resolver/Formatter.java
+++ b/text-minimessage/src/main/java/net/kyori/adventure/text/minimessage/tag/resolver/Formatter.java
@@ -114,11 +114,10 @@ public final class Formatter {
   public static TagResolver choice(final @NotNull String key, final long number) {
     return TagResolver.resolver(key, ((argumentQueue, context) -> {
       final String format = argumentQueue.popOr("Format expected.").value();
-      ChoiceFormat choiceFormat = new ChoiceFormat(format);
+      final ChoiceFormat choiceFormat = new ChoiceFormat(format);
       return Tag.inserting(context.deserialize(choiceFormat.format(number)));
     }));
   }
-
 
   /**
    * Creates a replacement that inserts a choice formatted text. The component will be formatted by the provided ChoiceFormat.
@@ -135,7 +134,7 @@ public final class Formatter {
   public static TagResolver choice(final @NotNull String key, final double number) {
     return TagResolver.resolver(key, ((argumentQueue, context) -> {
       final String format = argumentQueue.popOr("Format expected.").value();
-      ChoiceFormat choiceFormat = new ChoiceFormat(format);
+      final ChoiceFormat choiceFormat = new ChoiceFormat(format);
       return Tag.inserting(context.deserialize(choiceFormat.format(number)));
     }));
   }
--- a/text-minimessage/src/main/java/net/kyori/adventure/text/minimessage/tag/resolver/Formatter.java
+++ b/text-minimessage/src/main/java/net/kyori/adventure/text/minimessage/tag/resolver/Formatter.java
@@ -1,0 +1,100 @@
+/*
+ * This file is part of adventure, licensed under the MIT License.
+ *
+ * Copyright (c) 2017-2022 KyoriPowered
+ *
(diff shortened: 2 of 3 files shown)


text-minimessage/src/main/java/net/kyori/adventure/text/minimessage/tag/resolver/Placeholder.java
```diff
@@ -81,4 +86,42 @@ private Placeholder() {
...
+   *
+   * @param key the key
+   * @param number the number
+   * @return the placeholder
+   * @since 4.10.0
+   */
+  public static TagResolver formatNumber(final @NotNull String key, final Number number) {
+    return TagResolver.resolver(key, (argumentQueue, context) -> {
+      final String format = argumentQueue.popOr("Format expected.").value();
+      return Tag.selfClosingInserting(Component.text(new DecimalFormat(format, new DecimalFormatSymbols(Locale.ROOT)).format(number)));
```
Joo200: Missing a good possibility to remove DecimalFormatSymbols with a better approach.
Unfortunatly DecimalFormat doesn't allow changing the separator in the format.

text-minimessage/src/main/java/net/kyori/adventure/text/minimessage/tag/resolver/Placeholder.java
```diff
@@ -81,4 +86,42 @@ private Placeholder() {
   public static TagResolver.@NotNull Single component(final @NotNull String key, final @NotNull ComponentLike value) {
     return TagResolver.resolver(key, Tag.selfClosingInserting(value));
   }
+
+  /**
+   * Creates a replacement that inserts a number as a component. The component will be formatted by the provided DecimalFormat.
+   *
+   * <p>Refer to https://docs.oracle.com/javase/7/docs/api/java/text/DecimalFormat.html for usable patterns.</p>
```
kezz: `{@link DecimalFormat}`

text-minimessage/src/main/java/net/kyori/adventure/text/minimessage/tag/resolver/Placeholder.java
```diff
@@ -81,4 +86,42 @@ private Placeholder() {
   public static TagResolver.@NotNull Single component(final @NotNull String key, final @NotNull ComponentLike value) {
     return TagResolver.resolver(key, Tag.selfClosingInserting(value));
   }
+
+  /**
+   * Creates a replacement that inserts a number as a component. The component will be formatted by the provided DecimalFormat.
```
kezz: I think this needs elaborating a bit - perhaps to say that it accepts one argument that is the format.

kezz (changes requested):
Also, I don't really think this fits in `Placeholder` as nothing else in this class supports arguments. Although I suppose it technically it a placeholder as it just inserts one value but just optionally has formatting? Not sure, would like to hear others' thoughts.

MiniDigger:
maybe we can have a TagResolvers class for "utils" like this?

text-minimessage/src/main/java/net/kyori/adventure/text/minimessage/tag/resolver/Placeholder.java
Joo200: I added some explanation for it. This should cover the use cases.

Joo200:
> maybe we can have a TagResolvers class for "utils" like this?

I moved it to a new Formatter class. This might be not the final name, Maybe "FormatterTags" would be better?

And I added a syntax to allow formatting the decimal and grouping separator. The syntax is currently:
`<name:'#,00':'AB'>`, where A is the decimal separator and B is the grouping separator. The second argument is optional, the grouping separator is optional too.

Should I convert the decimal and grouping separator to single arguments? So the syntax would be instead `<name:'#,00':'A':'B'>`?

text-minimessage/src/main/java/net/kyori/adventure/text/minimessage/tag/resolver/Formatter.java
```diff
@@ -0,0 +1,96 @@
...
+   * @param key the key
+   * @param number the number
+   * @return the placeholder
+   * @since 4.10.0
+   */
+  public static TagResolver formatNumber(final @NotNull String key, final Number number) {
+    return TagResolver.resolver(key, (argumentQueue, context) -> {
+      final String format = argumentQueue.popOr("Format expected").value();
+      final DecimalFormat decimalFormat = new DecimalFormat(format);
+      final DecimalFormatSymbols formatSymbols = new DecimalFormatSymbols(Locale.ROOT);
```
zml2008: Do we want to be able to provide a locale here? or as an argument to the method? That would allow doing some level of localized number formatting

Joo200: Yea, that would be a good idea. Imo Localisation should be in the tag. E.g. I want to use MiniMessage to parse localized messages for different languages. Passing the Locale to the method wouldn't really help in this case.

Maybe better add it as an argument instead? I'm not sure what's a good syntax for that, I welcome every suggestion.

zml2008: I was sort of thinking of offering some sort of typed map on the parse context that would hold the locale? maybe base off of pointers or something?

That way, you could do sth like:

```java
Component localized(final Player player, final String key) {
  final String message = this.messagesBundle.get(player.locale(), key);
  return this.mm.deserialize(message, player.pointers(), /* any extra meta */);
}
```

or more generally:

```java
Component localized(final Locale lang, final String key) {
  final String message = this.messagesBundle.get(lang, key);
  return this.mm.deserialize(message, Pointers.builder().withStatic(Identity.LOCALE, lang).build(), /* any extra meta */);
}
```

does that seem reasonable? is it too tied in with adventure?

Joo200: I don't think that's a good idea. It might be better to provide the locale in a tag?
Smth like `<lang:DE_de>Number should be formatted with , separator: <num:'#.00'>`?

I'm not sure but maybe there will be multi language text?
E.g. `<lang:EN_en>The translation of the text <blue>The number is <num:'#.00'></blue> in German is:<lang:DE_de><blue> Die Nummer ist <num:'#.00></blue>` That might be a edge case but Imo it should be covered by minimessage.

Joo200:
Currently waiting for https://github.com/KyoriPowered/adventure/pull/735

zml2008:
Can you explain why you chose to make these pre-process tags? I think they'd be better suited as `inserting` tags, unless the formatted date can somehow contain MiniMessage tags.

Joo200:
The idea is this:
`"<double:'en-EN':'<green>#.00;<red>-#.00'><blue> is a nice number"`
This is a format which will display positive numbers in green and negative numbers in red. This is only possible with PreProcess tags.

zml2008:
Hm, that makes sense. Do we want style to bleed through though? it might make more sense to just parse the output of the formatter as components.

Joo200:
I'm not sure. Using PreProcess tags should be flexible here.

E.g. a gradient tag should work with the formatting in the tag. We can change it to a inserting tag instead too,

zml2008:
Gradient tags should work with uncoloured `Inserting` tags as children?

I'd like to avoid pre-process tags since they have such special-cased handling within the parser at the moment, which causes problems with error message offsets mostly.

Joo200:
Yes, you're correct. I'll change it to `inserting`.

Joo200:
I rebased the PR and squashed the commits.

The class is now Formatter and I changed some stuff to be more intuitive. E.g. detecting automatically whether you provide a locale arg, a formatting arg or both.

juanvlr:
Nice work !
It would be good to handle choice formats as well (https://docs.oracle.com/javase/7/docs/api/java/text/ChoiceFormat.html)
Exemple :
There <choice:'1#is one player':'2>are \<number\> players'> on the server

text-minimessage/src/main/java/net/kyori/adventure/text/minimessage/tag/resolver/Formatter.java
```diff
@@ -0,0 +1,100 @@
...
+   * <p>You can specify the decimal and grouping separator by adding those as another attribute.</p>
+   *
+   * <p>This replacement is auto-closing, so its style will not influence the style of following components.</p>
+   *
+   * @param key the key
+   * @param number the number
+   * @return the placeholder
+   * @since 4.10.0
+   */
+  public static TagResolver formatNumber(final @NotNull String key, final Number number) {
```
kezz: Needs some nullability annotations (and on the other method too). Also, perhaps overloads for primitives? Might be a bit boring and annoying tbh.

text-minimessage/src/main/java/net/kyori/adventure/text/minimessage/tag/resolver/Formatter.java
```diff
@@ -0,0 +1,100 @@
...
+   * <p>This tag expects a format as attribute. Refer to {@link DecimalFormat} for usable patterns.</p>
+   *
+   * <p>You can specify the decimal and grouping separator by adding those as another attribute.</p>
+   *
+   * <p>This replacement is auto-closing, so its style will not influence the style of following components.</p>
+   *
+   * @param key the key
+   * @param number the number
+   * @return the placeholder
+   * @since 4.10.0
```
kezz: All need updating to 4.11.0

kezz (changes requested):
Minor nitpicks before merging!

Joo200:
> Nice work ! It would be good to handle choice formats as well (https://docs.oracle.com/javase/7/docs/api/java/text/ChoiceFormat.html) Exemple : There <choice:'1#is one player':'2>are <number> players'> on the server

I agree, that would be nice too. I'll take a look at the weekend. Should be pretty easy

text-minimessage/src/main/java/net/kyori/adventure/text/minimessage/tag/resolver/Formatter.java
Joo200: I added the @NotNull annotations. I'm not sure whether we should add the different number primitives.

Joo200:
Added ChoiceFormat tags in the last commit.
The syntax is `<tagname:'pattern'>` and can be used with Formatter.choice(String tagname, double|long number) 

This will add https://github.com/KyoriPowered/adventure/issues/759 to adventure

text-minimessage/src/main/java/net/kyori/adventure/text/minimessage/tag/resolver/Formatter.java
kezz: Need them on the return types too! Yeah, idk about different primitives - maybe someone else can weigh in, @zml2008?

zml2008: I'm not sure it's necessary? Number is prooobably fine?

text-minimessage/src/main/java/net/kyori/adventure/text/minimessage/tag/Tag.java
```diff
@@ -216,5 +217,21 @@ default boolean isFalse() {
...
+
+    /**
+     * Try to return this argument as a {@code char}.
+     *
+     * <p>The optional will only be present if the value is a valid char.</p>
+     *
+     * @return an optional providing the value of this argument as an integer
+     * @since 4.10.0
+     */
+    default @NotNull Optional<Character> asChar() {
```
kezz: Is this used anywhere? Why was this added?

text-minimessage/src/main/java/net/kyori/adventure/text/minimessage/tag/resolver/Formatter.java
```diff
@@ -0,0 +1,141 @@
...
+ * @since 4.11.0
+ */
+public final class Formatter {
+  private Formatter() {
+  }
+
+  /**
+   * Creates a replacement that inserts a number as a component. The component will be formatted by the provided DecimalFormat.
+   *
+   * <p>This tag expects a format as attribute. Refer to {@link DecimalFormat} for usable patterns.</p>
```
kezz: This isn't quite correct as it does fallback to `DecimalFormat.getInstance()`.

text-minimessage/src/main/java/net/kyori/adventure/text/minimessage/tag/resolver/Formatter.java
```diff
@@ -0,0 +1,141 @@
...
+   * <p>This tag expectes a format as attribute. Refer to {@link ChoiceFormat} for usable patterns.</p>
+   *
+   * <p>This replacement is auto-closing, so its style will not influence the style of following components.</p>
+   *
+   * @param key the key
+   * @param number the number
+   * @return the placeholder
+   * @since 4.11.0
+   */
+  public static @NotNull TagResolver choice(final @NotNull String key, final long number) {
```
kezz: Could this just be one function with a `Number` arg as well?

kezz (changes requested):
Minor comments - also needs an accompanying docs PR.

text-minimessage/src/main/java/net/kyori/adventure/text/minimessage/tag/Tag.java
Joo200: I tested some stuff but I don't know why I added this. I'll remove it.

kezz (approved):
LGTM (pending docs pr)! Thanks for getting to all the requested changes and such :) Very excited to start using these tags!

zml2008 (approved):

PR #714, opened 2022-02-28
text-minimessage: Expose start/end indices on parsing exceptions

zml2008:
This allows, for example, command libraries to get the actual index where parsing failed when reporting errors.

(note before this is actually merged: would be nice to expose the raw exception message, not just the one formatted with location information)


kezz (approved):
Also agree that exposing the constructor passed message would be useful as well. Otherwise, LGTM.

PR #717, opened 2022-02-28
text-minimessage: Limit characters allowed in tag names

kezz:
Limits tag names to the pattern `[!?#]?[a-z0-9_-]*`.

Or, in plain text:
> Tag names must only contain the characters a-z, 0-9, `_`, and `-`. They can also optionally start with any of the following characters: `!?#`.

PR #723, opened 2022-03-03
minimessage: Ignore invalid tags when parsing

rymiel:
As of #717, these would cause a parse error, but they should just be outputted literally, as to not break on arbitrary input

Note: pretty ugly fix, adding a second tag checker helper and using a pretty magical children-traversing method, but i didn't want to resort to a try-catch, please advise if something like that would be better

### DIFF
--- a/text-minimessage/src/main/java/net/kyori/adventure/text/minimessage/internal/parser/node/TagNode.java
+++ b/text-minimessage/src/main/java/net/kyori/adventure/text/minimessage/internal/parser/node/TagNode.java
@@ -26,7 +26,6 @@ package net.kyori.adventure.text.minimessage.internal.parser.node;
 import java.util.ArrayList;
 import java.util.List;
 import java.util.Objects;
-import net.kyori.adventure.text.minimessage.internal.TagInternals;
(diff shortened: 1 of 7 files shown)


kezz (approved):

text-minimessage/src/main/java/net/kyori/adventure/text/minimessage/internal/parser/node/TagNode.java
```diff
@@ -68,7 +68,7 @@ public TagNode(
 
     // Then assert the tag node has a proper name.
     try {
-      TagInternals.sanitizeAndCheckTagName(this.name());
+      TagInternals.sanitizeAndAssertValidTagName(this.name());
```
zml2008: is this still needed? since we don't create tag nodes anyways for things with invalid names

rymiel: It never fails when running the minimessage test suite so let's assume it's not needed :p

zml2008 (approved):

PR #730, opened 2022-03-08
Move to version catalogs and properties-based configuration

zml2008:
With these changes, no regular version bumps should require modifying the project's buildscripts directly. This should allow buildscript compilation to be cached with less invalidation.


bom/build.gradle.kts
```diff
@@ -1,5 +1,5 @@
 plugins {
-  id("java-platform")
+  `java-platform`
```
kashike: Not a fan of this change.

build-logic/settings.gradle.kts
```diff
@@ -0,0 +1,21 @@
...
+      mavenContent { snapshotsOnly() }
+    }
+  }
+  
+  versionCatalogs {
+    register("libs") {
+      from(files("../gradle/libs.versions.toml")) // include from parent project
+    }
+  }
+}
```
kashike: newline

kezz (approved):
I suppose plugins in the common conventions build logic can't use typesafe accessors still right?

zml2008:
Unfortunately no -- since convention plugin plugin blocks can't actually affect the classpath, they're just for application.

PR #734, opened 2022-03-25
Cache component's hashcode for faster equality check

TheMode:
Drastically improve `TextComponent#equals` performance when the two components are "constant" (used for equality more than once) and not equals. Dynamic components are not affected by this change.

OpenJDK 18 bench result on `main
```
Benchmark                                          Mode  Cnt    Score    Error  Units
ComponentEqualityBenchmark.constantAlmostEquals    avgt   30   71.454 ± 14.591  ns/op
ComponentEqualityBenchmark.constantEquals          avgt   30   76.149 ± 14.636  ns/op
ComponentEqualityBenchmark.constantNotEquals       avgt   30   11.750 ±  0.090  ns/op
ComponentEqualityBenchmark.singleConstantEquals    avgt   30  248.790 ±  7.518  ns/op
ComponentEqualityBenchmark.singleConstantNotEqual  avgt   30   69.418 ±  8.904  ns/op
```
my branch:
```
Benchmark                                          Mode  Cnt    Score    Error  Units
ComponentEqualityBenchmark.constantAlmostEquals    avgt   30    2.501 ±  0.039  ns/op
ComponentEqualityBenchmark.constantEquals          avgt   30   83.614 ± 14.917  ns/op
ComponentEqualityBenchmark.constantNotEquals       avgt   30    2.495 ±  0.035  ns/op
ComponentEqualityBenchmark.singleConstantEquals    avgt   30  246.227 ±  6.951  ns/op
ComponentEqualityBenchmark.singleConstantNotEqual  avgt   30   79.535 ±  9.221  ns/op
```

I am getting some weird delta on both jdk 18 & 11, so another check would be appreciated.

Overall, I agree that such change could bring unnecessary complexity, but this does seem to help a lot and is IMO easily rollback-able/test-able.

### DIFF
--- a/api/src/jmh/java/net/kyori/adventure/text/ComponentEqualityBenchmark.java
+++ b/api/src/jmh/java/net/kyori/adventure/text/ComponentEqualityBenchmark.java
@@ -1,0 +1,103 @@
+package net.kyori.adventure.text;
+
+import net.kyori.adventure.text.format.NamedTextColor;
+import net.kyori.adventure.text.format.Style;
+import net.kyori.adventure.text.format.TextDecoration;
+import org.openjdk.jmh.annotations.*;
+import org.openjdk.jmh.infra.Blackhole;
+
+import java.util.concurrent.TimeUnit;
+
+import static net.kyori.adventure.key.Key.key;
+import static net.kyori.adventure.text.Component.text;
+import static net.kyori.adventure.text.format.Style.style;
+
+@Warmup(iterations = 5, time = 1000, timeUnit = TimeUnit.MILLISECONDS)
+@Measurement(iterations = 10, time = 1000, timeUnit = TimeUnit.MILLISECONDS)
+@Fork(3)
+@BenchmarkMode(Mode.AverageTime)
+@OutputTimeUnit(TimeUnit.NANOSECONDS)
+@State(Scope.Benchmark)
+public class ComponentEqualityBenchmark {
+
+  Component component1;
+  Component componentAlmost;
+  Component component2;
+
+  Component component3;
+
+  @Setup
+  public void setup() {
+    component1 = create1();
+    componentAlmost = createAlmost1();
+    component2 = create1();
+
+    component3 = create2();
+  }
+
+  @Benchmark
+  public void constantEquals(final Blackhole blackhole) {
+    blackhole.consume(component1.equals(component2));
+  }
+
+  @Benchmark
+  public void constantAlmostEquals(final Blackhole blackhole) {
+    blackhole.consume(component1.equals(componentAlmost));
+  }
+
+  @Benchmark
+  public void constantNotEquals(final Blackhole blackhole) {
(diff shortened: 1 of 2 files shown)


kashike:
Why?

kashike:
Additionally, this would be nuked when Adventure 5.0.0 moves to records. Is there really a point?

TheMode:
Because Component is a low-level object that can cause bottlenecks (or at least have a noticeable impact) depending on how they get used. This is our case in an experiment comparing components to detect changes (which is a questionable use-case, but bottlenecked for no obvious reason)

Is there any reasoning behind the use of records? I would argue that such micro-optimizations, even if niche, are worth more.

kashike:
> Is there any reasoning behind the use of records? I would argue that such micro-optimizations, even if niche, are worth more.

Because there's no reason not to do so - less code to maintain ourselves.

TheMode:
Well in this case you would trade performance for maintainability, which I can understand.

broccolai:
This would be better solved by users checking if their data has changed rather than using the component's equality.

Imagining a use-case for this: 
- The user wants to send a new scoreboard component with integer X.
- They construct a new component with X.
- They then use the hashcode comparison to check if the new component is equal to the current component.

If the component IS equal, they would have then constructed the component for no reason which goes against the performance goals of this PR. 

What would be much better, and not require any additional maintenance for Kyori, would be the users tracking the current components related data and checking equality with that against their new data. Promoting the use of the components hashcode for this feels like encouraging people's bad design.

If there's something I'm missing I'd love to know though.

TheMode:
I definitely agree that such use-case shouldn't be promoted, and that there are more efficient way to design such API.

But ignoring all that, it is still a fairly easy optimization improving equality check by up to x20 and can happen in existing code without promoting it.

Ultimately I do not plan to force this to be merged as I can understand the maintainability worry, although I still vote for.

kennytv:
A 70ns worst-case is nothing, even if you call it a million times a second, so that's not a bottleneck (and if you call it a billion times, see brocc's comment). this pr is only "easy" because you chose to single out text components instead of every component/the parent component where the actual complexity lies. aside from that, the "changed" boolean is implemented poorly with it being set *before* the cached hash is set in equals() and the check against 0 being bad in hashCode()

TheMode:
This 70ns is many time more than the allocation of the component itself. And can definitely be noticeable. For reference a simple map lookup takes roughly 4ns on my system.
I indeed took a simple case, and the most likely one. I do not see how this is a problem? Other components are not affected in any bad way.

I do not understand how the field order actually matter. The code is intentionally subject to benign data race, the fields are not part of the API, and the boolean is only present to prevent "dynamic"/one-time components (i.e. those that keep getting allocated during runtime) from suffering poor equals performance.

MiniDigger:
> And can definitely be noticeable.

please show so in a real-world scenario by providing a profiler report then

PR #735, opened 2022-03-27
MiniMessage: Handle arguments for PreProcessTags correctly

### LINKED ISSUE #732: PreProcess tags do not handle arguments correctly
Following unit test will fail:

```java
  @Test
  void testNumberStuff() {
    final String input = "<test:'bla'> is a nice number";
    final Component expected = text("bla is a nice number");

    this.assertParsedEquals(
      expected,
      input,
      TagResolver.resolver("test", (argumentQueue, context) -> {
        if (argumentQueue.hasNext()) argumentQueue.pop();
        return Tag.preProcessParsed("bla");
      })
    );
  }
```
The Tag.preProcessParsed doesn't allow arguments. It will just remove the tag and doesn't insert the text.

I debugged MiniMessage and I was able to find:
https://github.com/KyoriPowered/adventure/blob/9403962bb1afbd3392b580654f68403dcd426f6a/text-minimessage/src/main/java/net/kyori/adventure/text/minimessage/internal/parser/match/StringResolvingMatchedTokenConsumer.java
`String tag` is `"test:'bla'`, therefore `TagInternals.sanitizeAndCheckValidTagName(tag)` doesn't accept the tag as a valid tag and doesn't insert anything.

The arguments should be parsed and passed to the `tagProvider.resolve` call instead.

Joo200:
This should fix https://github.com/KyoriPowered/adventure/issues/732

The tests should be a good example how to use arguments in pre process tags.

Currently the code looks a little bit ugly. Maybe use the first token from the childTokens? Or something else.

### DIFF
--- a/text-minimessage/src/main/java/net/kyori/adventure/text/minimessage/internal/parser/match/StringResolvingMatchedTokenConsumer.java
+++ b/text-minimessage/src/main/java/net/kyori/adventure/text/minimessage/internal/parser/match/StringResolvingMatchedTokenConsumer.java
@@ -23,15 +23,22 @@
  */
 package net.kyori.adventure.text.minimessage.internal.parser.match;
 
+import java.util.ArrayList;
+import java.util.List;
 import java.util.Objects;
 import net.kyori.adventure.text.minimessage.internal.TagInternals;
+import net.kyori.adventure.text.minimessage.internal.parser.Token;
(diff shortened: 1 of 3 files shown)


zml2008 (approved):

kezz (approved):

kezz (dismissed):
Docs in `PreProcess` need updating to specify that they can accept arguments.

PR #740, opened 2022-04-01
Implement component interning for faster equality check

mdcfe:
Hey guys! I've been working on a cool optimisation feature for a while and now I'm finally ready to show this to you, going back to the all beloved heap! I have taken it upon myself to implement component interning, so here are a few examples of how it currently looks like:

Components have a simple, yet wonderful new method to them, with functionality the current components do not have:

`Component.intern()`!

Object churn is back to its old glory with amazing looking memory leaks: 

idea64.exe has stopped responding

Text components are as powerful as ever with different colors and clickable info links: 

Something happened
Something happened

And here is a preview of some benchmarks; instead of the component interning being slower than not interning the component, I have made it take longer, as a cheeky easter egg ;) 

Let's name your PCMake it your with a unique name that's easy to recognize when connecting to it from other devices. Your PC will restart after you name it.No more than 15 charactersNo spaces or any of the following special characters:

I am eager to receive feedback on this new optimisation, I'm sure it's finally going to be the last one! It simply needs some slight adjustments to the build script so it can build again.

### DIFF
--- a/api/src/main/java/net/kyori/adventure/text/AbstractComponent.java
+++ b/api/src/main/java/net/kyori/adventure/text/AbstractComponent.java
@@ -23,6 +23,7 @@
  */
 package net.kyori.adventure.text;
 
+import java.util.HashMap;
 import java.util.List;
 import java.util.Objects;
 import java.util.stream.Stream;
@@ -44,6 +45,8 @@ import org.jetbrains.annotations.Nullable;
 @Debug.Renderer(text = "this.debuggerString()", childrenArray = "this.children().toArray()", hasChildren = "!this.children().isEmpty()")
 @Deprecated
 public abstract class AbstractComponent implements Component {
+  private static final HashMap<Integer, Component> COMPONENT_POOL = new HashMap<>();
+
   protected final List<Component> children;
   protected final Style style;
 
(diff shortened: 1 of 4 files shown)


api/src/main/java/net/kyori/adventure/text/AbstractComponent.java
```diff
@@ -44,6 +45,8 @@
 @Debug.Renderer(text = "this.debuggerString()", childrenArray = "this.children().toArray()", hasChildren = "!this.children().isEmpty()")
 @Deprecated
 public abstract class AbstractComponent implements Component {
+  private static final HashMap<Integer, Component> COMPONENT_POOL = new HashMap<>();
```
Camotoy, line 48: The diamond operator here is just a decoration. You should remove it.

Owen1212055: I think the Diamond operator wants friends instead…



```suggestion
  private static final HashMap<Integer, Component> COMPONENT_POOL = new HashMap<Integer, Component>();
```

Camotoy: But functionally that still *does* nothing. It just adds more clutter to the code!

PR #751, opened 2022-04-14
fix: Fixed TagResolvers equality

4drian3d:


text-minimessage/src/main/java/net/kyori/adventure/text/minimessage/tag/resolver/EmptyTagResolver.java
```diff
@@ -56,4 +56,9 @@ public boolean contributeToMap(final @NotNull Map<String, Tag> map) {
   @Override
   public void handle(final @NotNull Component serializable, final @NotNull ClaimConsumer consumer) {
   }
+
+  @Override
+  public boolean equals(final Object other) {
```
zml2008: this is unneeded, equality uses object identity by default.

text-minimessage/src/main/java/net/kyori/adventure/text/minimessage/tag/resolver/CachingTagResolver.java
```diff
@@ -78,4 +79,24 @@ public void handle(final @NotNull Component serializable, final @NotNull ClaimCo
...
+
+  @Override
+  public boolean equals(final Object other) {
+    if (this == other) {
+      return true;
+    }
+    if (other == null) {
+      return false;
+    }
+    if (other.getClass() != CachingTagResolver.class) {
```
zml2008:
```suggestion
    if (!(other instanceof CachingTagResolver)) {
```

text-minimessage/src/main/java/net/kyori/adventure/text/minimessage/tag/resolver/CachingTagResolver.java
```diff
@@ -78,4 +79,24 @@ public void handle(final @NotNull Component serializable, final @NotNull ClaimCo
       ((SerializableResolver) this.resolver).handle(serializable, consumer);
     }
   }
+
+  @Override
+  public boolean equals(final Object other) {
```
zml2008:
```suggestion
  public boolean equals(final @Nullable Object other) {
```

text-minimessage/src/main/java/net/kyori/adventure/text/minimessage/tag/resolver/MapTagResolver.java
```diff
@@ -45,4 +46,24 @@ public boolean contributeToMap(final @NotNull Map<String, Tag> map) {
     map.putAll(this.tagMap);
     return true;
   }
+
+  @Override
+  public boolean equals(final Object other) {
```
zml2008:
```suggestion
  public boolean equals(final @Nullable Object other) {
```

text-minimessage/src/main/java/net/kyori/adventure/text/minimessage/tag/resolver/MapTagResolver.java
```diff
@@ -45,4 +46,24 @@ public boolean contributeToMap(final @NotNull Map<String, Tag> map) {
...
+    if (this == other) {
+      return true;
+    }
+    if (other == null) {
+      return false;
+    }
+    if (other.getClass() != MapTagResolver.class) {
```
zml2008:
```suggestion
    if (!(other instanceof MapTagResolver)) {
```

text-minimessage/src/main/java/net/kyori/adventure/text/minimessage/tag/resolver/SequentialTagResolver.java
```diff
@@ -89,4 +90,24 @@ public void handle(@NotNull final Component serializable, @NotNull final ClaimCo
       }
     }
   }
+
+  @Override
+  public boolean equals(final Object other) {
```
zml2008:
```suggestion
  public boolean equals(final @Nullable Object other) {
```

text-minimessage/src/main/java/net/kyori/adventure/text/minimessage/tag/resolver/SequentialTagResolver.java
```diff
@@ -89,4 +90,24 @@ public void handle(@NotNull final Component serializable, @NotNull final ClaimCo
...
+    if (other == this) {
+      return true;
+    }
+    if (other == null) {
+      return false;
+    }
+    if (other.getClass() != SequentialTagResolver.class) {
```
zml2008:
```suggestion
    if (!(other instanceof SequentialTagResolver)) {
```

zml2008 (changes requested):

4drian3d:
Applied suggested changes

zml2008 (approved):

text-minimessage/src/main/java/net/kyori/adventure/text/minimessage/tag/resolver/CachingTagResolver.java
```diff
@@ -78,4 +79,21 @@ public void handle(final @NotNull Component serializable, final @NotNull ClaimCo
...
+  @Override
+  public boolean equals(final @Nullable Object other) {
+    if (this == other) {
+      return true;
+    }
+    if (!(other instanceof CachingTagResolver)) {
+      return false;
+    }
+    final CachingTagResolver that = (CachingTagResolver) other;
+    return Objects.equals(this.cache, that.cache);
```
kezz: Do we not need to check the resolver here as well?

kezz (changes requested):

text-minimessage/src/main/java/net/kyori/adventure/text/minimessage/tag/resolver/CachingTagResolver.java
4drian3d: I don't think so, since it would be rare for someone to implement a TagResolver.WithoutArguments in a separate class (and even rarer to implement a proper equals method), everyone would use a lambda to implement it directly, making that between 2 identically created instances of CachingTagResolver and applying the same arguments to them, the only different thing is the resolver.
In case the TagResolver.WithoutArguments was a CachingTagResolver, in the only method where this class is created it will return the same resolver casted to a CachingTagResolver

zml2008: the cache should not be part of equality either way - it records thd output if thd underlying resolver, and two caching resolvers can draw from the same backing resolvers while having different sets of keys in their cache.

imo even if it's not always got an equals override, the backing resolver is the only sensible field to check for equality

text-minimessage/src/main/java/net/kyori/adventure/text/minimessage/tag/resolver/CachingTagResolver.java
```diff
@@ -78,4 +79,22 @@ public void handle(final @NotNull Component serializable, final @NotNull ClaimCo
...
+  @Override
+  public boolean equals(final @Nullable Object other) {
+    if (this == other) {
+      return true;
+    }
+    if (!(other instanceof CachingTagResolver)) {
+      return false;
+    }
+    final CachingTagResolver that = (CachingTagResolver) other;
+    return Objects.equals(this.resolver, that.resolver)
```
zml2008: just check the resolver, not the cache

zml2008 (changes requested):

zml2008 (approved):

kezz (approved):
Thanks for the PR! :)

PR #757, opened 2022-04-26
text-logger-slf4j: introduce wrapper for formatted logging

### LINKED ISSUE #731: Add SLF4J wrapper for logging components
This wrapper would make it easy for platforms to provide implementations of standard SLF4J loggers that can support logging components.

There would be a SPI for platforms to provide specific serializers (such as a future ANSI serializer), as well as translating with a specific locale.

To start with, this would only support SLF4J 1.7/1.8, but upon a stable release of SLF4J 2.0 we would add support for that.

See https://github.com/PEXPlugins/PermissionsEx/blob/master/api/src/main/java/ca/stellardrift/permissionsex/logging/FormattedLogger.java for a reference to start from

zml2008:
Fixes GH-731

TODOs:

- [x] make sure location information is passed through to logger implementations properly -- we might have to call LocationAwareLogger directly?
- [x] is our handling of argument arrays correct
- [x] test implemenations -- we should have prototypes for paper and fabric
- [x] does SLF4J have a test harness? or can we mockito up something?
- [x] is it worth doing something with `ComponentMessageThrowable`s as special handling for the throwable param?
- [x] Override all String logging methods to process component arguments/throwables passed

### DIFF
--- a/bom/build.gradle.kts
+++ b/bom/build.gradle.kts
@@ -18,6 +18,7 @@ dependencies {
       "nbt",
       "serializer-configurate3",
       "serializer-configurate4",
+      "text-logger-slf4j",
       "text-minimessage",
       "text-serializer-gson",
       "text-serializer-gson-legacy-impl",
--- a/text-logger-slf4j/src/main/java/net/kyori/adventure/text/logger/slf4j/ComponentLoggerProvider.java
+++ b/text-logger-slf4j/src/main/java/net/kyori/adventure/text/logger/slf4j/ComponentLoggerProvider.java
@@ -59,7 +59,7 @@ public interface ComponentLoggerProvider {
      * @return a plain serializer
      * @since 4.11.0
      */
-    Function<Component, String> plainSerializer();
+    @NotNull Function<Component, String> plainSerializer();
 
     /**
      * Create a component logger based on one which delegates to an underlying plain {@link Logger} implementation.
--- a/text-logger-slf4j/src/main/java/net/kyori/adventure/text/logger/slf4j/ComponentLoggerProvider.java
+++ b/text-logger-slf4j/src/main/java/net/kyori/adventure/text/logger/slf4j/ComponentLoggerProvider.java
@@ -1,0 +1,76 @@
+/*
+ * This file is part of adventure, licensed under the MIT License.
+ *
+ * Copyright (c) 2017-2022 KyoriPowered
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in all
+ * copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+ * SOFTWARE.
+ */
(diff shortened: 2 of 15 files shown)


bom/build.gradle.kts
```diff
@@ -18,6 +18,7 @@ dependencies {
       "nbt",
       "serializer-configurate3",
       "serializer-configurate4",
+      "text-logger-slf4j",
```
kezz, line 21: I wonder if we perhaps want to make this `text-logger-slf4j1` in line with the configurate serializers and considering the eventual release of slf4j 2?

text-logger-slf4j/src/main/java/net/kyori/adventure/text/logger/slf4j/ComponentLoggerProvider.java
```diff
@@ -0,0 +1,77 @@
...
+  @ApiStatus.NonExtendable
+  interface LoggerHelper {
+
+    /**
+     * Create a serializer function that will translate logged output into the system default locale, and then serialize it to plain text.
+     *
+     * @return a plain serializer
+     * @since 4.11.0
+     */
+    @NotNull Function<Component, String> plainSerializer();
```
kezz, line 63: This should probably be `ComponentSerializer<Component, Component, String>` so we can use the overrides/etc on that class and make it easier for people to just throw serializers in here.

kezz, line 63: This should probably be `ComponentSerializer<Component, Component, String>` so we can use the overrides/etc on that class and make it easier for people to just throw serializers in here.

text-logger-slf4j/src/main/java/net/kyori/adventure/text/logger/slf4j/ComponentLoggerProvider.java
```diff
@@ -0,0 +1,77 @@
...
+     * Create a component logger based on one which delegates to an underlying plain {@link Logger} implementation.
+     *
+     * <p>This sort of logger requires Components to be serialized to some sort of formatted {@link String} to match the SLF4J contract.</p>
+     *
+     * @param base the base logger
+     * @param serializer the serializer to translate and format a component in a log message.
+     * @return a new logger
+     * @since 4.11.0
+     */
+    @NotNull ComponentLogger delegating(final @NotNull Logger base, final @NotNull Function<Component, String> serializer);
```
kezz, line 75: Same here re serializer

kezz, line 75: Same here re serializer

text-logger-slf4j/src/main/java/net/kyori/adventure/text/logger/slf4j/Handler.java
```diff
@@ -0,0 +1,88 @@
...
+
+    @Override
+    public @NotNull ComponentLogger logger(final @NotNull LoggerHelper helper, final @NotNull String name) {
+      final ComponentLogger initial = this.loggers.get(name);
+      if (initial != null) return initial;
+
+      final Logger backing = LoggerFactory.getLogger(name);
+      final ComponentLogger created = helper.delegating(backing, helper.plainSerializer());
+      final ComponentLogger existing = this.loggers.putIfAbsent(name, created);
+      return existing == null ? created : existing;
```
kezz, line 63: Can we not simplify this with a computeIfAbsent call? Would also avoid the very rare case in which another call to this method inserts/creates a logger after the initial get call in this method.

kezz, line 63: Can we not simplify this with a computeIfAbsent call? Would also avoid the very rare case in which another call to this method inserts/creates a logger after the initial get call in this method.

kezz (changes requested):
Some very very minor comments here - overall this is some super dope stuff and it's really cool :)

bom/build.gradle.kts
zml2008: Looking at the alphas for SLF4J 2, they say that the only breaking changes are for logger discovery. For us the change is purely additive -- it'd be adding a layer over the fluent logging system.

text-logger-slf4j/src/main/java/net/kyori/adventure/text/logger/slf4j/ComponentLoggerProvider.java
zml2008: Currently, the plain serializer function here doesn't just serialize, it handles translating through `GlobalTranslator` as well -- which i would like to leave up to the platform. Function seemed the best choice for serializer + some other pre-processing

text-logger-slf4j/src/main/java/net/kyori/adventure/text/logger/slf4j/Handler.java
zml2008: I think what I wanted to do here was avoid the unconditional instantiation of a lambda instance, since we'd have to capture local variables in the creator function

kezz: Seems ridiculously minor but I guess that makes sense

kezz (approved):

PR #760, opened 2022-04-30
Add Context#preProcess to resolve all preProcess tags

Joo200:
This PR adds a new method to Context. With this method you can resolve preProcess tags in your own TagResolver.

This should be handy for TagResolvers which do not directly insert components and want to have some sort of recursive resolvers.

I added a recursion limit to throw an exception once the TagResolvers tried 10 times to preProcess those tags.

### DIFF
--- a/text-minimessage/src/main/java/net/kyori/adventure/text/minimessage/Context.java
+++ b/text-minimessage/src/main/java/net/kyori/adventure/text/minimessage/Context.java
@@ -70,6 +70,15 @@ public interface Context {
   @NotNull Component deserialize(final @NotNull String message, final @NotNull TagResolver@NotNull... resolvers);
 
   /**
+   * Resolve PreProcess tags and insert them in the String. This has a recursion limit of 10.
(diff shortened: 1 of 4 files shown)


zml2008:
I'm sort of reluctant to expand the API around pre-process tags -- it's one more thing to deprecate if we end up finding a better alternative to those tags.

The alternative currently would be deserializing a string as a `Component` and then serializing it using the plain text serializer, which would have the bonus of supporting *all* tag types. Thoughts?

kezz:
I agree with zml here, I don't want us to introduce a "two-tier" system of tags in MiniMessage, as that was one of the major points we moved away from when creating the release version of MiniMessage. I think the use cases are perhaps too narrow, or at least I can't think of any that wouldn't be solved by plain serializing a component argument. We could even introduce something to arguments that does this for us, even supplying a plain serializer instance in the builder. 

Thanks for the PR, but I'm going to go ahead and close this for now.

PR #764, opened 2022-05-14
Improve compaction of empty/blank TextComponents

Moulberry:
## Improve compaction of empty/blank TextComponents

I believe that for a TextComponent comprised of only space characters (" "), TextColor and ITALIC stylings are redundant.
We can take advantage of this in order to further compact components.

Font and BOLD stylings also seem like candidates for removal, but afaik BOLD affects the advance of space character and in 1.19 Font will do so too.

### Removal of useless styling on blank components
Input: `{"text":"   ","color":"red"}`
Before: `{"text":"   ","color":"red"}`
After: `{"text":"   "}`

### Compaction of blank components (side effect of previous)
Input: `{"text":"Hello ","extra":[{"color":"red","text":"   "}]}`
Before: `{"text":"Hello ","extra":[{"color":"red","text":"   "}]}`
After: `{"text":"Hello    "}`

## #Removal of empty children, regardless of style
Input: `{"text":"Hello ","color":"red","extra":[{"strikethrough":true,"text":""}]}`
Before: `{"text":"Hello ","color":"red","extra":[{"strikethrough":true,"text":""}]}`
After: `{"text":"Hello ","color":"red"}`

### DIFF
--- a/api/src/main/java/net/kyori/adventure/text/ComponentCompaction.java
+++ b/api/src/main/java/net/kyori/adventure/text/ComponentCompaction.java
@@ -221,6 +221,7 @@ final class ComponentCompaction {
     
     builder.color(null);
     builder.decoration(TextDecoration.ITALIC, TextDecoration.State.NOT_SET);
+    builder.decoration(TextDecoration.OBFUSCATED, TextDecoration.State.NOT_SET);
     
     return builder.build();
   }
--- a/api/src/main/java/net/kyori/adventure/text/ComponentCompaction.java
+++ b/api/src/main/java/net/kyori/adventure/text/ComponentCompaction.java
@@ -222,7 +222,7 @@ final class ComponentCompaction {
     // TextColor doesn't affect spaces
     builder.color(null);
 
-    // ITALIC/OBFUSCATED don't affect spaces, as these styles only affect glyph rendering
+    // ITALIC/OBFUSCATED don't affect spaces (in modern versions), as these styles only affect glyph rendering
     builder.decoration(TextDecoration.ITALIC, TextDecoration.State.NOT_SET);
     builder.decoration(TextDecoration.OBFUSCATED, TextDecoration.State.NOT_SET);
 
--- a/api/src/main/java/net/kyori/adventure/text/ComponentCompaction.java
+++ b/api/src/main/java/net/kyori/adventure/text/ComponentCompaction.java
@@ -48,6 +48,11 @@ final class ComponentCompaction {
     final int childrenSize = children.size();
 
     if (childrenSize == 0) {
+      // no children, style can be further simplified if self is blank
+      if (isBlank(self)) {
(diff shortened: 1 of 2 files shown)


api/src/main/java/net/kyori/adventure/text/ComponentCompaction.java
```diff
@@ -166,6 +190,41 @@ static Component compact(final @NotNull Component self, final @Nullable Style pa
...
+  * Checks whether the Component is blank (a TextComponent containing only space characters)
+  *
+  * @param component the component to check
+  * @return true if the provided component is blank, false otherwise
+  */
+  private static boolean isBlank(Component component) {
+    if (component instanceof TextComponent) {
+      final TextComponent textComponent = (TextComponent) component;
+      
+      for (char c : textComponent.content().toCharArray()) {
```
zml2008: can you do this using indexes and `charAt`, to avoid allocating the character array?

api/src/main/java/net/kyori/adventure/text/ComponentCompaction.java
```diff
@@ -166,6 +190,41 @@ static Component compact(final @NotNull Component self, final @Nullable Style pa
...
+  * given that the content is blank
+  *
+  * @param style style to simplify
+  * @return a new, simplified style
+  */
+  private static @NotNull Style simplifyStyleForBlank(final @NotNull Style style) {
+    final Style.Builder builder = style.toBuilder();
+    
+    builder.color(null);
+    builder.decoration(TextDecoration.ITALIC, TextDecoration.State.NOT_SET);
```
zml2008, line 235: only italics here?

zml2008 (changes requested):
Could you add some tests for this as well?

rebase needed as well, but that can happen with other changes

api/src/main/java/net/kyori/adventure/text/ComponentCompaction.java
Moulberry: Sorry, I was writing the description. I believe italics are the only redundant styling. If you have others please let me know

zml2008: Ah hm, thinking about it more, makes sense. Could you add a comment in the source here for the other decorations + style attributes and why they're not needed, for the benefit of those reading the code later on?

Moulberry: See 1fcc46c

api/src/main/java/net/kyori/adventure/text/ComponentCompaction.java
Moulberry: See 66d2ac3

Moulberry:
Last commit was supposed to say: Use `optimized` instead of `self`
It also fixed checkstyle errors

zml2008 (approved):

PR #766, opened 2022-05-15
Switched from if to switch case.

Plugrol:
Switch statements often perform faster than if-else constructs. Since the possible values of a switch statement are laid out beforehand, compilers are able to optimize performance by constructing jump tables. Each condition doesn't have to be tested as in an if/else construct


zml2008:
No -- we're leaving these until we can use switch expressions.

PR #770, opened 2022-05-31
Implement seed field on sound, and add a builder

### LINKED ISSUE #754: 1.19: Sound seeds

zml2008:
Fixes GH-754

### DIFF
--- a/api/src/main/java/net/kyori/adventure/sound/Sound.java
+++ b/api/src/main/java/net/kyori/adventure/sound/Sound.java
@@ -23,13 +23,17 @@
  */
 package net.kyori.adventure.sound;
 
+import java.util.OptionalLong;
+import java.util.function.Consumer;
 import java.util.function.Supplier;
+import net.kyori.adventure.builder.AbstractBuilder;
 import net.kyori.adventure.key.Key;
 import net.kyori.adventure.key.Keyed;
 import net.kyori.adventure.util.Index;
 import net.kyori.examination.Examinable;
 import org.jetbrains.annotations.ApiStatus;
 import org.jetbrains.annotations.NotNull;
+import org.jetbrains.annotations.Range;
 
 import static java.util.Objects.requireNonNull;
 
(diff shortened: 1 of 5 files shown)


rymiel (approved):

api/src/main/java/net/kyori/adventure/sound/Sound.java
```diff
@@ -274,4 +303,112 @@ interface Emitter {
...
   }
+
+  /**
+   * A builder for sound instances.
+   *
+   * <p>Type is required, all other options are optional.</p>
+   *
+   * @since 4.12.0
+   */
+  interface Builder extends AbstractBuilder<Sound> {
```
kezz, line 317: Do we not normally instead extend Buildable/Builder rather than abstract builder?

kezz (changes requested):

api/src/main/java/net/kyori/adventure/sound/Sound.java
zml2008: AbstractBuilder is a new interface replacing `Builder` -- me and @kashike are discussing a new pattern for builders that replaces `Built#toBuilder()` with `Built.built(existing)`, though we're unsure about its exact form.

kezz: Sounds dope to me then!

kezz (approved):

api/src/main/java/net/kyori/adventure/sound/Sound.java
```diff
@@ -274,4 +303,112 @@ interface Emitter {
...
+    /**
+     * The volume for this sound, indicating how far away it can be heard.
+     *
+     * <p>Default value is {@code 1}.</p>
+     *
+     * @param volume the sound volume
+     * @return this builder
+     * @since 4.12.0
+     */
+    @NotNull Builder volume(final @Range(from = 0, to = Integer.MAX_VALUE) float volume);
```
kezz, line 382: Thought for the future: do we want to error on values outside of these?

PR #774, opened 2022-06-02
feat(key): #773 Expose methods to check if a Key can be parsed

### LINKED ISSUE #773: Expanding the Key API by adding methods to check if a string may be parsed into a Key
Basically what the title says.

From what I've seen, the only way to check if a string (or `namespace:value` combination) is a valid Key is by:

1) Duplicating the validation logic in KeyImpl 
2) Using exception handling as a control flow mechanism, e.g.

```java 
public static boolean isValidKey(String namespace, String value)
    try {
        Key.key(namespace, string);
        return true;
    }
    catch (InvalidKeyException ignored) {
        return false;
    }
}
```
Both of these options work, but aren't particularly elegant in my opinion. Instead, it would be better if the Key interface included `isValidKey`-like methods corresponding to the `Key#key` overloads. 

This would be primarily useful for validating user input.

kashike:


key/src/main/java/net/kyori/adventure/key/Key.java
```diff
@@ -142,6 +150,79 @@ public interface Key extends Comparable<Key>, Examinable, Namespaced {
...
+  }
+
+  /**
+   * Checks if {@code value} is a valid namespace.
+   *
+   * @param namespace the string to check
+   * @return {@code true} if {@code value} is a valid namespace, {@code false} otherwise
+   * @since 4.12.0
+   */
+  @VisibleForTesting
```
zml2008:
```suggestion
```

key/src/main/java/net/kyori/adventure/key/Key.java
```diff
@@ -142,6 +150,79 @@ public interface Key extends Comparable<Key>, Examinable, Namespaced {
...
+  static boolean parseableValue(final @NotNull String value) {
+    for (int i = 0, length = value.length(); i < length; i++) {
+      if (!allowedInValue(value.charAt(i))) {
+        return false;
+      }
+    }
+    return true;
+  }
+
+  /**
```
zml2008, line 201: idk if these two methods should be API?

key/src/main/java/net/kyori/adventure/key/Key.java
```diff
@@ -142,6 +150,79 @@ public interface Key extends Comparable<Key>, Examinable, Namespaced {
...
+   * Checks if {@code value} is a valid value.
+   *
+   * @param value the string to check
+   * @return {@code true} if {@code value} is a valid value, {@code false} otherwise
+   * @since 4.12.0
+   */
+  @VisibleForTesting
```
zml2008:
```suggestion
```

zml2008 (changes requested):

key/src/main/java/net/kyori/adventure/key/Key.java
kashike: Exposed so that anyone parsing or validating a certain part of a `Key` does not have to implement the logic themselves.

zml2008 (approved):

PR #777, opened 2022-06-08
feature: 1.19 chat changes

### LINKED ISSUE #765: 1.19: Signed chat messages
This feature is still in flux, but it will involve quite a bit of work to represent within the Adventure API.

Things to think about:

- representing custom chat types -- do we expose the registry?
- representing signed messages (do we want one holder object in-api for these?)

### LINKED ISSUE #769: 1.19: Identity with name and optional team name
As supplied with signed messages

kezz:


api/src/main/java/net/kyori/adventure/identity/PlayerIdentified.java
```diff
@@ -0,0 +1,48 @@
...
+   * Gets the player identity.
+   *
+   * @return the player identity
+   * @since 4.12.0
+   */
+  @Contract(pure = true)
+  @NotNull PlayerIdentity playerIdentity();
+
+  @Override
+  default @NotNull Identity identity() {
```
kashike: Why not just change the return type on this method? I don't think the new method is necessary?

api/src/main/java/net/kyori/adventure/identity/PlayerIdentity.java
```diff
@@ -0,0 +1,94 @@
...
+import org.jetbrains.annotations.NotNull;
+import org.jetbrains.annotations.Nullable;
+
+/**
+ * An identity for a specific player.
+ *
+ * @since 4.12.0
+ */
+public interface PlayerIdentity extends Examinable, Identity {
+  // todo do we want identity/identified overloads here?
```
KingOfSquares: Cant see any proper use case for this? If some class like a custom player implements `Identified` there is no reason for it not just implementing `PlayerIdentified`.

Only use case I can think of is if there was anything that could only be identified as a player at certain points in time.

api/src/main/java/net/kyori/adventure/identity/PlayerIdentityImpl.java
```diff
@@ -0,0 +1,79 @@
...
+    if (this == other) return true;
+    if (!(other instanceof PlayerIdentity)) return false;
+    final PlayerIdentity that = (PlayerIdentity) other;
+    return this.uuid == that.uuid() && this.name == that.name() && this.teamName == that.teamName();
+  }
+
+  @Override
+  public int hashCode() {
+    int result = this.uuid.hashCode();
+    result = 31 * result + this.name.hashCode();
```
broccolai: tiny nitpick but to be consistent 
```suggestion
    result = (31 * result) + this.name.hashCode();
```

broccolai (approved):
LGTM. Apprehensive about adding empty default methods to Audience but after speaking to kezz, I don't think there's anything else we could do

api/src/main/java/net/kyori/adventure/audience/ForwardingAudience.java
```diff
@@ -96,12 +100,27 @@ default void forEachAudience(final @NotNull Consumer<? super Audience> action) {
...
+    for (final Audience audience : this.audiences()) audience.sendMessage(message, signedMessage, source, chatType);
+  }
+
+  @Override
+  default void sendMessage(final @NotNull Component message, final @NotNull SignedMessage signedMessage, final @NotNull PlayerIdentity source, final @NotNull ChatType chatType) {
+    for (final Audience audience : this.audiences()) audience.sendMessage(message, signedMessage, source, chatType);
+  }
+
+  @Override
+  default void sendMessage(@NotNull Identified source, @NotNull Component message, @NotNull MessageType type) {
```
zml2008:
```suggestion
  default void sendMessage(final @NotNull Identified source, final @NotNull Component message, final @NotNull MessageType type) {
```

api/src/main/java/net/kyori/adventure/audience/ForwardingAudience.java
```diff
@@ -96,12 +100,27 @@ default void forEachAudience(final @NotNull Consumer<? super Audience> action) {
...
     for (final Audience audience : this.audiences()) audience.sendMessage(source, message, type);
   }
 
   @Override
-  default void sendMessage(final @NotNull Identity source, final @NotNull Component message, final @NotNull MessageType type) {
+  default void sendMessage(@NotNull Identity source, @NotNull Component message, @NotNull MessageType type) {
```
zml2008:
```suggestion
  default void sendMessage(final @NotNull Identity source, final @NotNull Component message, final @NotNull MessageType type) {
```

api/src/main/java/net/kyori/adventure/chat/ChatType.java
```diff
@@ -0,0 +1,112 @@
...
+import net.kyori.examination.ExaminableProperty;
+import org.jetbrains.annotations.NotNull;
+
+/**
+ * A type of chat.
+ *
+ * @since 4.12.0
+ * @sinceMinecraft 1.19
+ */
+public interface ChatType extends Examinable, Keyed {
```
zml2008, line 45: `ChatType`s should not be `Keyed`, there is no one-to-one relationship between Key and ChatType, since registries are dynamic -- a client may have many different chat types over one jvm lifetime

api/src/main/java/net/kyori/adventure/chat/SignedMessage.java
```diff
@@ -0,0 +1,69 @@
...
+import org.jetbrains.annotations.NotNull;
+
+/**
+ * A signed chat message.
+ *
+ * @since 4.12.0
+ * @sinceMinecraft 1.19
+ */
+@ApiStatus.NonExtendable
+public interface SignedMessage extends PlayerIdentified {
```
zml2008: This isn't really a signed message, since it doesn't contain the message.

I think this would be nicest if the message was incorporated into this, and then `SignedMessage` could implement `ComponentLike`

Alternatively, this could be renamed something like `MessageSignature`

api/src/main/java/net/kyori/adventure/identity/PlayerIdentity.java
```diff
@@ -0,0 +1,103 @@
...
+
+  /**
+   * Creates a new player identity.
+   *
+   * @param uuid the uuid
+   * @param name the name
+   * @return the player identity
+   * @since 4.12.0
+   */
+  static @NotNull PlayerIdentity playerIdentity(final @NotNull UUID uuid, final @NotNull Component name) {
```
zml2008: Maybe add in an additional factory method, `PlayerIdentity.playerIdentity(Pointered)` here?

api/src/main/java/net/kyori/adventure/identity/PlayerIdentified.java
```diff
@@ -0,0 +1,42 @@
...
+package net.kyori.adventure.identity;
+
+import org.jetbrains.annotations.NotNull;
+
+/**
+ * Something that can be identified with a {@link PlayerIdentity}.
+ *
+ * @since 4.12.0
+ */
+public interface PlayerIdentified extends Identified {
```
zml2008: Maybe NamedIdentified or DisplayIdentified? this doesn't strictly have to be a player.

Alternatively, Identified.Named/Display/Player since this is a specialization of Identified

api/src/main/java/net/kyori/adventure/identity/PlayerIdentity.java
```diff
@@ -0,0 +1,103 @@
...
+import org.jetbrains.annotations.Contract;
+import org.jetbrains.annotations.NotNull;
+import org.jetbrains.annotations.Nullable;
+
+/**
+ * An identity for a specific player.
+ *
+ * @since 4.12.0
+ */
+public interface PlayerIdentity extends Examinable, Identity {
```
zml2008: Same naming opinions apply to this as to PlayerIdentified above

api/src/main/java/net/kyori/adventure/audience/Audience.java
```diff
@@ -225,118 +234,265 @@ default void sendMessage(final @NotNull Identity source, final @NotNull Componen
...
+  /**
+   * Sends a signed player chat message from the given {@link PlayerIdentified} to this {@link Audience} with the {@link ChatType#CHAT chat} chat type.
+   *
+   * @param message the message
+   * @param signedMessage the signed message data
+   * @param source the source of the message
+   * @since 4.12.0
+   */
+  @ForwardingAudienceOverrideNotRequired
+  default void sendMessage(final @NotNull ComponentLike message, final @NotNull SignedMessage signedMessage, final @NotNull PlayerIdentified source) {
```
zml2008: If the `SignedMessage` is a `PlayerIdentified`, why is there another `PlayerIdentified` parameter here?

If the `SignedMessage` can only be used with one specific combination of message + source, imo it makes most sense to store those all in one object, and only expose the ChatType as an alternate option here.

I don't like that the system and chat message methods have the same name, but I think it'd be tough to justify the API churn needed for a change, especially on the side of system messages.

api/src/main/java/net/kyori/adventure/chat/ChatType.java
```diff
@@ -0,0 +1,112 @@
...
+  /**
+   * A chat message from a player.
+   *
+   * @since 4.12.0
+   * @sinceMinecraft 1.19
+   */
+  ChatType CHAT = new ChatTypeImpl(Key.key("chat"));
```
zml2008, line 52:
```suggestion
  Keyed CHAT = Key.key("chat"));
```

and so on (after #779)

and have `ChatType` be an empty interface for now -- I fully expect things to change for 1.19.2/whatever mojang calls it, so I'd like to avoid stating too much about this api structure as-is

api/src/main/java/net/kyori/adventure/audience/Audience.java
```diff
@@ -225,118 +234,265 @@ default void sendMessage(final @NotNull Identity source, final @NotNull Componen
...
 
+  /**
+   * Sends a system chat message to this {@link Audience} with the provided {@link ChatType}.
+   *
+   * @param message the message
+   * @param chatType the chat type
+   * @since 4.12.0
+   */
+  @ForwardingAudienceOverrideNotRequired
+  default void sendMessage(final @NotNull ComponentLike message, final @NotNull ChatType chatType) {
```
zml2008:
```suggestion
  default void sendMessage(final @NotNull ComponentLike message, final @NotNull Keyed chatType) {
```

api/src/main/java/net/kyori/adventure/audience/Audience.java
```diff
@@ -225,118 +234,265 @@ default void sendMessage(final @NotNull Identity source, final @NotNull Componen
...
+  }
+
+  /**
+   * Sends a system chat message to this {@link Audience} with the provided {@link ChatType}.
+   *
+   * @param message the message
+   * @param chatType the chat type
+   * @since 4.12.0
+   */
+  default void sendMessage(final @NotNull Component message, final @NotNull ChatType chatType) {
```
zml2008:
```suggestion
  default void sendMessage(final @NotNull Component message, final @NotNull Keyed chatType) {
```

etc...

zml2008 (dismissed):

api/src/main/java/net/kyori/adventure/audience/Audience.java
```diff
@@ -225,118 +234,265 @@ default void sendMessage(final @NotNull Identity source, final @NotNull Componen
...
+   * @param source the source of the message
+   * @since 4.12.0
+   */
+  @ForwardingAudienceOverrideNotRequired
+  default void sendMessage(final @NotNull Component message, final @NotNull SignedMessage signedMessage, final @NotNull PlayerIdentified source) {
+    this.sendMessage(message, signedMessage, source, ChatType.CHAT);
+  }
+
+  /**
+   * Sends a signed player chat message from player identified by the provided {@link PlayerIdentity} to this {@link Audience} with the {@link ChatType#CHAT chat} chat type.
```
lynxplay:
```suggestion
   * Sends a signed player chat message from the player identified by the provided {@link PlayerIdentity} to this {@link Audience} with the {@link ChatType#CHAT chat} chat type.
```

api/src/main/java/net/kyori/adventure/audience/Audience.java
```diff
@@ -225,118 +234,265 @@ default void sendMessage(final @NotNull Identity source, final @NotNull Componen
...
    */
+  @Deprecated
   @ForwardingAudienceOverrideNotRequired
   default void sendMessage(final @NotNull Identity source, final @NotNull Component message) {
     this.sendMessage(source, message, MessageType.SYSTEM);
   }
 
   /**
-   * Sends a chat message with a {@link Identity#nil() nil} identity to this {@link Audience}.
+   * Sends a system chat message this {@link Audience} with the {@link ChatType} corresponding to the provided {@link MessageType}.
```
lynxplay:
```suggestion
   * Sends a system chat message to this {@link Audience} with the {@link ChatType} corresponding to the provided {@link MessageType}.
```

api/src/main/java/net/kyori/adventure/audience/Audience.java
```diff
@@ -175,7 +180,7 @@ default void forEachAudience(final @NotNull Consumer<? super Audience> action) {
   }
 
   /**
-   * Sends a chat message with a {@link Identity#nil() nil} identity to this {@link Audience}.
+   * Sends a system chat message with to this {@link Audience} with the {@link ChatType#SYSTEM system} chat type.
```
lynxplay:
```suggestion
   * Sends a system chat message to this {@link Audience} with the {@link ChatType#SYSTEM system} chat type.
```

api/src/main/java/net/kyori/adventure/audience/Audience.java
```diff
@@ -225,118 +234,265 @@ default void sendMessage(final @NotNull Identity source, final @NotNull Componen
...
+   * @param source the identity of the source of the message
+   * @since 4.12.0
+   */
+  @ForwardingAudienceOverrideNotRequired
+  default void sendMessage(final @NotNull ComponentLike message, final @NotNull SignedMessage signedMessage, final @NotNull PlayerIdentity source) {
+    this.sendMessage(message.asComponent(), signedMessage, source);
+  }
+
+  /**
+   * Sends a signed player chat message from player identified by the provided {@link PlayerIdentity} to this {@link Audience} with the {@link ChatType#CHAT chat} chat type.
```
lynxplay:
```suggestion
   * Sends a signed player chat message from the player identified by the provided {@link PlayerIdentity} to this {@link Audience} with the {@link ChatType#CHAT chat} chat type.
```

api/src/main/java/net/kyori/adventure/audience/Audience.java
```diff
@@ -225,118 +234,265 @@ default void sendMessage(final @NotNull Identity source, final @NotNull Componen
...
+   * @param signedMessage the signed message data
+   * @param source the source of the message
+   * @param chatType the chat type
+   * @since 4.12.0
+   */
+  default void sendMessage(final @NotNull Component message, final @NotNull SignedMessage signedMessage, final @NotNull PlayerIdentified source, final @NotNull ChatType chatType) {
+  }
+
+  /**
+   * Sends a signed player chat message from player identified by the provided {@link PlayerIdentity} to this {@link Audience} with the provided {@link ChatType}.
```
lynxplay:
```suggestion
   * Sends a signed player chat message from the player identified by the provided {@link PlayerIdentity} to this {@link Audience} with the provided {@link ChatType}.
```

api/src/main/java/net/kyori/adventure/audience/Audience.java
```diff
@@ -225,118 +234,265 @@ default void sendMessage(final @NotNull Identity source, final @NotNull Componen
...
+   * @param chatType the chat type
+   * @since 4.12.0
+   */
+  @ForwardingAudienceOverrideNotRequired
+  default void sendMessage(final @NotNull ComponentLike message, final @NotNull SignedMessage signedMessage, final @NotNull PlayerIdentity source, final @NotNull ChatType chatType) {
+    this.sendMessage(message.asComponent(), signedMessage, source, chatType);
+  }
+
+  /**
+   * Sends a signed player chat message from player identified by the provided {@link PlayerIdentity} to this {@link Audience} with the provided {@link ChatType}.
```
lynxplay:
```suggestion
   * Sends a signed player chat message from the player identified by the provided {@link PlayerIdentity} to this {@link Audience} with the provided {@link ChatType}.
```

api/src/main/java/net/kyori/adventure/chat/SignedMessage.java
lynxplay: I agree with this, during my first read through, a `sendMessage(Component, SignedMessage)` is rather confusing as it sounds like we are sending two messages.

api/src/main/java/net/kyori/adventure/identity/PlayerIdentity.java
kezz: Need to put a bit of thought into this: do we want a pointer for component name vs. string name? How do we handle missing pointers? Should the method throw an exception or return an Optional<PlayerIdentity>?

Machine-Maker:
Ok, this has been updated for the 1.19.1 chat changes.

There are a couple potential problems I see with some of the javadocs here. We make assumptions about what the implementations of Audience will do. I don't think we can do that.

All the deprecated Identity/Identified sendMessage methods in Audience say something like
> Sends an unsigned player chat message from the given {@link Identified} to this {@link Audience} with the {@link ChatType#CHAT system} chat type.

That could be right, or it could be wrong based on what the implementation does. It could choose to send an unsigned player chat message, or it could send a system message, or a system player message with a "raw" chat type.

Also the "with the ChatType#CHAT chat type" is maybe not right either. it all depends on the implementation.

api/src/main/java/net/kyori/adventure/chat/SignedMessage.java
```diff
@@ -0,0 +1,188 @@
...
+  @Contract(pure = true)
+  boolean canSendAsHeader();
+
+  /**
+   * A signature wrapper type.
+   *
+   * @since 4.12.0
+   * @sinceMinecraft 1.19
+   */
+  final class Signature {
```
kezz: Should be split into interface/impl.

api/src/main/java/net/kyori/adventure/chat/SignedMessageImpl.java
```diff
@@ -0,0 +1,82 @@
...
+    return this.instant;
+  }
+
+  @Override
+  public long salt() {
+    return this.salt;
+  }
+
+  @Override
+  public Signature signature() {
```
kezz, line 60: Feels really weird to me we can have a "signed message" without a signature. I wonder if we just don't support sending system messages with a signed message instance?

api/src/main/java/net/kyori/adventure/audience/MessageType.java
```diff
@@ -23,22 +23,34 @@
...
 package net.kyori.adventure.audience;
 
+import net.kyori.adventure.chat.ChatType;
+import org.jetbrains.annotations.ApiStatus;
+
 /**
  * Message types.
  *
  * @since 4.0.0
+ * @deprecated for removal since 4.12.0, use corresponding {@link ChatType} instead
```
kezz: Needs updating

api/src/main/java/net/kyori/adventure/chat/SignedMessageImpl.java
Machine-Maker: well "technically" a system signedmessage does have a signature. its just not used, because the uuid is a nil one. so the client doesnt even bother with the signature, it just sends it as a system message.

But it is important I think, because you can use chattypes with system messages this way.

api/src/main/java/net/kyori/adventure/audience/Audience.java
```diff
@@ -185,157 +191,261 @@ default void forEachAudience(final @NotNull Consumer<? super Audience> action) {
...
+   *
+   * @param signedMessage the message to delete
+   * @see SignedMessage#canDelete()
+   * @since 4.12.0
+   * @sinceMinecraft 1.19
+   */
+  @ForwardingAudienceOverrideNotRequired
+  default void deleteMessage(final @NotNull SignedMessage signedMessage) {
+    if (signedMessage.canDelete()) {
+      this.deleteMessage(Objects.requireNonNull(signedMessage.signature(), "signedMessage must have a signature"));
```
NoahvdAa: `SignedMessage#canDelete` already does a null check on the signature, so this null check is redundant. Is this intentional due to possible future changes to the canDelete method?

Machine-Maker: I *think* I just put that there to tell the IDE to stop complaining, but it was a while ago so idk for sure. Or maybe it was leftover from something.

api/src/main/java/net/kyori/adventure/chat/SignedMessageImpl.java
Machine-Maker: To follow up on this, the server can use a system signed chat message in certain situations (when dealing with some command arguments) and if they were ever included in an event on a specific platform, there should be a way to represent them with the same type.

PR #778, opened 2022-06-09
Add PlatformAPI Annotation

### LINKED ISSUE #699: Add @PlatformAPI annotation for implementation-specific API
Currently `@ApiStatus.Internal` is used. While this is not *incorrect*, more information could be conveyed with a specialized annotation.

KingOfSquares:
Proposed implementation of #699 

Closes #699

### DIFF
--- a/annotation-processors/build.gradle.kts
+++ b/annotation-processors/build.gradle.kts
@@ -3,5 +3,7 @@ plugins {
 }
 
 dependencies {
-  compileOnlyApi(libs.jetbrainsAnnotations)
+  annotationProcessor(libs.autoService.processor)
+  compileOnlyApi(libs.autoService.annotations)
+  api(libs.jetbrainsAnnotations)
 }
--- a/annotation-processors/build.gradle.kts
+++ b/annotation-processors/build.gradle.kts
@@ -7,3 +7,7 @@ dependencies {
   compileOnlyApi(libs.autoService.annotations)
   api(libs.jetbrainsAnnotations)
 }
+
+tasks.withType(AbstractPublishToMaven::class) {
+  isEnabled = false
+}
--- a/annotation-processors/build.gradle.kts
+++ b/annotation-processors/build.gradle.kts
@@ -1,0 +1,7 @@
+plugins {
+  id("adventure.common-conventions")
+}
+
+dependencies {
+  compileOnlyApi(libs.jetbrainsAnnotations)
+}
--- a/api/src/main/java/net/kyori/adventure/util/PlatformAPI.java
+++ b/api/src/main/java/net/kyori/adventure/util/PlatformAPI.java
@@ -23,6 +23,7 @@
  */
 package net.kyori.adventure.util;
 
+import java.lang.annotation.Documented;
 import java.lang.annotation.ElementType;
 import java.lang.annotation.Retention;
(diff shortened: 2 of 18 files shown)


api/src/main/java/net/kyori/adventure/util/PlatformAPI.java
```diff
@@ -0,0 +1,42 @@
...
+import java.lang.annotation.Target;
+import org.jetbrains.annotations.ApiStatus;
+
+/**
+ * Elements annotated with the {@code @PlatformAPI} annotation are intended for platform implementations of the Adventure api
+ * only.
+ *
+ * @since 4.12.0
+ */
+@ApiStatus.Internal
```
kezz, line 42: Does this warning carry over into usages of the `@PlatformAPI` annotation?

api/src/main/java/net/kyori/adventure/util/PlatformAPI.java
```diff
@@ -0,0 +1,42 @@
...
+@Target({ElementType.TYPE, ElementType.METHOD, ElementType.CONSTRUCTOR, ElementType.PACKAGE})
+@Retention(RetentionPolicy.RUNTIME)
+public @interface PlatformAPI {
```
kezz, line 46: Can we make this `@Documented` too?

api/src/main/java/net/kyori/adventure/util/PlatformAPI.java
```diff
@@ -0,0 +1,42 @@
...
+
+import java.lang.annotation.ElementType;
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+import org.jetbrains.annotations.ApiStatus;
+
+/**
+ * Elements annotated with the {@code @PlatformAPI} annotation are intended for platform implementations of the Adventure api
+ * only.
```
kezz:
```suggestion
 * Elements annotated with the {@link PlatformAPI} annotation are intended for platform implementations of the Adventure API
 * only and should not be used by standard developers. They are not public API and may change or be removed without warning at any time.
```

kezz (dismissed):
Just some minor comments!

api/src/main/java/net/kyori/adventure/util/PlatformAPI.java
KingOfSquares: Yes, if you try to use this annotation in a separate project it will show a warning(In IntelliJ at least). Not sure if you think this should be useable by the platforms also?

annotation-processors/build.gradle.kts
```diff
@@ -0,0 +1,9 @@
+plugins {
+  id("adventure.common-conventions")
+}
+
+dependencies {
+  annotationProcessor(libs.autoService.processor)
+  compileOnlyApi(libs.autoService.annotations)
+  api(libs.jetbrainsAnnotations)
+}
```
zml2008, line 9:
```suggestion
}

tasks.withType(AbstractPublishToMaven::class) {
  isEnabled = false
}
```

one last thing, just to make sure we don't publish this module externally

zml2008 (changes requested):

zml2008 (approved):

PR #779, opened 2022-06-09
key: Make Key implement Keyed

zml2008:
Allows unifying API exposed in some cases

### DIFF
--- a/key/src/main/java/net/kyori/adventure/key/Key.java
+++ b/key/src/main/java/net/kyori/adventure/key/Key.java
@@ -55,7 +55,7 @@ import org.jetbrains.annotations.NotNull;
  *
  * @since 4.0.0
  */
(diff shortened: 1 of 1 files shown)


kezz (approved):
Assuming this doesn't have any annoyances when developing (will it make `font(null)` require casting?) I'm all happy!

zml2008:
`font()` on StyleSetter seems to only support `Key` as a parameter, not `Keyed`. We could maybe transition over, but it would have to be an overload added in the last release before 5.0.0

(though: i wonder if it would be possible to work around this by marking certain methods as synthetic... hmmm)

kezz:
> `font()` on StyleSetter seems to only support `Key` as a parameter, not `Keyed`.

Honestly this is the best scenario - we can swap in 5.x, will only require a recompile for devs.

PR #783, opened 2022-06-11
Expose WrappingComponentLoggerImpl through ComponentLogger

JRoy:
there's this perfectly good logger impl here and im just tryin to use it


zml2008:
what are you trying to do? this is really for use by platforms, not by end users?

JRoy:
> what are you trying to do? this is really for use by platforms, not by end users?

Want to use this for Paper (where we don't need to change anything in WrappingComponentLoggerImpl other than the underlying serializer), kash told me to make a static method for it

zml2008:
The point of the service provider is so that *any* logger created uses the correct serializer, not just some magical 'blessed' loggers you choose. Why does the service provider not work?

JRoy:
discussed internally that I'm a dumbass

PR #793, opened 2022-06-30
api: Style#unmerge

kashike:


api/src/main/java/net/kyori/adventure/text/format/Style.java
```diff
@@ -529,6 +529,15 @@ default boolean hasDecoration(final @NotNull TextDecoration decoration) {
...
   @NotNull Style merge(final @NotNull Style that, final Merge.@NotNull Strategy strategy, final @NotNull Set<Merge> merges);
 
+  /**
+   * Simplify this style to remove any information that is redundant.
+   *
+   * @param that parent to compare against
+   * @return a new, simplified style
+   * @since 4.12.0
+   */
+  @NotNull Style asChildOf(final @NotNull Style that);
```
zml2008: dunno how I feel about this name -- maybe something like `minus`, `without`, or `unmerge`? to reflect that this returns a style without the information from the 'parent' style

zml2008 (changes requested):

zml2008 (approved):

PR #806, opened 2022-08-02
1.19.1 chat changes

Machine-Maker:
Updating for changes made in 1.19.1

There is no more "SYSTEM" chat type. If a message is a system message, it uses its own packet, and is just a raw component message.

SignedMessage should not extend PlayerIdentified, just Identified. The only player-related stuff in a SignedMessage is the UUID of the signer, so just Identified works.

Its possible the PlayerIdentified/PlayerIdentity interfaces should just be removed. I opted to kind of mirror how vanilla handles chat types, creating `ChatType$Bound` which holds the chat type, the sender component, and the nullable target component.

I don't know the exact policy on removing implementation of `Audience#sendMessage(Component)`. That is now for only system messages, but since there is no system chat type, it can't call another method without using deprecated types (`MessageType`).

I updated the default `ChatType`s that exist in vanilla 1.19.1.

I created an `UnsignedSignedMessageImpl` type so the api can create unsigned messages. All other `SignedMessage` impulse should come from platforms that use the adventure api.

### DIFF
--- a/api/src/main/java/net/kyori/adventure/chat/ChatType.java
+++ b/api/src/main/java/net/kyori/adventure/chat/ChatType.java
@@ -125,7 +125,7 @@ public interface ChatType extends Examinable, Keyed {
   interface Bound extends Examinable {
 
     @Contract(pure = true)
-    @NotNull ChatType chatType();
+    @NotNull ChatType type();
 
     @Contract(pure = true)
     @NotNull Component sender();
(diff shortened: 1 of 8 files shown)


api/src/main/java/net/kyori/adventure/chat/ChatType.java
```diff
@@ -105,8 +108,39 @@ public interface ChatType extends Examinable, Keyed {
...
+
   @Override
   default @NotNull Stream<? extends ExaminableProperty> examinableProperties() {
     return Stream.of(ExaminableProperty.of("key", this.key()));
   }
+
+  interface Bound extends Examinable {
+
+    @Contract(pure = true)
+    @NotNull ChatType chatType();
```
kashike:
```suggestion
    @NotNull ChatType type();
```

PR #819, opened 2022-09-13
minimessage: Parsing corner cases with quotes

rymiel:
Don't enter string parsing mode during the first pass when encountering a quote, but no matching quote exists later on in the message.

Note that this doesn't catch every case of invalid tags causing other valid tags to be ignored, in this case there could simply be an unrelated quote elsewhere in the message and the bug would still occur.

The parser could:

  * Only allow quotes after specific characters, such as the colon `:`, which is currently the only place where quotes are really valid. However, even then, the buggy payload could just include a colon before the quote

  * Only allow quotes when inside of a valid tag, as in the test. (`<3` isn't a valid introducer for a tag). However, even then, the buggy payload could use a character valid for tag. Additionally, checking the validity of a tag isn't done by the first parse stage

  * Backtrack when a quote-delimited section doesn't end up being inside of a valid tag. That would (presumably) require some pretty complicated checks in the first phase, but checking for it in the second phase is probably too late, because it'd have to reparse something it has already passed by

### DIFF
--- a/text-minimessage/src/main/java/net/kyori/adventure/text/minimessage/internal/parser/TokenParser.java
+++ b/text-minimessage/src/main/java/net/kyori/adventure/text/minimessage/internal/parser/TokenParser.java
@@ -248,7 +248,7 @@ public final class TokenParser {
             case '"':
               currentStringChar = (char) codePoint;
               // Look ahead if the quote being opened is ever closed
-              if (message.substring(i + 1).indexOf(codePoint) != -1) {
+              if (message.indexOf(codePoint, i + 1) != -1) {
                 state = FirstPassState.STRING;
               }
               break;
--- a/text-minimessage/src/main/java/net/kyori/adventure/text/minimessage/internal/parser/TokenParser.java
+++ b/text-minimessage/src/main/java/net/kyori/adventure/text/minimessage/internal/parser/TokenParser.java
@@ -246,8 +246,11 @@ public final class TokenParser {
               break;
             case '\'':
             case '"':
-              state = FirstPassState.STRING;
               currentStringChar = (char) codePoint;
+              // Look ahead if the quote being opened is ever closed
+              if (message.substring(i + 1).indexOf(codePoint) != -1) {
+                state = FirstPassState.STRING;
+              }
               break;
           }
           break;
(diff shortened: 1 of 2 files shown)


MiniDigger:
I think this is a good fix for an edge case, which doesn't fix all edge cases, but is a good enough improvement as is.

rymiel:
I've added another edge case fix:

A tag is opened by `<` and closed by `>`. Upon encountering an opening
angle bracket `<`, the parser enters the "TAG" state, during which it'll
parse quote-delimited sequences. However, if by the end of the input,
a closing angle bracket `>` is never encountered, the parser simply
adds everything as a single text node, including the contents of the
quote-delimited sequences.

These quote delimited sequences could have contained other valid tags
that are now directly added to the output without being parsed as tags.

With this change, when the parser finds itself at the end of the input
in the TAG state, it will backtrack back to the beginning of the `<`
that caused a state change, and parse again from there on, but this time
*not* in the tag state. This means that quotes are effectively ignored
and their contents parsed as normal text

This *still* doesn't account for all corner case, for example:

`<red>foo <tag:"</red>"> bar`. This is an example of a *correct* use of
quotes, and the expected behaviour is for the whole output to be red,
and the actual behaviour matches that, however:

`<red>foo <3:"</red>"> bar`. This is nearly identical to the above, but
<3> isn't a valid tag, which means the red tag should be closed in the
middle of the message. Instead, the message ends up wholly red.

`<red>foo <tag"</red>"> bar`. The same applies here, but instead of an
invalid tag, it is invalid syntax, since quotes should follow a colon.

text-minimessage/src/main/java/net/kyori/adventure/text/minimessage/internal/parser/TokenParser.java
```diff
@@ -246,8 +246,11 @@ public static void parseString(final String message, final MatchedTokenConsumer<
               break;
             case '\'':
             case '"':
-              state = FirstPassState.STRING;
               currentStringChar = (char) codePoint;
+              // Look ahead if the quote being opened is ever closed
+              if (message.substring(i + 1).indexOf(codePoint) != -1) {
```
zml2008:
```suggestion
              if (message.indexOf(codePoint, i + 1) != -1) {
```

just to save that allocation :p

zml2008 (changes requested):
one tiny nitpick, but that and a rebase should make this ready to merge

rymiel:
rebased

zml2008 (approved):

PR #820, opened 2022-09-18
Add methods to Audience allowing access to previously shown BossBars

WasabiThumb:
See [this PR](https://github.com/KyoriPowered/adventure-platform/pull/105) for how it may be implemented in adventure-platform.


api/src/main/java/net/kyori/adventure/audience/Audience.java
```diff
@@ -608,6 +611,25 @@ default void showBossBar(final @NotNull BossBar bar) {
...
+  default boolean hasBossBar(final @NotNull BossBar bar) {
+    return false;
+  }
+
+  /**
+   * Returns a set containing all BossBars this Audience has been shown
+   * @return An unmodifiable set of BossBars
+   * @since 4.12.0
+   */
+  default @NotNull @UnmodifiableView Set<BossBar> getBossBars() {
```
zml2008, line 629: 1. This does not follow Kyori's organizational naming standards -- you shouldn't have `get` prefixes.
2. We can't really get *all* boss bars here, especially if we're in a proxy environment.

zml2008 (changes requested):

zml2008:
Superceded by #894

PR #832, opened 2022-11-07
fix(text-minimessage): Properly handle escaped tag opens

### LINKED ISSUE #821: escapeTags doesnt works correctly
I'm using `miniMessage.deserialize(miniMessage.escapeTags(playerChatEvent.getMessage()))` but when i use `<<aqua> a` it will escape tags but not correctly 
image

WIth `<agua> a` it works fine:
image

zml2008:
Fixes GH-821

### DIFF
--- a/text-minimessage/src/main/java/net/kyori/adventure/text/minimessage/internal/parser/Token.java
+++ b/text-minimessage/src/main/java/net/kyori/adventure/text/minimessage/internal/parser/Token.java
@@ -24,6 +24,7 @@
 package net.kyori.adventure.text.minimessage.internal.parser;
 
 import java.util.List;
+import java.util.Objects;
 import java.util.stream.Stream;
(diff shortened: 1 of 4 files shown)


text-minimessage/src/main/java/net/kyori/adventure/text/minimessage/internal/parser/Token.java
```diff
@@ -126,6 +127,19 @@ public CharSequence get(final CharSequence message) {
     );
   }
 
+  @Override
+  public boolean equals(final Object other) {
```
rymiel, line 131: This is just used for tests, yes?

zml2008: yeah, so we can directly compare token output

rymiel (approved):
idk i can't find any flaws

PR #833, opened 2022-11-07
fix(text-minimessage): Be more lenient with input when stripping/escaping tags

### LINKED ISSUE #799: `MiniMessage#escapeTags` throws an exception when input contains section symbols
i want to be able to take untrusted user input which may contain legacy section symbol color codes and sanitize and tags from them before converting the section symbols to tags (because vise versa would cause the converted tags to be escaped)

zml2008:
Fixes GH-799


rymiel (approved):
really hope this doesn't end up as API

PR #842, opened 2022-12-01
feat(api): virtual components

### LINKED ISSUE #872: Provide a way to retain tags such as <rainbow> when serializing
Attempting to serialize components which have rainbow or gradient tags results in the tag getting converted to a lot of individual color tags (see below). It would be useful to keep the rainbow tags when e.g. the user wants to convert from non-strict to strict format. 

Original text:
`<rainbow>test</rainbow>`
Expected output once deserialized and serialized back:
`<rainbow>test</rainbow>`
Actual output:
`<#f3801f>t</#f3801f><#4bff2c>e</#4bff2c><#0c80e0>s</#0c80e0><#b401d3>t</#b401d3>`

kashike:
- [ ] look over component flattening with virtual components
- [ ] figure out how to do a builder that's useful for outside people

### DIFF
--- a/api/src/main/java/net/kyori/adventure/text/Component.java
+++ b/api/src/main/java/net/kyori/adventure/text/Component.java
@@ -1276,6 +1276,48 @@ public interface Component extends ComponentBuilderApplicable, ComponentLike, Ex
     return VirtualComponentImpl.createVirtual(virtual);
   }
 
+  /**
+   * Creates a virtual component with a value.
+   *
+   * @param virtual the value
+   * @param style the style
+   * @return a virtual component
+   * @since 4.13.0
+   */
+  @Contract(value = "_, _ -> new", pure = true)
+  static @NotNull VirtualComponent virtual(final @NotNull VirtualComponentHolder<?> virtual, final @NotNull Style style) {
+    requireNonNull(virtual, "virtual");
+    return VirtualComponentImpl.createVirtual(virtual);
+  }
+
+  /**
+   * Creates a virtual component with a value.
+   *
+   * @param virtual the value
+   * @param style the style elements
+   * @return a virtual component
+   * @since 4.13.0
+   */
+  @Contract(value = "_, _ -> new", pure = true)
+  static @NotNull VirtualComponent virtual(final @NotNull VirtualComponentHolder<?> virtual, final @NotNull StyleBuilderApplicable... style) {
+    requireNonNull(virtual, "virtual");
+    return VirtualComponentImpl.createVirtual(Collections.emptyList(), Style.style(style), virtual);
+  }
+
+  /**
+   * Creates a virtual component with a value.
+   *
+   * @param virtual the value
+   * @param style the style elements
+   * @return a virtual component
+   * @since 4.13.0
+   */
+  @Contract(value = "_, _ -> new", pure = true)
+  static @NotNull VirtualComponent virtual(final @NotNull VirtualComponentHolder<?> virtual, final @NotNull Iterable<StyleBuilderApplicable> style) {
+    requireNonNull(virtual, "virtual");
+    return VirtualComponentImpl.createVirtual(Collections.emptyList(), Style.style(style), virtual);
+  }
+
   /*
    * -------------------------------
    * ---- TranslatableComponent ----
--- a/api/src/main/java/net/kyori/adventure/text/Component.java
+++ b/api/src/main/java/net/kyori/adventure/text/Component.java
@@ -1264,58 +1264,70 @@ public interface Component extends ComponentBuilderApplicable, ComponentLike, Ex
    */
 
   /**
-   * Creates a virtual component with a value.
+   * Creates a virtual component.
    *
-   * @param virtual the value
+   * @param <C> the context type
+   * @param context the context type
+   * @param renderer the renderer
    * @return a virtual component
-   * @since 4.13.0
+   * @since 4.14.0
    */
-  @Contract(value = "_ -> new", pure = true)
-  static @NotNull VirtualComponent virtual(final @NotNull VirtualComponentHolder<?> virtual) {
-    requireNonNull(virtual, "virtual");
-    return VirtualComponentImpl.createVirtual(virtual);
+  @Contract(value = "_, _ -> new", pure = true)
+  static <C> @NotNull VirtualComponent virtual(final @NotNull Class<C> context, final @NotNull VirtualComponentRenderer<C> renderer) {
+    requireNonNull(context, "context");
+    requireNonNull(renderer, "renderer");
+    return VirtualComponentImpl.createVirtual(context, renderer);
   }
 
   /**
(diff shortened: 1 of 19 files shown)


api/src/main/java/net/kyori/adventure/text/renderer/AbstractComponentRenderer.java
```diff
@@ -129,6 +133,17 @@
...
 
+  /**
+   * Renders a virtual component.
+   *
+   * @param component the component
+   * @param context the context
+   * @return the rendered component
+   */
+  protected @NotNull Component renderVirtual(final @NotNull VirtualComponent component, final @NotNull C context) {
+    return component;
```
kezz, line 145: wonder if this should return `Component.empty()` instead? just so that it will always "clean" up any virtual components

api/src/main/java/net/kyori/adventure/text/renderer/AbstractComponentRenderer.java
```diff
@@ -129,6 +133,17 @@
    */
   protected abstract @NotNull Component renderText(final @NotNull TextComponent component, final @NotNull C context);
 
+  /**
+   * Renders a virtual component.
+   *
+   * @param component the component
+   * @param context the context
+   * @return the rendered component
```
kezz, line 141:
```suggestion
   * @return the rendered component
   * @since 4.13.0
```

kezz (approved):
overall lgtm!

do we need to make any changes to serializers to make sure they throw errors if any `VirtualComponent` instances get through unrendered?

additionally this likely needs MM changes too to add this component type (unless we're happy with it just throwing errors?)

api/src/main/java/net/kyori/adventure/text/renderer/AbstractComponentRenderer.java
```diff
@@ -129,6 +133,18 @@
...
+   * @since 4.13.0
+   */
+  protected @NotNull Component renderVirtual(final @NotNull VirtualComponent component, final @NotNull C context) {
+    return Component.empty();
```
zml2008: should the default here just be a TextComponent with the same style, contents, and children as the virtual component instead?

zml2008 (changes requested):

api/src/main/java/net/kyori/adventure/text/renderer/AbstractComponentRenderer.java
zml2008: or hm, maybe even pass the component through unchanged just to be safe in multistage rendering processes?

zml2008:
@kezz right now the virtual component is treated as/turns into a text component whenever its serialized -- when we get a public custom serializer API on MM we will have to make sure user-provided serializers can claim virtual components before the built-in serializers get a say.

api/src/main/java/net/kyori/adventure/text/renderer/AbstractComponentRenderer.java
kezz: yeah - id rather we just do smth ugly like toString or whatever

api/src/main/java/net/kyori/adventure/text/Component.java
```diff
@@ -1225,6 +1225,25 @@ public interface Component extends ComponentBuilderApplicable, ComponentLike, Ex
...
+
+  /**
+   * Creates a virtual component with a value.
+   *
+   * @param virtual the value
+   * @return a virtual component
+   * @since 4.13.0
+   */
+  @Contract(value = "_ -> new", pure = true)
+  static @NotNull VirtualComponent virtual(final @NotNull VirtualComponentHolder<?> virtual) {
```
zml2008: hm, maybe worth adding a StyleBuilderApplicable... varargs overload, and a `Style` overload, as kezz mentioned for the translatable fallback PR?

zml2008 (dismissed):

zml2008:
After fiddling around with this a bit to use the API in MM, I think this will need some more iteration before it's ready to push out.

Some key things:
- equality and examinableness of virtual holders
- ensuring api operations preserve virtual components where possible
- thinking about the level of boiler plate needed to create data -- do we need the two-layer system?

Pablete1234:
> thinking about the level of boiler plate needed to create data -- do we need the two-layer system?

Would it make sense to, instead of having the 2-layer VirtualComponent -> VirtualComponentHolder<?> -> boxed data, as well as the need for a custom renderer,  have the component itself have a "virtual component renderer" instead?

A way to do that would be to make it so `VirtualComponent` internally has a `Function<C, Component>` (or functionally equivalent VirtualComponentRenderer) which is used at render time to get the actual component, replacing  the current `unbox`. Such function is part of 3rd party plugins and so they can have as much state or non-state as they want within it, and adventure doesn't really need to care about it. This essentially shifts the focus of the "virtual component holder" so instead of "holding data" it's the "how does this get rendered"

An example of a simple component which would show each player their own name:
`virtual(context -> text(context.getOrDefault(Identity.NAME, "no name :(")));`

Main issue I see (correct me if i'm wrong) is context is generic in adventure, and only in platform does it get a specific type, so one platform may use pointered and others may use something else? i have not looked into that one.


> equality and examinableness of virtual holders

I think that just goes out the window, simply because it no longer becomes adventure's territory, it's a "whatever that 3rd party plugin decided" territory. How relevant is it to adventure how internally the renderer of the 3rd party plugin decides to be? Do equal-ness checks break adventure in significant ways if they're not respected for the renderer of virtual components?

api/src/main/java/net/kyori/adventure/text/VirtualComponentImpl.java
```diff
@@ -29,36 +29,43 @@
...
   @Override
   VirtualComponent create0(final @NotNull List<? extends ComponentLike> children, final @NotNull Style style, final @NotNull String content) {
-    return new VirtualComponentImpl(ComponentLike.asComponents(children, IS_NOT_EMPTY), style, content, this.virtual);
+    return new VirtualComponentImpl(ComponentLike.asComponents(children, IS_NOT_EMPTY), style, content, this.context, this.renderer);
   }
 
   @Override
-  public @NotNull VirtualComponentHolder<?> holder() {
-    return this.virtual;
+  public @NotNull Class<?> context() {
```
zml2008: contextType?

Pablete1234:
Hey, any updates on this? It'd be great to see the functionality finally landing on adventure

FxMorin:
I would like it if you could add a builder wrapper for use with the virtual component system.
In our project we create and use custom components. Having the ability to build on top of the virtual component system would be very helpful and would help reduce the current amount of hackery we do. 

Currently the abstract builders are private (_for good reasons_). However we are willing to deal with breaking changes when needed, the virtual components however would solve this ongoing issue and prevent any more breaking changes from affecting us.

kezz (approved):
`@since` tags need updating, other than that LGTM!

PR #848, opened 2022-12-10
feat(mini-message): #791 add mini-message audience

### LINKED ISSUE #791: `MiniMessageAudience` that provides shorthand methods to send messages/etc with MiniMessage strings

AV3RG:
Adds shorthand methods to send messages with mini-message strings in them easily. The current approach utilizes the delegation provided by `ForwardingAudience` and allows users to use a custom `MiniMessage` instance for each audience instead of just the default one.

TODO:

- [x] Shorthand method for chat messages/action bars/title parts/tab list header and footer
- [ ] Unit tests for MiniMessageAudience

Closes #791

### DIFF
--- a/text-minimessage/src/main/java/net/kyori/adventure/text/minimessage/audience/MiniMessageAudience.java
+++ b/text-minimessage/src/main/java/net/kyori/adventure/text/minimessage/audience/MiniMessageAudience.java
@@ -23,9 +23,7 @@
  */
 package net.kyori.adventure.text.minimessage.audience;
 
-import java.util.Collections;
 import net.kyori.adventure.audience.Audience;
-import net.kyori.adventure.audience.ForwardingAudience;
 import net.kyori.adventure.chat.ChatType;
 import net.kyori.adventure.text.Component;
 import net.kyori.adventure.text.minimessage.MiniMessage;
@@ -34,19 +32,11 @@ import net.kyori.adventure.title.TitlePart;
 import org.jetbrains.annotations.NotNull;
 
 /**
- * Implementation of {@link ForwardingAudience} that provides additional methods for sending MiniMessage strings.
+ * Implementation of {@link Audience} that provides additional methods for sending MiniMessage strings.
  *
  * @since 4.13.0
  */
-public final class MiniMessageAudience implements ForwardingAudience {
-
-  private final @NotNull Iterable<Audience> delegate;
-  private final @NotNull MiniMessage mini;
-
-  private MiniMessageAudience(final @NotNull Audience delegate, final @NotNull MiniMessage mini) {
-    this.delegate = Collections.singleton(delegate);
-    this.mini = mini;
-  }
+public interface MiniMessageAudience extends Audience {
 
   /**
    * Creates a new {@link MiniMessageAudience} that forwards to the provided {@link Audience}.
@@ -56,8 +46,8 @@ public final class MiniMessageAudience implements ForwardingAudience {
    * @return a new MiniMessageAudience
    * @since 4.13.0
    */
-  public static @NotNull MiniMessageAudience audience(final @NotNull Audience audience) {
-    return new MiniMessageAudience(audience, MiniMessage.miniMessage());
+  static @NotNull MiniMessageAudience audience(final @NotNull Audience audience) {
+    return audience(audience, MiniMessage.miniMessage());
   }
 
   /**
@@ -68,8 +58,8 @@ public final class MiniMessageAudience implements ForwardingAudience {
    * @return a new MiniMessageAudience
    * @since 4.13.0
    */
-  public static @NotNull MiniMessageAudience audience(final @NotNull Audience audience, final @NotNull MiniMessage mini) {
-    return new MiniMessageAudience(audience, mini);
+  static @NotNull MiniMessageAudience audience(final @NotNull Audience audience, final @NotNull MiniMessage mini) {
+    return new MiniMessageAudienceImpl(audience, mini);
   }
 
   /**
@@ -78,9 +68,7 @@ public final class MiniMessageAudience implements ForwardingAudience {
    * @param message the message with MiniMessage strings to send
    * @since 4.13.0
    */
-  public void sendMessage(final @NotNull String message) {
-    this.sendMessage(this.mini.deserialize(message));
-  }
+  void sendMessage(final @NotNull String message);
 
   /**
    * Sends a MiniMessage string to this audience.
@@ -89,9 +77,7 @@ public final class MiniMessageAudience implements ForwardingAudience {
    * @param tagResolver the tag resolver to parse additional tags
    * @since 4.13.0
    */
-  public void sendMessage(final @NotNull String message, final @NotNull TagResolver tagResolver) {
-    this.sendMessage(this.mini.deserialize(message, tagResolver));
-  }
+  void sendMessage(final @NotNull String message, final @NotNull TagResolver tagResolver);
 
   /**
    * Sends a MiniMessage string to this audience.
@@ -100,9 +86,7 @@ public final class MiniMessageAudience implements ForwardingAudience {
    * @param tagResolvers the tag resolvers to parse additional tags
    * @since 4.13.0
    */
-  public void sendMessage(final @NotNull String message, final @NotNull TagResolver @NotNull... tagResolvers) {
-    this.sendMessage(this.mini.deserialize(message, tagResolvers));
-  }
+  void sendMessage(final @NotNull String message, final @NotNull TagResolver @NotNull ... tagResolvers);
 
   /**
    * Sends a MiniMessage string to this audience.
@@ -111,9 +95,7 @@ public final class MiniMessageAudience implements ForwardingAudience {
    * @param bound the bound chat type
    * @since 4.13.0
    */
-  public void sendMessage(final @NotNull String message, final ChatType.@NotNull Bound bound) {
-    this.sendMessage(this.mini.deserialize(message), bound);
-  }
+  void sendMessage(final @NotNull String message, ChatType.@NotNull Bound bound);
 
   /**
    * Sends a MiniMessage string to this audience as action bar.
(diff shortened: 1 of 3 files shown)


zml2008:
I'm not entirely sure how I feel about this yet so I'm hoping for feedback from more people.

I do think `MiniMessageAudience` should be an interface that directly extends Audience and calls through to the `Audience` methods. A default implementation could exist private to the package, so it's only exposed via `MiniMessageAudience.miniMessageAudience(Audience delegate)` or something like that, rather than exposing the whole concrete class.

AV3RG:
> I do think `MiniMessageAudience` should be an interface that directly extends Audience and calls through to the `Audience` methods. A default implementation could exist private to the package, so it's only exposed via `MiniMessageAudience.miniMessageAudience(Audience delegate)` or something like that, rather than exposing the whole concrete class.

I agree that `MiniMessageAudience` should be an interface and should not directly expose the whole class but I do still think that it should extend `ForwardingAudience` instead of `Audience`. Forwarding audience provides easy delegation for the methods that are meant to be implemented by platforms with just a single `audiences()` method.

kezz:
I agree - it should be an interface that implements `Audience`, not `ForwardingAudience`. Platforms with a player-like object shouldn't be forced to handle the overhead of all forwarding audience calls for no reason.

text-minimessage/src/main/java/net/kyori/adventure/text/minimessage/audience/MiniMessageAudience.java
```diff
@@ -0,0 +1,236 @@
...
+
+  /**
+   * Creates a new {@link MiniMessageAudience} that forwards to the provided {@link Audience}.
+   * Uses the {@link MiniMessage} instance provided by {@link MiniMessage#miniMessage()}.
+   *
+   * @param audience the audience to forward to
+   * @return a new MiniMessageAudience
+   * @since 4.13.0
+   */
+  static @NotNull MiniMessageAudience audience(final @NotNull Audience audience) {
```
kezz, line 49: should be `miniMessageAudience` so it's nice for static imports

text-minimessage/src/main/java/net/kyori/adventure/text/minimessage/audience/MiniMessageAudience.java
```diff
@@ -0,0 +1,236 @@
...
+
+  /**
+   * Creates a new {@link MiniMessageAudience} that forwards to the provided {@link Audience}.
+   *
+   * @param audience the audience to forward to
+   * @param mini the MiniMessage instance to use
+   * @return a new MiniMessageAudience
+   * @since 4.13.0
+   */
+  static @NotNull MiniMessageAudience audience(final @NotNull Audience audience, final @NotNull MiniMessage mini) {
```
kezz, line 61:
```suggestion
  static @NotNull MiniMessageAudience audience(final @NotNull Audience audience, final @NotNull MiniMessage miniMessage) {
```

text-minimessage/src/main/java/net/kyori/adventure/text/minimessage/audience/MiniMessageAudienceImpl.java
```diff
@@ -0,0 +1,219 @@
...
+import org.jetbrains.annotations.NotNull;
+
+final class MiniMessageAudienceImpl implements MiniMessageAudience {
+
+  private final @NotNull Audience delegate;
+  private final @NotNull MiniMessage mini;
+
+  MiniMessageAudienceImpl(final @NotNull Audience delegate, final @NotNull MiniMessage mini) {
+    this.delegate = delegate;
+    this.mini = mini;
```
kezz, line 48:
```suggestion
    this.delegate = Objects.requireNotNull(delegate, "delegate");
    this.mini = Objects.requireNotNull(mini, "mini");
```

text-minimessage/src/main/java/net/kyori/adventure/text/minimessage/audience/MiniMessageAudienceImpl.java
```diff
@@ -0,0 +1,219 @@
...
+import net.kyori.adventure.text.Component;
+import net.kyori.adventure.text.minimessage.MiniMessage;
+import net.kyori.adventure.text.minimessage.tag.resolver.TagResolver;
+import net.kyori.adventure.title.TitlePart;
+import org.jetbrains.annotations.NotNull;
+
+final class MiniMessageAudienceImpl implements MiniMessageAudience {
+
+  private final @NotNull Audience delegate;
+  private final @NotNull MiniMessage mini;
```
kezz, line 44: should be be `miniMessage`

kezz (changes requested):
Overall, I like the idea behind this. I can see it being a useful way to introduce people to MM/adventure for platforms.

However, I do worry about the actual use of this in practice. For example, Paper couldn't really use this as `sendMessage(String)` conflicts. I guess they could just do a check if it contains legacy and then forward it on but I wonder if we should have these methods being `sendMiniMessage`, `sendMiniMessageBook`, etc.

Would be interested in hearing any thoughts anyone else has.

imDaniX:
Maybe extend the idea a little for more general usage? An audience that is created with a serializer and audience instances, and then uses the serializer to convert and send string messages to the audience. Dunno about the name tho, maybe something like `FancyAudience` or `ConvertingAudience`.
Then make `MiniMessageAudience` extend said interface with all its `TagResolver`s thingies.

It might be worth to include such audience creation into the `Audience` interface, something like `Audience#audience(ComponentSerializer<Component, Component, String> componentSerializer)`, but it's rather questionable.

kezz:
> Maybe extend the idea a little for more general usage? An audience that is created with a serializer and audience instances, and then uses the serializer to convert and send string messages to the audience.

Hmmm, I'm not really sure I like this idea. I think it is potentially _too_ generic to be useful. I envision this `MiniMessageAudience` being a wrapper, with methods like `sendMiniMessageBook`, etc. I'm not sure what we'd even call a generic, String-parameterized `sendMessage` alternative. Plus, the docs would have to be very specific too... 

Overall, when you can just `sendMessage(thing.deserialize("hi"))` I'm not sure how much utility this has. I can see it being useful for `MiniMessage` strings, considering how common they seem to be used anyway, but yeah... Overall not sure.

imDaniX:
> Hmmm, I'm not really sure I like this idea. I think it is potentially _too_ generic to be useful. I envision this `MiniMessageAudience` being a wrapper, with methods like `sendMiniMessageBook`, etc.  I'm not sure what we'd even call a generic, String-parameterized `sendMessage` alternative. Plus, the docs would have to be very specific too...

Well, I dunno why I've mentioned only String. The concept surely can be extended by using java generics. Something `ConvertingAudience<T>`, where's the `T` stands for the type that's being converted from, and sending methods would look like `sendConvertedMessage(T message)`.

> Overall, when you can just `sendMessage(thing.deserialize("hi"))` I'm not sure how much utility this has. I can see it being useful for `MiniMessage` strings, considering how common they seem to be used anyway, but yeah... Overall not sure.

For me, MiniMessage just isn't the serializer I prefer at all, and I don't think I'm alone. It feels it would be a bit too selective to add such audience only for MiniMessage.
One can say "If they need it - they'll do it themselves", and I guess I agree. But, well, wrappers are always a pain in the arse, so if anyone would like to use something aside of MiniMessage for their project, such API would be rather helpful and will save quite a bit of time.

AV3RG:
> However, I do worry about the actual use of this in practice. For example, Paper couldn't really use this as `sendMessage(String)` conflicts. I guess they could just do a check if it contains legacy and then forward it on but I wonder if we should have these methods being `sendMiniMessage`, `sendMiniMessageBook`, etc.

I am on phone, so sorry for any incorrect quotes. Just had my last semester exams and I will get right back on this.

Yea I agree with this, I think its a simple approach that will fit well with any audience implementation.

AV3RG:
> > Hmmm, I'm not really sure I like this idea. I think it is potentially _too_ generic to be useful. I envision this `MiniMessageAudience` being a wrapper, with methods like `sendMiniMessageBook`, etc.  I'm not sure what we'd even call a generic, String-parameterized `sendMessage` alternative. Plus, the docs would have to be very specific too...
> 
> Well, I dunno why I've mentioned only String. The concept surely can be extended by using java generics. Something `ConvertingAudience<T>`, where's the `T` stands for the type that's being converted from, and sending methods would look like `sendConvertedMessage(T message)`.
> 
> > Overall, when you can just `sendMessage(thing.deserialize("hi"))` I'm not sure how much utility this has. I can see it being useful for `MiniMessage` strings, considering how common they seem to be used anyway, but yeah... Overall not sure.
> 
> For me, MiniMessage just isn't the serializer I prefer at all, and I don't think I'm alone. It feels it would be a bit too selective to add such audience only for MiniMessage.
> One can say "If they need it - they'll do it themselves", and I guess I agree. But, well, wrappers are always a pain in the arse, so if anyone would like to use something aside of MiniMessage for their project, such API would be rather helpful and will save quite a bit of time.
> 

I have another idea. How about we promote the existing methods or make new methods that accept an interface called something like `ComponentSerializable` or `Sendable` something that has a `serializeToComponent()` method that returns a Component. That will allow you to inherit that in your classes/data types and send them directly. Ofc this approach wont work with standard java classes like `String` but I can see that being helpful for a lot of other stuff.

Would love to hear opinions on this.

zml2008:
That already exists with `ComponentLike`. 

I think making this interface more generic (in supporting non-MM serializers) would result in loss of functionality without benefits that outweigh that cost -- things like custom tag resolvers are nice to have in-line.

AV3RG:
> That already exists with `ComponentLike`. 
> 
> I think making this interface more generic (in supporting non-MM serializers) would result in loss of functionality without benefits that outweigh that cost -- things like custom tag resolvers are nice to have in-line.

I did not know that class :P

I am not saying replace `MiniMessageAudience`. I am saying keep it like it is and add the `ComponentLike` functionality alongside in the `Audience` class

But then if `ComponentLike` already exists, I am not too sure of changing/adding new methods is worth it when you can just add a function call and then send it

AV3RG:
Does anyone have any opinions on what the unit tests should check?
Kinda lost on that

Leguan16:
> Does anyone have any opinions on what the unit tests should check?
> Kinda lost on that

I can take a look at it on the weekend if you like.

PR #860, opened 2023-01-14
feat(api): Add emulated callback click event

### LINKED ISSUE #803: ClickEvent.callback()
Many of our platform implementations have implementation of a system which, using a serverside function cache accessed by a command run on the client.

It would improve usability and discoverability to expose this system as part of common API.

# downsides
This callback API would need to be implemented on each platform, and could not have a sensible default implementation - most other service providers can at least function without specific platform support

# implementation

- `ClickEvent.callback(Consumer<? super Audience>)`
- A CallbackProvider service for platforms to implement
- options for single-use vs multi-user/time-based expiration callbacks?

zml2008:
This will require platform support from each platform in order to make this actually functional.

Closes #803

### DIFF
--- a/api/src/main/java/net/kyori/adventure/text/event/ClickCallback.java
+++ b/api/src/main/java/net/kyori/adventure/text/event/ClickCallback.java
@@ -31,7 +31,6 @@ import net.kyori.adventure.builder.AbstractBuilder;
 import net.kyori.adventure.permission.PermissionChecker;
 import net.kyori.adventure.text.Component;
 import net.kyori.adventure.util.PlatformAPI;
-import net.kyori.adventure.util.TriState;
 import net.kyori.examination.Examinable;
 import org.jetbrains.annotations.ApiStatus;
 import org.jetbrains.annotations.NotNull;
@@ -124,7 +123,7 @@ public interface ClickCallback<T extends Audience> {
   /**
    * Require that audiences receiving this callback have a certain permission.
    *
-   * <p>For audiences without permissions information, this test will always pass.</p>
+   * <p>For audiences without permissions information, this test will always fail.</p>
    *
    * <p>Actions from audiences that do not match this predicate will be silently ignored.</p>
    *
@@ -139,7 +138,7 @@ public interface ClickCallback<T extends Audience> {
   /**
    * Require that audiences receiving this callback have a certain permission.
    *
-   * <p>For audiences without permissions information, this test will always pass.</p>
+   * <p>For audiences without permissions information, this test will always fail.</p>
    *
    * @param permission the permission to check
    * @param failureMessage a message to send if the conditions are not met
@@ -147,7 +146,7 @@ public interface ClickCallback<T extends Audience> {
    * @since 4.13.0
    */
   default @NotNull ClickCallback<T> requiringPermission(final @NotNull String permission, final @Nullable Component failureMessage) {
-    return this.filter(audience -> audience.getOrDefault(PermissionChecker.POINTER, PermissionChecker.always(TriState.TRUE)).test(permission), failureMessage);
+    return this.filter(audience -> audience.getOrDefault(PermissionChecker.POINTER, ClickCallbackInternals.ALWAYS_FALSE).test(permission), failureMessage);
   }
 
   /**
@@ -157,6 +156,13 @@ public interface ClickCallback<T extends Audience> {
    */
   @ApiStatus.NonExtendable
   interface Options extends Examinable {
+    /**
+     * Indicate that a callback should have unlimited uses.
+     *
+     * @since 4.13.0
+     */
+    int UNLIMITED_USES = -1;
+
     /**
      * Create a new builder.
      *
@@ -179,19 +185,19 @@ public interface ClickCallback<T extends Audience> {
     }
 
     /**
-     * Whether the callback can be executed multiple times.
+     * The number of times this callback can be executed.
      *
      * <p>By default callbacks are single-use.</p>
      *
-     * @return whether this callback is multi-use
+     * @return allowable use count, or {@link #UNLIMITED_USES}
      * @since 4.13.0
      */
-    boolean multiUse();
+    int uses();
 
     /**
      * How long this callback will last until it is made invalid.
      *
-     * <p>By default callbacks last 12 hours..</p>
+     * <p>By default callbacks last 12 hours.</p>
      *
      * @return the duration of this callback
      * @since 4.13.0
@@ -206,13 +212,13 @@ public interface ClickCallback<T extends Audience> {
     @ApiStatus.NonExtendable
     interface Builder extends AbstractBuilder<Options> {
       /**
-       * Set whether the callback can be multi-use.
+       * Set the number of uses allowed for this callback.
        *
-       * @param multiUse whether multiple clicks are allowed
+       * @param useCount the number of allowed uses, or {@link Options#UNLIMITED_USES}
        * @return this builder
        * @since 4.13.0
        */
-      @NotNull Builder multiUse(boolean multiUse);
+      @NotNull Builder uses(int useCount);
 
       /**
        * Set how long the callback should last from sending.
--- a/api/src/main/java/net/kyori/adventure/text/event/ClickCallback.java
+++ b/api/src/main/java/net/kyori/adventure/text/event/ClickCallback.java
@@ -1,0 +1,145 @@
+/*
+ * This file is part of adventure, licensed under the MIT License.
+ *
+ * Copyright (c) 2017-2023 KyoriPowered
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in all
+ * copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+ * SOFTWARE.
+ */
+package net.kyori.adventure.text.event;
+
+import java.time.Duration;
+import java.time.temporal.TemporalAmount;
+import net.kyori.adventure.audience.Audience;
+import net.kyori.adventure.builder.AbstractBuilder;
+import net.kyori.adventure.util.PlatformAPI;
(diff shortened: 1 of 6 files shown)


api/src/main/java/net/kyori/adventure/text/event/ClickCallback.java
```diff
@@ -0,0 +1,246 @@
...
+      } else if (failureMessage != null) {
+        audience.sendMessage(failureMessage);
+      }
+    };
+  }
+
+  /**
+   * Require that audiences receiving this callback have a certain permission.
+   *
+   * <p>For audiences without permissions information, this test will always pass.</p>
```
kashike: Is this a good idea?

api/src/main/java/net/kyori/adventure/text/event/ClickCallback.java
```diff
@@ -0,0 +1,246 @@
...
+   * @since 4.13.0
+   */
+  default @NotNull ClickCallback<T> requiringPermission(final @NotNull String permission) {
+    return this.requiringPermission(permission, null);
+  }
+
+  /**
+   * Require that audiences receiving this callback have a certain permission.
+   *
+   * <p>For audiences without permissions information, this test will always pass.</p>
```
kashike: Is this a good idea?

api/src/main/java/net/kyori/adventure/text/event/ClickCallback.java
```diff
@@ -0,0 +1,246 @@
...
+
+    /**
+     * Whether the callback can be executed multiple times.
+     *
+     * <p>By default callbacks are single-use.</p>
+     *
+     * @return whether this callback is multi-use
+     * @since 4.13.0
+     */
+    boolean multiUse();
```
kashike:
```suggestion
    int uses();
```

api/src/main/java/net/kyori/adventure/text/event/ClickCallback.java
```diff
@@ -0,0 +1,246 @@
...
+    @ApiStatus.NonExtendable
+    interface Builder extends AbstractBuilder<Options> {
+      /**
+       * Set whether the callback can be multi-use.
+       *
+       * @param multiUse whether multiple clicks are allowed
+       * @return this builder
+       * @since 4.13.0
+       */
+      @NotNull Builder multiUse(boolean multiUse);
```
kashike:
```suggestion
      @NotNull Builder uses(final int uses);
```

api/src/main/java/net/kyori/adventure/text/event/ClickCallback.java
zml2008: I think it's the most sensible -- think of something like a console or a user-provided audience implementation that may not be permission-aware. Callbacks can only be executed by single-user audiences anyways.

api/src/main/java/net/kyori/adventure/text/event/ClickCallback.java
zml2008: Maybe? Are there many use cases for use counts between 1 and "infinite"?

kashike: could be? just seems a bit friendlier to allow the user of the library to determine how many uses a callback can have?

kezz: thinking of stuff like "first five people to click this message gets some diamonds"

api/src/main/java/net/kyori/adventure/text/event/ClickCallback.java
kezz: i think we should either a) have it fail by default or b) force the user to specify the default state

api/src/main/java/net/kyori/adventure/text/event/ClickCallback.java
```diff
@@ -0,0 +1,252 @@
...
+   *
+   * @param <W> the wider type
+   * @param <N> the narrower type
+   * @param original the original callback of a narrower audience type
+   * @param type the audience type to accept
+   * @param failureMessage the message to send to the audience if it is not of the appropriate type
+   * @return a new callback
+   * @since 4.13.0
+   */
+  static <W extends Audience, N extends W> @NotNull ClickCallback<W> widen(final @NotNull ClickCallback<N> original, final @NotNull Class<N> type, final @Nullable Component failureMessage) {
```
kezz: Is there a reason these are static rather than instance methods like the rest?

api/src/main/java/net/kyori/adventure/text/event/ClickCallback.java
```diff
@@ -0,0 +1,252 @@
...
+  /**
+   * Filter audiences that receive this click callback.
+   *
+   * <p>Actions from audiences that do not match this predicate will be silently ignored.</p>
+   *
+   * @param filter the filter to test audiences with
+   * @return a filtered callback action
+   * @since 4.13.0
+   */
+  default @NotNull ClickCallback<T> filter(final @NotNull Predicate<T> filter) {
```
kezz, line 123: these could have pure contract annotations

api/src/main/java/net/kyori/adventure/text/event/ClickCallback.java
```diff
@@ -0,0 +1,252 @@
...
+  interface Provider {
+    /**
+     * Create a real click event based on the provided parameters.
+     *
+     * @param callback the callback to execute
+     * @param options the options to apply to this callback
+     * @return a created click event that will execute the provided callback with options
+     * @since 4.13.0
+     */
+    @NotNull ClickEvent create(final @NotNull ClickCallback<Audience> callback, final ClickCallback.@NotNull Options options);
```
kashike:
```suggestion
    @NotNull ClickEvent create(final @NotNull ClickCallback<Audience> callback, final @NotNull Options options);
```

api/src/main/java/net/kyori/adventure/text/event/ClickCallback.java
```diff
@@ -0,0 +1,252 @@
...
+     *
+     * @return allowable use count, or {@link #UNLIMITED_USES}
+     * @since 4.13.0
+     */
+    int uses();
+
+    /**
+     * How long this callback will last until it is made invalid.
+     *
+     * <p>By default callbacks last 12 hours.</p>
```
kashike: perhaps a constant? `Duration DEFAULT_LIFETIME = Duration.ofHours(12);` roughly

kashike (changes requested):

api/src/main/java/net/kyori/adventure/text/event/ClickCallback.java
```diff
@@ -0,0 +1,252 @@
...
+   * @param failureMessage the message to send to the audience if it is not of the appropriate type
+   * @return a new callback
+   * @since 4.13.0
+   */
+  static <W extends Audience, N extends W> @NotNull ClickCallback<W> widen(final @NotNull ClickCallback<N> original, final @NotNull Class<N> type, final @Nullable Component failureMessage) {
+    return audience -> {
+      if (type.isInstance(audience)) {
+        original.accept(type.cast(audience));
+      } else if (failureMessage != null) {
+        audience.sendMessage(failureMessage);
```
kashike: Perhaps instead of having a `Component` argument here, we could have a `Consumer<? super Audience>` instead?
```suggestion
  static <W extends Audience, N extends W> @NotNull ClickCallback<W> widen(final @NotNull ClickCallback<N> original, final @NotNull Class<N> type, final @Nullable Consumer<? super Audience> onTypeMismatch) {
    return audience -> {
      if (type.isInstance(audience)) {
        original.accept(type.cast(audience));
      } else if (onTypeMismatch != null) {
        onTypeMismatch.accept(audience);
```

api/src/main/java/net/kyori/adventure/text/event/ClickCallback.java
```diff
@@ -0,0 +1,252 @@
...
+
+  /**
+   * Filter audiences that receive this click callback.
+   *
+   * @param filter the filter to test audiences with
+   * @param failureMessage a message to send if the conditions are not met
+   * @return a filtered callback action
+   * @since 4.13.0
+   */
+  default @NotNull ClickCallback<T> filter(final @NotNull Predicate<T> filter, final @Nullable Component failureMessage) {
```
kashike: same here re: last parameter

kashike (changes requested):

api/src/main/java/net/kyori/adventure/text/event/ClickCallback.java
```diff
@@ -0,0 +1,259 @@
...
+     *
+     * @return allowable use count, or {@link #UNLIMITED_USES}
+     * @since 4.13.0
+     */
+    int uses();
+
+    /**
+     * How long this callback will last until it is made invalid.
+     *
+     * <p>By default callbacks last {@link #DEFAULT_LIFETIME 12 hours}.</p>
```
kezz: id prefer we just have the "12 hours" in the default lifetime annotation so we don't inevitably forget to change this comment

api/src/main/java/net/kyori/adventure/text/event/ClickCallback.java
```diff
@@ -0,0 +1,259 @@
...
+   * @since 4.13.0
+   */
+  @ApiStatus.NonExtendable
+  interface Options extends Examinable {
+    /**
+     * Indicate that a callback should have unlimited uses.
+     *
+     * @since 4.13.0
+     */
+    int UNLIMITED_USES = -1;
```
kezz: any reason why this one is in the options but the duration is in the callback? should really be both here

api/src/main/java/net/kyori/adventure/text/event/ClickCallback.java
```diff
@@ -0,0 +1,259 @@
...
+public interface ClickCallback<T extends Audience> {
+  /**
+   * The default lifetime of a callback after creating it.
+   *
+   * @since 4.13.0
+   */
+  Duration DEFAULT_LIFETIME = Duration.ofHours(12);
+
+  /**
+   * Adjust this callback to accept any audience, and perform the appropriate filtering.
```
kezz, line 64: any reason this has to be static rather than an instance method?

api/src/main/java/net/kyori/adventure/text/event/ClickEvent.java
```diff
@@ -136,6 +139,46 @@ public final class ClickEvent implements Examinable, StyleBuilderApplicable {
     return new ClickEvent(Action.COPY_TO_CLIPBOARD, text);
   }
 
+  /**
+   * Create a click event that, when clicked, will schedule a callback function to be executed on the server.
+   *
+   * <p>By default, this will be a single-use function that expires after 12 hours. Exact lifetime is subject to change.</p>
```
kezz: also here - id just link to the timeout object

kezz (changes requested):

api/src/main/java/net/kyori/adventure/text/event/ClickCallback.java
zml2008: I did try to do this as an instance method, but the generics only worked out as a static method for a reason I don't remember enough to articulate well at the moment.

api/src/main/java/net/kyori/adventure/text/event/ClickCallback.java
zml2008: thb probably best to put them both at top-level, just for better visibility and shorter field names

api/src/main/java/net/kyori/adventure/text/event/ClickCallback.java
zml2008: See comment above, it's to make the generics work out

api/src/main/java/net/kyori/adventure/text/event/ClickCallbackOptionsImpl.java
```diff
@@ -0,0 +1,100 @@
...
+  public String toString() {
+    return Internals.toString(this);
+  }
+
+  static final class BuilderImpl implements Builder {
+    private int uses;
+    private Duration lifetime;
+
+    BuilderImpl() {
+      this.uses = 1;
```
kashike: Is it worth having a constant for this as well?

kashike (approved):

PR #862, opened 2023-01-17
Make DecorationMap$KeySet and $Values follow Collection::toArray() contract

emilyy-dev:
`Collection::toArray()` javadoc states that the array's runtime component type must be `Object`


kezz (approved):

rymiel (approved):
thank you java, very cool

kashike (approved):

PR #870, opened 2023-02-06
fix build status badge in readme

neziw:


zml2008 (changes requested):
wrong branch, that was never updated when we changed over

neziw:
> wrong branch, that was never updated when we changed over

so should I change to `main/4`?

kashike (approved):

zml2008 (approved):

PR #871, opened 2023-02-07
feat(api): allow Translators to create their own Components

qixils:
Allows Translators to bypass the loading and parsing of MessageFormat in favor of providing their own rendered Components. Use cases may include using MiniMessage in i18n files (instead of §) or even raw JSON strings/objects.

Was kinda surprised to see GlobalTranslator worked exclusively with MessageFormat (seems very implementation-specific to how vanilla does it) so I whipped this up… tried to keep the PR simple but I could also see it being desirable to have a dedicated class/interface for a ComponentTranslator. Open to feedback/suggestions/etc 🙂

### DIFF
--- a/api/src/main/java/net/kyori/adventure/text/renderer/TranslatableComponentRenderer.java
+++ b/api/src/main/java/net/kyori/adventure/text/renderer/TranslatableComponentRenderer.java
@@ -73,6 +73,13 @@ public abstract class TranslatableComponentRenderer<C> extends AbstractComponent
       protected @Nullable MessageFormat translate(final @NotNull String key, final @NotNull Locale context) {
         return source.translate(key, context);
       }
+
+      @Override
+      protected @NotNull Component renderTranslatable(final @NotNull TranslatableComponent component, final @NotNull Locale context) {
+        final @Nullable Component translated = source.translate(component, context);
+        if (translated != null) return translated;
+        return super.renderTranslatable(component, context);
+      }
     };
   }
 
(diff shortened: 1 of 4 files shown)


zml2008:
I feel like there's some overlap between the purpose of #842 and the purpose of this PR. I'm not sure there's much *harm* in having both, but do any others have thoughts on ways those could/should work together?

qixils:
oops, just saw this-- I did see that PR but I had only glanced at it. looking at it now, I'm honestly still not quite sure I understand how it's supposed to be used, at least in this context? it doesn't seem like there's any way for a plugin to add their own serializer/translator for virtual components which brings us kinda back to square one of why I wrote this PR in the first place 😛 if that were possible then it probably could obsolete this PR, sure. either way i don't immediately see any ways to integrate the two PRs

api/src/main/java/net/kyori/adventure/text/renderer/TranslatableComponentRenderer.java
```diff
@@ -73,6 +73,13 @@
       protected @Nullable MessageFormat translate(final @NotNull String key, final @NotNull Locale context) {
         return source.translate(key, context);
       }
+
+      @Override
+      protected @NotNull Component renderTranslatable(final @NotNull TranslatableComponent component, final @NotNull Locale context) {
```
kashike, line 81: I think it would be best to document the new behavior as well.

qixils: 👍🏻

kashike (approved):

zml2008 (approved):

kezz (approved):

PR #886, opened 2023-03-25
feat(api): Add split method to Component

Joo200:
Add a method to Component to split it into a List.

This can be used to split a Component into a List for a Lore:
```java
MiniMessage.miniMessage().deserialize("<red>Hello world, this is<br>my lore").split(Component.newline());
```

# How Components in the list are formatted

The style of the root component is applied as fallback style to each Component in the list.
```java
    final Component example = Component.text("first", NamedTextColor.RED).appendNewline().append(Component.text("second"));
```
will translate to the following List:
```
[TextComponentImpl{
  content="", style=StyleImpl{obfuscated=not_set, bold=not_set, strikethrough=not_set, underlined=not_set, italic=not_set, color=NamedTextColor{name="red", value="#ff5555"}, clickEvent=null, hoverEvent=null, insertion=null, font=null}, 
    children=[
      TextComponentImpl{content="first", style=StyleImpl{obfuscated=not_set, bold=not_set, strikethrough=not_set, underlined=not_set, italic=not_set, color=null, clickEvent=null, hoverEvent=null, insertion=null, font=null}, children=[]}
    ]
  }, 
TextComponentImpl{
  content="", style=StyleImpl{obfuscated=not_set, bold=not_set, strikethrough=not_set, underlined=not_set, italic=not_set, color=NamedTextColor{name="red", value="#ff5555"}, clickEvent=null, hoverEvent=null, insertion=null, font=null}, 
    children=[
      TextComponentImpl{content="second", style=StyleImpl{obfuscated=not_set, bold=not_set, strikethrough=not_set, underlined=not_set, italic=not_set, color=null, clickEvent=null, hoverEvent=null, insertion=null, font=null}, children=[]}
    ]
  }
]
```

This is still a proof of concept. I'm happy on every optimization.

Currently it's not possible to split the Component by some text. E.g. `Component.text("mydummytext").split(Component.text("dummy"))` will not be split to `Component.text("my")` and `Component.text("text")`

However this works fine to split newlines.

### DIFF
--- a/api/src/main/java/net/kyori/adventure/text/Component.java
+++ b/api/src/main/java/net/kyori/adventure/text/Component.java
@@ -2642,6 +2642,19 @@ public interface Component extends ComponentBuilderApplicable, ComponentLike, Ex
     return this.replaceText(b -> b.match(pattern).replacement(replacement).condition(fn));
   }
 
+  /**
+   * Split an existing Component into multiple Components based on a separator.
+   *
+   * <p>This will return a list of Components</p>
+   *
+   * @param separator a separator
+   * @return a list of components
+   * @since 4.15.0
+   */
+  default @NotNull List<Component> split(final @NotNull Component separator) {
+    return ComponentSplit.split(this, separator);
+  }
+
   @Override
   default void componentBuilderApply(final @NotNull ComponentBuilder<?, ?> component) {
     component.append(this);
--- a/api/src/main/java/net/kyori/adventure/text/ComponentSplit.java
+++ b/api/src/main/java/net/kyori/adventure/text/ComponentSplit.java
@@ -1,0 +1,56 @@
+/*
+ * This file is part of adventure, licensed under the MIT License.
+ *
+ * Copyright (c) 2017-2023 KyoriPowered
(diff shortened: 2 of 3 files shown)


zml2008 (changes requested):
Thanks for taking a stab at this!

I'm afraid I'm going to be a bit particular here, since this doesn't need any internal access so a basic implementation *could* be done externally. I'd like this to work for arbitrary substrings, the same way replaceText does and preserving style there.

CubBossa:
Hey!
I had some issues with the suggested split method in combination with Components parsed from MiniMessage.
MiniMessage created TextComponents that had a content like "line1\nline2" and the check for equallity with a Component.newLine() did not match.
This is my temporary solution to preserve the style and split a component by its content and a regex. It does not match expressions that go beyond one component though (like `<a>x</a>y` would not match xy), so it would not be sufficient as actual implementation for a split by regex.

Maybe someone has need of it until there is an official solution
https://gist.github.com/CubBossa/8bc84706b7e1e393bebb7dc9cf8a9ed5

PR #890, opened 2023-04-02
Added legacy show achievements

LOOHP:
Before Minecraft 1.12, there is a hover event action called `show_achievement`.
Here's an example of it:
```json
{"action":"show_achievement","value":"achievement.bakeCake"}
```
This PR adds legacy support to it. 

Vanilla clients on or after 1.12 can silently ignore these hover events, so there should be no need to deliberately remove them if someone were to send them to modern clients.

### DIFF
--- a/api/src/main/java/net/kyori/adventure/text/event/HoverEvent.java
+++ b/api/src/main/java/net/kyori/adventure/text/event/HoverEvent.java
@@ -210,6 +210,19 @@ public final class HoverEvent<V> implements Examinable, HoverEventSource<V>, Sty
   }
 
   /**
+   * Creates a hover event that shows an achievement on hover.
+   *
+   * @param value the achievement value
+   * @return a hover event
+   * @since 4.13.1
+   * @deprecated Removed in Vanilla 1.12, but we keep it for backwards compat
+   */
+  @Deprecated
+  public static @NotNull HoverEvent<String> showAchievement(final @NotNull String value) {
+    return new HoverEvent<>(Action.SHOW_ACHIEVEMENT, value);
+  }
+
+  /**
    * Creates a hover event.
    *
    * @param action the action
@@ -699,13 +712,26 @@ public final class HoverEvent<V> implements Examinable, HoverEventSource<V>, Sty
         return value.name(renderer.render(value.name, context));
       }
     });
+    /**
+     * Shows a {@link Component} when hovered over.
+     *
+     * @since 4.13.1
+     * @deprecated Removed in Vanilla 1.12, but we keep it for backwards compat
+     */
+    @Deprecated
+    public static final Action<String> SHOW_ACHIEVEMENT = new Action<>("show_achievement", String.class, true, new Renderer<String>() {
+      @Override
+      public <C> @NotNull String render(final @NotNull ComponentRenderer<C> renderer, final @NotNull C context, final @NotNull String value) {
+        return value;
+      }
+    });
 
     /**
      * The name map.
      *
      * @since 4.0.0
      */
-    public static final Index<String, Action<?>> NAMES = Index.create(constant -> constant.name, SHOW_TEXT, SHOW_ITEM, SHOW_ENTITY);
+    public static final Index<String, Action<?>> NAMES = Index.create(constant -> constant.name, SHOW_TEXT, SHOW_ITEM, SHOW_ENTITY, SHOW_ACHIEVEMENT);
     private final String name;
     private final Class<V> type;
     /**
--- a/api/src/main/java/net/kyori/adventure/text/event/HoverEvent.java
+++ b/api/src/main/java/net/kyori/adventure/text/event/HoverEvent.java
@@ -214,7 +214,7 @@ public final class HoverEvent<V> implements Examinable, HoverEventSource<V>, Sty
    *
    * @param value the achievement value
    * @return a hover event
-   * @since 4.13.1
+   * @since 4.14.0
    * @deprecated Removed in Vanilla 1.12, but we keep it for backwards compat
    */
   @Deprecated
@@ -715,7 +715,7 @@ public final class HoverEvent<V> implements Examinable, HoverEventSource<V>, Sty
     /**
      * Shows a {@link Component} when hovered over.
      *
-     * @since 4.13.1
+     * @since 4.14.0
      * @deprecated Removed in Vanilla 1.12, but we keep it for backwards compat
      */
     @Deprecated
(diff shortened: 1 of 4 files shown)


api/src/main/java/net/kyori/adventure/text/event/HoverEvent.java
```diff
@@ -209,6 +209,19 @@
     return new HoverEvent<>(Action.SHOW_ENTITY, entity);
   }
 
+  /**
+   * Creates a hover event that shows an achievement on hover.
+   *
+   * @param value the achievement value
+   * @return a hover event
+   * @since 4.13.1
```
zml2008:
```suggestion
   * @since 4.14.0
```

(we wouldn't add new API in a semver patch release anyways)

api/src/main/java/net/kyori/adventure/text/event/HoverEvent.java
```diff
@@ -699,13 +712,26 @@ public String toString() {
         return value.name(renderer.render(value.name, context));
       }
     });
+    /**
+     * Shows a {@link Component} when hovered over.
+     *
+     * @since 4.13.1
```
zml2008:
```suggestion
     * @since 4.14.0
```

zml2008 (changes requested):
I think this should be fine to add, as much as it is of limited use to most users it doesn't really harm anything or increase maintenance burden. Just a few nitpicks on style though.

LOOHP:
There you go :)

zml2008 (dismissed):

kashike (dismissed):

text-serializer-gson/src/main/java/net/kyori/adventure/text/serializer/gson/SerializerFactory.java
```diff
@@ -26,6 +26,7 @@
 import com.google.gson.Gson;
 import com.google.gson.TypeAdapter;
 import com.google.gson.TypeAdapterFactory;
+import com.google.gson.internal.bind.TypeAdapters;
```
emilyy-dev: Shouldn't depend on `com.google.gson.**internal**.bind.TypeAdapters` (it will also fail in a JPMS environment)

text-serializer-gson/src/main/java/net/kyori/adventure/text/serializer/gson/SerializerFactory.java
```diff
@@ -78,6 +80,8 @@ public <T> TypeAdapter<T> create(final Gson gson, final TypeToken<T> type) {
       return (TypeAdapter<T>) ShowItemSerializer.create(gson);
     } else if (SHOW_ENTITY_TYPE.isAssignableFrom(rawType)) {
       return (TypeAdapter<T>) ShowEntitySerializer.create(gson);
+    } else if (STRING_TYPE.isAssignableFrom(rawType)) {
+      return (TypeAdapter<T>) TypeAdapters.STRING;
```
emilyy-dev: This is not needed, gson will already provide that adapter when requesting one for `TypeToken<String>`

text-serializer-gson/src/main/java/net/kyori/adventure/text/serializer/gson/StyleSerializer.java
```diff
@@ -200,7 +204,7 @@ private boolean readBoolean(final JsonReader in) throws IOException {
   }
 
   private Object legacyHoverEventContents(final HoverEvent.Action<?> action, final Component rawValue) {
-    if (action == HoverEvent.Action.SHOW_TEXT) {
+    if (action == HoverEvent.Action.SHOW_TEXT || action == HoverEvent.Action.SHOW_ACHIEVEMENT) {
```
emilyy-dev: I'm not entirely sure how the hover worked back then, but `rawValue` here is a `Component` and the function is only ever called when the action is `show_text` (it's only called when the action type is `Component`, that only corresponds to `show_text`), maybe I'm missing something but why is this needed if the `show_achievement` action of type `String` (and therefore this function never called for it)?

emilyy-dev (changes requested):

LOOHP:
There you go (again)

text-serializer-gson/src/main/java/net/kyori/adventure/text/serializer/gson/StyleSerializer.java
```diff
@@ -298,7 +304,7 @@ public void write(final JsonWriter out, final Style value) throws IOException {
   }
 
   private void serializeLegacyHoverEvent(final HoverEvent<?> hoverEvent, final JsonWriter out) throws IOException {
-    if (hoverEvent.action() == HoverEvent.Action.SHOW_TEXT) { // serialization is the same
+    if (hoverEvent.action() == HoverEvent.Action.SHOW_TEXT || hoverEvent.action() == HoverEvent.Action.SHOW_ACHIEVEMENT) { // serialization is the same
       this.gson.toJson(hoverEvent.value(), SerializerFactory.COMPONENT_TYPE, out);
```
zml2008, line 308: wrong type here for show_achievement

zml2008 (changes requested):

LOOHP:
And again :)

text-serializer-gson/src/main/java/net/kyori/adventure/text/serializer/gson/StyleSerializer.java
```diff
@@ -277,6 +281,8 @@ public void write(final JsonWriter out, final Style value) throws IOException {
         this.gson.toJson(hoverEvent.value(), SerializerFactory.SHOW_ENTITY_TYPE, out);
       } else if (action == HoverEvent.Action.SHOW_TEXT) {
         this.gson.toJson(hoverEvent.value(), SerializerFactory.COMPONENT_TYPE, out);
+      } else if (action == HoverEvent.Action.SHOW_ACHIEVEMENT) {
+        this.gson.toJson(hoverEvent.value(), SerializerFactory.STRING_TYPE, out);
       } else {
         throw new JsonParseException("Don't know how to serialize " + hoverEvent.value());
       }
```
emilyy-dev: I've been revisiting this a few times today (specifically in the context of [these lines](https://github.com/KyoriPowered/adventure/pull/890/files#diff-5109808b182eb56ebbaf1596841ba42362108e40c680c60a55b0f6ac568368bbR276-R292)) and I can't help but think that:
1. serialization of `"show_achievement"` should go in `serializeLegacyHoverEvent`, it does not belong in `"contents"` as during its lifetime it's only existed in `"value"`… which raises the question:
2. should the exception for "unknown hover actions" (Don't know how to serialize X) not be thrown when it's `"show_achievement"`? What should happen when `emitLegacyHover` is disabled too? Hover json serialization might need a review although that probably fits much better in a separate issue/discussion/pull request

This is mostly food for thought but I'd like to know the opinions from kashike and zml (and any other lurking living being!)

zml2008: I've pushed a change to emit the legacy hover event in *only* the legacy `value` field, rather than the modern `contents` field. Additionally, I've modified the serialization logic to just not serialize a hover event at all if legacy hover event output is not enabled and the action is `show_achievement`, just for efficiency's sake.

zml2008 (approved):

PR #896, opened 2023-04-09
fix(text-minimessage): Correctly handle phase and multi-colour gradients

### LINKED ISSUE #889: MiniMessage - Gradients with > 3 colors render issue
Today i worked around a bit with some bigger gradients (3 colors or more). 
Apparently if there are more then 2 colors and you try to animate them with the offset number it bugs out. For example, first image works correct and second image should be offset half way but it kinda stucks in the middle.

(Images are from the WebUI but behave the same ingame)

---

For 4 colors it groups to 3 cut of sections and so on...

zml2008:
Fixes GH-889

This will result in some colour shifting for end users, but the results seem like a huge improvement to me at least.

### DIFF
--- a/text-minimessage/src/main/java/net/kyori/adventure/text/minimessage/tag/standard/GradientTag.java
+++ b/text-minimessage/src/main/java/net/kyori/adventure/text/minimessage/tag/standard/GradientTag.java
@@ -58,7 +58,7 @@ final class GradientTag extends AbstractColorChangingTag {
   private @Range(from = -1, to = 1) double phase;
 
   static Tag create(final ArgumentQueue args, final Context ctx) {
-    float phase = 0;
+    double phase = 0;
     final List<TextColor> textColors;
     if (args.hasNext()) {
       textColors = new ArrayList<>();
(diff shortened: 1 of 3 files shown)


text-minimessage/src/main/java/net/kyori/adventure/text/minimessage/tag/standard/GradientTag.java
```diff
@@ -51,12 +51,11 @@ final class GradientTag extends AbstractColorChangingTag {
...
   private int index = 0;
-  private int colorIndex = 0;
 
-  private float factorStep = 0;
+  private double multiplier = 1;
+
   private final TextColor[] colors;
-  private float phase;
-  private final boolean negativePhase;
+  private @Range(from = -1, to = 1) double phase;
```
rymiel, line 58: miniscule nit: here's a place below on line 71 (which github won't let me comment on) where phase is assigned from a double casted to a float which is probably redundant now that phase is a double (altho the information loss is so inconsequential)

rymiel (approved):

PR #898, opened 2023-04-10
ANSIComponentSerializer

rymiel:
Supersedes #140
Fixes KyoriPowered/ansi#7
Fixes #125

PR #899, opened 2023-04-11
ComponentEncoder

rymiel:
ComponentEncoder is akin to ComponentSerializer, except it doesn't provide deserialization. This is intended to be used in #898.

ComponentSerializer now extends ComponentEncoder, but still overrides the methods with exact copies, to keep the javadocs more clear. These originals also retain their old `@since` tags.

### DIFF
--- a/api/src/main/java/net/kyori/adventure/text/serializer/ComponentEncoder.java
+++ b/api/src/main/java/net/kyori/adventure/text/serializer/ComponentEncoder.java
@@ -24,85 +24,27 @@
 package net.kyori.adventure.text.serializer;
 
 import net.kyori.adventure.text.Component;
-import org.jetbrains.annotations.ApiStatus;
 import org.jetbrains.annotations.Contract;
 import org.jetbrains.annotations.NotNull;
 import org.jetbrains.annotations.Nullable;
 
 /**
- * A {@link Component} serializer and deserializer.
+ * A {@link Component} encoder, which provides serialization, but without deserialization.
+ *
+ * <p>For both serialization and deserialization, use {@link ComponentSerializer}</p>
  *
  * @param <I> the input component type
- * @param <O> the output component type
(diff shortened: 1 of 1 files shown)


api/src/main/java/net/kyori/adventure/text/encoder/package-info.java
```diff
@@ -0,0 +1,27 @@
...
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+ * SOFTWARE.
+ */
+/**
+ * Serialization for components.
```
kezz: Encoding mayhaps (or just throw it all in the serialiser package)?

kezz (approved):
Minor nitpick

api/src/main/java/net/kyori/adventure/text/encoder/package-info.java
zml2008: imo we should just put this in the serializer package, they're closely associated anyways

kezz (approved):

zml2008 (approved):

kashike (approved):

PR #903, opened 2023-04-15
feat: abstract legacy code conversion

Camotoy:
Bedrock already has one new "legacy" color code, "Minecoin gold". In 1.19.80, there will be a handful of new legacy color codes. The primary goal of this PR is to allow LegacyComponentSerializer usages to take advantage of these new color codes for hex color downsampling. While designing this PR, I also wanted to allow other non-color codes to be modified, since Bedrock doesn't have underscore and obfuscation. We already work around this, however, so this feature is not important to the PR.

For now, I only request feedback on the general structure of the PR. There are several Javadoc and checkstyle errors that I have yet to address until it is agreed upon that the implementation presented here is acceptable. Also, this PR has not yet been tested, for the same reason.

### DIFF
--- a/api/src/main/java/net/kyori/adventure/text/format/NamedTextColor.java
+++ b/api/src/main/java/net/kyori/adventure/text/format/NamedTextColor.java
@@ -156,7 +156,10 @@ public final class NamedTextColor implements TextColor {
    */
   public static final NamedTextColor WHITE = new NamedTextColor("white", WHITE_VALUE);
 
-  private static final List<NamedTextColor> VALUES = Collections.unmodifiableList(Arrays.asList(BLACK, DARK_BLUE, DARK_GREEN, DARK_AQUA, DARK_RED, DARK_PURPLE, GOLD, GRAY, DARK_GRAY, BLUE, GREEN, AQUA, RED, LIGHT_PURPLE, YELLOW, WHITE));
+  /**
+   * @since 4.14.0
+   */
+  public static final List<NamedTextColor> VALUES = Collections.unmodifiableList(Arrays.asList(BLACK, DARK_BLUE, DARK_GREEN, DARK_AQUA, DARK_RED, DARK_PURPLE, GOLD, GRAY, DARK_GRAY, BLUE, GREEN, AQUA, RED, LIGHT_PURPLE, YELLOW, WHITE));
   /**
    * An index of name to color.
    *
--- a/api/src/main/java/net/kyori/adventure/text/format/NamedTextColor.java
+++ b/api/src/main/java/net/kyori/adventure/text/format/NamedTextColor.java
@@ -156,10 +156,7 @@ public final class NamedTextColor implements TextColor {
    */
   public static final NamedTextColor WHITE = new NamedTextColor("white", WHITE_VALUE);
 
-  /**
-   * @since 4.14.0
-   */
-  public static final List<NamedTextColor> VALUES = Collections.unmodifiableList(Arrays.asList(BLACK, DARK_BLUE, DARK_GREEN, DARK_AQUA, DARK_RED, DARK_PURPLE, GOLD, GRAY, DARK_GRAY, BLUE, GREEN, AQUA, RED, LIGHT_PURPLE, YELLOW, WHITE));
+  private static final List<NamedTextColor> VALUES = Collections.unmodifiableList(Arrays.asList(BLACK, DARK_BLUE, DARK_GREEN, DARK_AQUA, DARK_RED, DARK_PURPLE, GOLD, GRAY, DARK_GRAY, BLUE, GREEN, AQUA, RED, LIGHT_PURPLE, YELLOW, WHITE));
   /**
    * An index of name to color.
    *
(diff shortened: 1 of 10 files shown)


Joo200:
~~First of all I don't think we should support or add more "legacy color codes". It's legacy because it's outdated and shouldn't be used anymore. New types can be added to MiniMessage with more possibilities and richer formatting.~~

Imo your changes are limited and adding new types is pretty complicated. A better approach could be an interface (LegacyCodeConverter) which converts the string to a style. E.g.
```java
public interface LegacyCodeConverter {
  /**
   * Convert the text with the legacy char to a format.
   * return: the style, null if no color can be determined.
   */ 
  TextFormat tagToFormat(final char legacy, final String input, final int pos);

  /**
   * Convert the TextFormat to a message.
   * return: a message, null if the color can't be formatted correctly
   */
  String formatToTag(TextFormat format);
}
```

This would create a few classes more but you would only need to store a List of Tag types and can add simply more formatting tags.

~~**But**: The more I read that code and think about that topic I disagree on expanding the Legacy Converter logic. Keep it limited, keep it stupid, let's move people to MiniMessage.
My suggested changes are already implemented in MiniMessage (with TagResolvers) and there is no need to manage multiple Converters with the same possibilities.~~

Edit: Just got the message that Bedrock doesn't use components at all. Ignore my blame of legacy stuff here.

text-serializer-legacy/src/main/java/net/kyori/adventure/text/serializer/legacy/LegacyCodeConverter.java
```diff
@@ -0,0 +1,171 @@
...
+  protected LegacyCodeConverter(final char character, final char hexCharacter, final boolean hexColours, final boolean useTerriblyStupidHexFormat) {
+    this.character = character;
+    this.hexCharacter = hexCharacter;
+    this.hexColours = hexColours;
+    this.useTerriblyStupidHexFormat = useTerriblyStupidHexFormat;
+  }
+
+  @NotNull abstract List<TextFormat> getFormats();
+
+  @NotNull abstract String getAllLegacyChars();
```
kashike: In my opinion, instead of this we should have some sort of object containing this information:
```java
public final class CharacterAndFormat {
  private final char character;
  private final TextFormat format;
}
```
and pass a `List<CharacterAndFormat>` to the constructor which can convert those into the internal list/string representation.

text-serializer-legacy/src/main/java/net/kyori/adventure/text/serializer/legacy/LegacyCodeConverter.java
```diff
@@ -0,0 +1,171 @@
...
+      final float distance = distance(any.asHSV(), potential.asHSV());
+      if (distance < matchedDistance) {
+        match = potential;
+        matchedDistance = distance;
+      }
+      if (distance == 0) {
+        break; // same colour! whoo!
+      }
+    }
+    return match;
```
kashike: Instead of duplicating this logic, we should refactor the original method in `NamedTextColor` to be something like `public <C extends TextColor> @NotNull C nearestColorTo(final @NotNull List<C> values, final @NotNull TextColor any)` - have the existing method in that class call it with `nearestColorTo(VALUES, any)` and have this new class call it as well.

kashike (changes requested):

api/src/main/java/net/kyori/adventure/text/format/NamedTextColor.java
```diff
@@ -156,7 +156,10 @@ public final class NamedTextColor implements TextColor {
    */
   public static final NamedTextColor WHITE = new NamedTextColor("white", WHITE_VALUE);
 
-  private static final List<NamedTextColor> VALUES = Collections.unmodifiableList(Arrays.asList(BLACK, DARK_BLUE, DARK_GREEN, DARK_AQUA, DARK_RED, DARK_PURPLE, GOLD, GRAY, DARK_GRAY, BLUE, GREEN, AQUA, RED, LIGHT_PURPLE, YELLOW, WHITE));
+  /**
+   * @since 4.14.0
+   */
+  public static final List<NamedTextColor> VALUES = Collections.unmodifiableList(Arrays.asList(BLACK, DARK_BLUE, DARK_GREEN, DARK_AQUA, DARK_RED, DARK_PURPLE, GOLD, GRAY, DARK_GRAY, BLUE, GREEN, AQUA, RED, LIGHT_PURPLE, YELLOW, WHITE));
```
kezz: Doesn't need to be made public - can be obtained from `NAMES.values()`

Camotoy: `NAMES.values()` returns a `Set`.

Camotoy:
Implemented kashike's suggestions. I don't think this is optimal, but maybe a step in the right direction. It might actually be possible to completely get rid of `LegacyCodeConverterImpl` now, but the `legacyFormat` method needs a home.

Camotoy:
Joo's suggestion is also possible, but implementing such a design would have to account for the different hex character styles and whatnot.

Camotoy:
I gave this some thought, and I think this could be implemented much more cleanly: https://github.com/Camotoy/adventure/tree/allow-custom-colors-legacy

This removes the LegacyCodeConverter class and has LegacyComponentSerializerImpl process the custom formats.

Camotoy:
Superseded by #906.

PR #906, opened 2023-04-22
feat: allow custom colors in LegacyComponentSerializer

Camotoy:
Supersedes #903 

Bedrock already has one new "legacy" color code, "Minecoin gold". In 1.19.80, there will be a handful of new legacy color codes. The primary goal of this PR is to allow LegacyComponentSerializer usages to take advantage of these new color codes for hex color downsampling. While designing this PR, I also wanted to allow other non-color codes to be modified, since Bedrock doesn't have underscore and obfuscation. We already work around this, however, so this feature is not important to the PR.

### DIFF
--- a/text-serializer-legacy/src/main/java/net/kyori/adventure/text/serializer/legacy/CharacterAndFormat.java
+++ b/text-serializer-legacy/src/main/java/net/kyori/adventure/text/serializer/legacy/CharacterAndFormat.java
@@ -1,0 +1,33 @@
+package net.kyori.adventure.text.serializer.legacy;
+
+import net.kyori.adventure.text.format.TextFormat;
+import org.jetbrains.annotations.NotNull;
+import org.jetbrains.annotations.Nullable;
+
+public final class CharacterAndFormat {
+  private final char character;
+  private final @Nullable TextFormat format;
+
+  private CharacterAndFormat(final char character) {
+    this.character = character;
+    this.format = null;
+  }
+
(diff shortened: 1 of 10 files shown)


Camotoy:
image
Here's how this looks if an implementation defines custom colors.

text-serializer-legacy/src/main/java/net/kyori/adventure/text/serializer/legacy/CharacterAndFormat.java
```diff
@@ -0,0 +1,275 @@
...
+import org.jetbrains.annotations.Nullable;
+
+import static java.util.Objects.requireNonNull;
+
+/**
+ * A combination of a {@code character} and a {@link TextFormat}.
+ *
+ * @since 4.14.0
+ */
+public final class CharacterAndFormat {
```
zml2008: Kind of a nitpick, but since this is sorta public API we should be a bit more defensive here. Could you change this so `CharacterAndFormat` is an interface implemented by `CharacterAndFormatImpl`, as we do in a lot of other things across the API? Then have a factory method (`CharacterAndFormat characterAndFormat(char, TextColor)`) type thing on the interface, rather than calling the constructor directly.

zml2008 (changes requested):

zml2008 (approved):

kashike (approved):

zml2008 (approved):

PR #925, opened 2023-05-13
fix(key): fix combined namespace+value pattern + extract to own annotations

### LINKED ISSUE #919: Key is parsable despite documented otherwise
String `:my_key` is parsable into `minecraft:my_key`, but it doesn't follow the documented pattern [`([a-z0-9_\\-.]+:)?[a-z0-9_\\-./]+`](https://jd.advntr.dev/key/4.13.1/net/kyori/adventure/key/Key.html#key(java.lang.String)) - the pattern requires at least one symbol for the namespace if such is used.

kashike:


zml2008 (approved):

key/src/main/java/net/kyori/adventure/key/KeyPattern.java
```diff
@@ -0,0 +1,51 @@
...
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+import org.intellij.lang.annotations.Pattern;
+
+@Documented
+@Pattern("(?:(" + KeyImpl.NAMESPACE_PATTERN + ":)?|:)" + KeyImpl.VALUE_PATTERN)
+@Retention(RetentionPolicy.CLASS)
+@Target({ElementType.FIELD, ElementType.LOCAL_VARIABLE, ElementType.METHOD, ElementType.PARAMETER})
+@interface KeyPattern {
```
zml2008, line 37: Do we want this to be public? or is package-private fine?

PR #929, opened 2023-05-27
Use constants for hex character

kashike:
Constant names provided by @EterNityCH and @zachbr.

### DIFF
--- a/api/src/main/java/net/kyori/adventure/text/format/TextColor.java
+++ b/api/src/main/java/net/kyori/adventure/text/format/TextColor.java
@@ -48,6 +48,19 @@ import static java.util.Objects.requireNonNull;
  */
 public interface TextColor extends Comparable<TextColor>, Examinable, RGBLike, StyleBuilderApplicable, TextFormat {
   /**
+   * The hex character.
+   *
+   * @since 4.14.0
+   */
+  char HEX_CHARACTER = '#';
+  /**
+   * The hex character, in {@code String} format.
+   *
+   * @since 4.14.0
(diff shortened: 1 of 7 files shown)


zml2008 (changes requested):
constant names should be SCREAMING_SNAKE_CASE

api/src/main/java/net/kyori/adventure/text/format/TextColor.java
```diff
@@ -47,6 +47,19 @@
  * @since 4.0.0
  */
 public interface TextColor extends Comparable<TextColor>, Examinable, RGBLike, StyleBuilderApplicable, TextFormat {
+  /**
+   * The hex character.
+   *
+   * @since 4.14.0
+   */
+  char SCREAMING_SNAKE_CASE1 = '#';
```
zml2008: maybe `HEX_CHARACTER` and `HEX_CHARACTER_BOXED`? idk

imDaniX: I think the name should somehow indicate that it's used for colors. `TextColor.HEX_CHARACTER` feels very wrong, like, why it's even there.
As such color specifiers are often referred as "color codes", maybe something like `COLOR_CODE_PREFIX`?

zml2008: I mean, it is in `TextColor` -- maybe `HEX_PREFIX` though?

zml2008 (approved):

PR #936, opened 2023-06-08
add javadoc links

powercasgamer:


zml2008:
Id rather not depend on javadoc.io, it's been known to go down sometimes and break builds

kezz:
I'm going to go ahead and close this for the reasons that zml mentioned above. Ideally we should add these to our Javadocs page directly and then link them using that.

PR #938, opened 2023-06-25
feat(mini-message): add Formatter#joining

tahmid-23:
This PR adds a method to Formatter to join components using JoinConfiguration.

### DIFF
--- a/text-minimessage/src/main/java/net/kyori/adventure/text/minimessage/tag/resolver/Formatter.java
+++ b/text-minimessage/src/main/java/net/kyori/adventure/text/minimessage/tag/resolver/Formatter.java
@@ -31,7 +31,6 @@ import java.time.format.DateTimeFormatter;
 import java.time.temporal.TemporalAccessor;
 import java.util.Arrays;
 import java.util.Locale;
-
 import net.kyori.adventure.text.Component;
 import net.kyori.adventure.text.ComponentLike;
 import net.kyori.adventure.text.JoinConfiguration;
@@ -155,9 +154,9 @@ public final class Formatter {
    * @param key the key
    * @param components the components to join
    * @return the placeholder
-   * @since 4.14.0
+   * @since 4.15.0
    */
-  public static TagResolver join(@TagPattern final @NotNull String key, final @NotNull Iterable<? extends ComponentLike> components) {
+  public static TagResolver joining(@TagPattern final @NotNull String key, final @NotNull Iterable<? extends ComponentLike> components) {
     return TagResolver.resolver(key, (argumentQueue, context) -> {
       final String separator = argumentQueue.popOr("Separator expected.").value();
       final JoinConfiguration.Builder configBuilder = JoinConfiguration.builder().separator(context.deserialize(separator));
@@ -167,7 +166,7 @@ public final class Formatter {
         configBuilder.lastSeparator(context.deserialize(lastSeparator));
       }
 
-      JoinConfiguration config = configBuilder.build();
+      final JoinConfiguration config = configBuilder.build();
       return Tag.inserting(Component.join(config, components));
     });
   }
@@ -182,10 +181,10 @@ public final class Formatter {
    * @param key the key
    * @param components the components to join
    * @return the placeholder
-   * @since 4.14.0
+   * @since 4.15.0
    */
-  public static TagResolver join(@TagPattern final @NotNull String key, final @NotNull ComponentLike@NotNull... components) {
-    return join(key, Arrays.asList(components));
+  public static TagResolver joining(@TagPattern final @NotNull String key, final @NotNull ComponentLike@NotNull... components) {
+    return joining(key, Arrays.asList(components));
   }
 
 }
--- a/text-minimessage/src/main/java/net/kyori/adventure/text/minimessage/tag/resolver/Formatter.java
+++ b/text-minimessage/src/main/java/net/kyori/adventure/text/minimessage/tag/resolver/Formatter.java
@@ -147,18 +147,25 @@ public final class Formatter {
   /**
    * Creates a replacement that inserts a list of components. These components are joined together by {@link Component#join(JoinConfiguration.Builder, ComponentLike...)}.
    *
-   * <p>This tag expects a separator as the first argument. A last separator may optionally be provided.</p>
+   * <p>This tag has three optional arguments; a separator, a last separator, and a last separator if serial.
+   * Each argument must be provided in order, with all preceding arguments present.
+   * The exact use of these three separators is documented in {@link JoinConfiguration}.</p>
    *
    * <p>This replacement is auto-closing, so its style will not influence the style of following components.</p>
    *
    * @param key the key
    * @param components the components to join
    * @return the placeholder
-   * @since 4.15.0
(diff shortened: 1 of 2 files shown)


text-minimessage/src/main/java/net/kyori/adventure/text/minimessage/tag/resolver/Formatter.java
```diff
@@ -139,4 +144,48 @@ public static TagResolver booleanChoice(@TagPattern final @NotNull String key, f
...
+   * Creates a replacement that inserts a list of components. These components are joined together by {@link Component#join(JoinConfiguration.Builder, ComponentLike...)}.
+   *
+   * <p>This tag expects a separator as the first argument. A last separator may optionally be provided.</p>
+   *
+   * <p>This replacement is auto-closing, so its style will not influence the style of following components.</p>
+   *
+   * @param key the key
+   * @param components the components to join
+   * @return the placeholder
+   * @since 4.14.0
```
zml2008:
```suggestion
   * @since 4.15.0
```

4.14.0 has already been released.....

text-minimessage/src/main/java/net/kyori/adventure/text/minimessage/tag/resolver/Formatter.java
```diff
@@ -139,4 +144,48 @@ public static TagResolver booleanChoice(@TagPattern final @NotNull String key, f
...
```
zml2008: same here

text-minimessage/src/main/java/net/kyori/adventure/text/minimessage/tag/resolver/Formatter.java
```diff
@@ -139,4 +144,48 @@ public static TagResolver booleanChoice(@TagPattern final @NotNull String key, f
...
+   */
+  public static TagResolver join(@TagPattern final @NotNull String key, final @NotNull Iterable<? extends ComponentLike> components) {
```
zml2008: maybe call this `joining`?

zml2008 (dismissed):
This is a nice addition -- just a few comments. Could you also open a PR on `adventure-docs` documenting this addition?

tahmid-23:
from what I can see, there is also a lastSeparatorIfSerial method which is mutually exclusive, in terms of functionality, with lastSeparator
should I add something like <tag:separator:last_separator:serial>? or <tag:separator:last_separator:true>? in order to use lastSeparatorIfSerial

kezz:
> from what I can see, there is also a lastSeparatorIfSerial method which is mutually exclusive, in terms of functionality, with lastSeparator should I add something like tag:separator:last_separator:serial? or tag:separator:last_separator:true? in order to use lastSeparatorIfSerial

IIRC `lastSeparatorIfSerial` is only used where there are three or more elements in the list (e.g. `x, y, and z`). It will still use `lastSeparator` in the case where there are two elements (e.g. `x and y`).

Joo200:
I already use such a resolver for my project but I don't use a collection of ComponentLike. Instead I use a Collection of TagResolvers instead, with:
```java
  public static TagResolver joining(@TagPattern final @NotNull String key, final @NotNull Iterable<? extends TagResolver> components) {
    return TagResolver.resolver(key, (arguments, ctx) -> {
        String format = arguments.popOr("expected list format").value();
        String separator = arguments.popOr("expected separator").value();
        if (arguments.hasNext()) {
            String emptyVal = arguments.pop().value();
            if (resolvers.isEmpty()) return Tag.inserting(ctx.deserialize(emptyVal));
        }
        List<Component> components = resolvers.stream().map(r -> ctx.deserialize(format, r)).toList();
        Component separatorComponent = ctx.deserialize(separator);
        Component all = components.stream().collect(Component.toComponent(separatorComponent));

        if (arguments.hasNext()) {
            int amount = arguments.popOr("maximum amount").asInt().orElse(5);
            String lastInfo = arguments.popOr("last entry").value();
            String hoverFormat = arguments.popOr("hover").value();
            if (resolvers.size() > amount) {
                Component hover = ctx.deserialize(hoverFormat,
                        Formatter.number("total", resolvers.size()),
                        Formatter.number("remaining", resolvers.size() - amount),
                        Placeholder.component("all", all));
                Component remaining = ctx.deserialize(lastInfo,
                        Formatter.number("total", resolvers.size()),
                        Formatter.number("remaining", resolvers.size() - amount)).hoverEvent(hover);
                return Tag.selfClosingInserting(components.stream().limit(amount).collect(Component.toComponent(separatorComponent)).append(remaining));
            }
        }
        return Tag.selfClosingInserting(all);
  });
}
```

My implementation is complex and I don't think we should use that as api. However you can easily change the format of the list entries, e.g. with such a minimessage format string
```
Governors: <gray><governors:'<online:green:gray><name></online>':', ':'<dark_gray>-- no one --</dark_gray>':'5':'<gold> and <remaining> more</gold>':'<gold>Total <total> governors:<br><gray><all>'>
```
This will be formatted to:
```
Governors: First, second, third, forth, fifth and 3 more.
```
Where `and 3 more` has an hover effect.

I don't think we should add such complex logic to minimessage however I would add the TagResolver approach as you can easily modify the formatting.

tahmid-23:
I added an argument for lastSeparatorIfSerial.

I'm not sure if TagResolvers are the best approach here.
Personally, I think I think that using Components would be more convenient. If I needed a placeholder, I could render it myself.

tahmid-23:
any love? :slightly_smiling_face:

text-minimessage/src/main/java/net/kyori/adventure/text/minimessage/tag/resolver/Formatter.java
```diff
@@ -139,4 +143,53 @@ public static TagResolver booleanChoice(@TagPattern final @NotNull String key, f
       return Tag.inserting(context.deserialize(value ? trueCase : falseCase));
     });
   }
+
+  /**
+   * Creates a replacement that inserts a list of components. These components are joined together by {@link Component#join(JoinConfiguration.Builder, ComponentLike...)}.
+   *
+   * <p>This tag expects a separator as the first argument. A last separator may optionally be provided.</p>
```
kezz: Can this be optional too? In the case where you're happy to just use `JoinConfiguration.noSeparators`?

text-minimessage/src/main/java/net/kyori/adventure/text/minimessage/tag/resolver/Formatter.java
```diff
@@ -139,4 +143,53 @@ public static TagResolver booleanChoice(@TagPattern final @NotNull String key, f
...
+   *
+   * <p>This replacement is auto-closing, so its style will not influence the style of following components.</p>
+   *
+   * @param key the key
+   * @param components the components to join
+   * @return the placeholder
+   * @since 4.15.0
```
kezz: Need updating before merge

kezz (changes requested):
Minor changes that can be done before merge

kezz (approved):
Made the changes myself, LGTM :p

zml2008 (approved):

tahmid-23:
now if only I still played Minecraft... 😛

PR #965, opened 2023-08-25
feat(nbt): expose size in CompoundBinaryTag

RealBauHD:


nbt/src/main/java/net/kyori/adventure/nbt/CompoundBinaryTag.java
```diff
@@ -91,6 +91,14 @@ public interface CompoundBinaryTag extends BinaryTag, CompoundTagSetter<Compound
    */
   @Nullable BinaryTag get(final String key);
 
+  /**
+   * Gets the size.
```
zml2008: Explain what 'size' means here please

zml2008 (changes requested):

nbt/src/main/java/net/kyori/adventure/nbt/CompoundBinaryTag.java
RealBauHD: In ListBinaryTag it is the same, but you mean something like "Gets the size of binary tags in the compound."

zml2008: `ListBinaryTag`'s javadoc could be improved too then in another PR -- but no. Size is meaningless without a unit -- for a concept like a list that people are already familiar with we can kinda get away with bad docs, but for a compound we should specify that `size()` returns the number of elements in the compound.

zml2008 (approved):

PR #968, opened 2023-09-01
feat(nbt): Implement nameless binary serialization

booky10:
Implements serialization of nameless compound tags. Mojang has removed these already useless names in 23w31a (1.20.2 snapshot) and this adds API for reading/writing these "nameless" compound tags.

---

Not sure about how it is exposed, could also be solved by specifying a boolean for `read`/`write` methods


zml2008:
I think this could be more nicely implemented with the upcoming feature flag system coming soon:tm:? Since it's sorta a version-specific serialization quirk.

zml2008:
So giving this more thought, with the changes I would like to see in #1009 being a bit too big to squish into 4.15.0, I think we're ok to take this design more or less as-is, and revision can occur in future versions. Thank you for your contribution!

zml2008 (approved):

PR #981, opened 2023-10-07
feat(api): introduce ResourcePackRequest

yusshu:


kashike (approved):

api/src/main/java/net/kyori/adventure/resourcepack/ResourcePack.java
```diff
@@ -0,0 +1,99 @@
...
+    return new ResourcePackImpl(url, hash, required, prompt);
+  }
+
+  /**
+   * Gets the url.
+   *
+   * @return the url
+   * @since 4.15.0
+   */
+  @NotNull String url();
```
zml2008: what about making this a `URI` instead of a `String`?

api/src/main/java/net/kyori/adventure/resourcepack/ResourcePack.java
```diff
@@ -0,0 +1,99 @@
...
+   * @param url the url
+   * @param hash the sha-1 hash
+   * @param required whether this resource pack is required or not
+   * @param prompt the prompt
+   * @return the resource pack
+   * @since 4.15.0
+   */
+  static @NotNull ResourcePack resourcePack(final @NotNull String url, final @NotNull String hash, final boolean required, final @Nullable Component prompt) {
+    return new ResourcePackImpl(url, hash, required, prompt);
+  }
```
zml2008: Should this be a builder? I feel like there are enough parameters where that could make sense.

Is it worth having a factory method/option in the builder that will fetch the resource from `url` and compute the sha1 hash of it automatically?

api/src/main/java/net/kyori/adventure/resourcepack/ResourcePackImpl.java
```diff
@@ -0,0 +1,98 @@
...
+    return Stream.of(
+      ExaminableProperty.of("url", this.url),
+      ExaminableProperty.of("hash", this.hash),
+      ExaminableProperty.of("required", this.required),
+      ExaminableProperty.of("prompt", this.prompt)
+    );
+  }
+
+  @Override
+  public boolean equals(final Object o) {
```
zml2008:
```suggestion
  public boolean equals(final @Nullable Object other) {

```

api/src/main/java/net/kyori/adventure/resourcepack/ResourcePackImpl.java
```diff
@@ -0,0 +1,98 @@
...
+  }
+
+  @Override
+  public int hashCode() {
+    int result = this.url.hashCode();
+    result = 31 * result + this.hash.hashCode();
+    result = 31 * result + (this.required ? 1 : 0);
+    result = 31 * result + (this.prompt != null ? this.prompt.hashCode() : 0);
+    return result;
+  }
```
zml2008: make sure to override toString() to call out to `Internals.toString()` -- this computes it based on the Examinable information.

api/src/main/java/net/kyori/adventure/resourcepack/package-info.java
```diff
@@ -0,0 +1,27 @@
...
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+ * SOFTWARE.
+ */
+/**
+ * Data for controlling resource packs
```
zml2008:
```suggestion
 * Data for controlling resource packs.

```

zml2008 (changes requested):
Thanks for the contributions! A few design things here, but overall pretty good

api/src/main/java/net/kyori/adventure/resourcepack/ResourcePack.java
kezz: Agreed about builder - people aren't going to be making packs often enough for a static method to make sense

api/src/main/java/net/kyori/adventure/resourcepack/ResourcePack.java
kezz: Agreed here also - URI makes sense for both client side and server side

api/src/main/java/net/kyori/adventure/resourcepack/package-info.java
```diff
@@ -0,0 +1,27 @@
...
+ */
+package net.kyori.adventure.resourcepack;
```
kezz: Is `resourcepack` a good package name? I do find concatenated names like that ugly - we should do `resource` instead; more future proof, esp if we add width calculation framework in the future, for example.

zml2008: (or `resource`, I think we want singular package names)

Machine-Maker:
Is `ResourcePack` a good name? I feel like you aren't really dealing with a resourcepack, like the object that might exist on the client, but rather `ResourcePackInfo` or something like that.

api/src/main/java/net/kyori/adventure/resourcepack/ResourcePack.java
yusshu: I will add the builder, but should I remove the static factory methods?

yusshu:
> Is `ResourcePack` a good name? I feel like you aren't really dealing with a resourcepack, like the object that might exist on the client, but rather `ResourcePackInfo` or something like that.

I'm not sure what its name should be, but I agree that it feels like we are not working with an actual resource pack.
I call this class "ResourcePackRequest" in [my plugin](https://github.com/unnamed/creative-central/blob/main/api/src/main/java/team/unnamed/creative/central/request/ResourcePackRequest.java)

api/src/main/java/net/kyori/adventure/resourcepack/ResourcePack.java
yusshu: I think I shouldn't, `Book` has both factory methods and a builder interface

api/src/main/java/net/kyori/adventure/resource/ResourcePack.java
```diff
@@ -0,0 +1,185 @@
...
+
+import java.net.URI;
+
+/**
+ * Represents a resource pack that can be sent to players.
+ *
+ * @see Audience#sendResourcePack(ResourcePack)
+ * @since 4.15.0
+ */
+public interface ResourcePack extends Buildable<ResourcePack, ResourcePack.Builder>, Examinable {
```
kashike: Do not extend `Buildable` - it is to be considered deprecated.

kashike (changes requested):

api/src/main/java/net/kyori/adventure/resource/ResourcePack.java
kashike: Also, rename to `ResourcePackRequest`.

yusshu: Should I rename the `sendResourcePack` methods to `sendResourcePackRequest` too?

zml2008: I think it makes more sense to just call the method `sendResourcePack`, you have the type info that indicates it's a request already.

zml2008:
Please remember to resolve the codestyle issues in this PR (first run `./gradlew spotlessApply`, then `./gradlew check` and fix any issues that remain)

zml2008 (approved):

api/src/main/java/net/kyori/adventure/resource/ResourcePackRequest.java
```diff
@@ -0,0 +1,182 @@
...
+   */
+  @Nullable Component prompt();
+
+  /**
+   * Create a new builder initialized with the attributes of this resource pack request.
+   *
+   * @return the builder
+   * @since 4.15.0
+   */
+  default @NotNull Builder toBuilder() {
```
kashike: Remove this method.

kashike (dismissed):

kezz (approved):
Should have an accompanying PR on adventure-docs, but otherwise LGTM! Thanks for the contribution :)

PR #986, opened 2023-10-11
1.20.3

zml2008:
This is a branch to gather work on supporting features and representation changes from Minecraft 1.20.3.

Relevant changelogs:

- [23w40a](https://mdcfe.dev/mc-changes?ver=23w40a)
- [23w42a](https://mdcfe.dev/mc-changes?ver=23w42a)

### DIFF
--- a/text-serializer-gson/src/main/java/net/kyori/adventure/text/serializer/gson/UUIDSerializer.java
+++ b/text-serializer-gson/src/main/java/net/kyori/adventure/text/serializer/gson/UUIDSerializer.java
@@ -1,0 +1,60 @@
+/*
+ * This file is part of adventure, licensed under the MIT License.
+ *
+ * Copyright (c) 2017-2023 KyoriPowered
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in all
+ * copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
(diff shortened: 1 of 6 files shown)


zml2008:
So, my current thought for handling version differences is, we should have a 'feature flag' system, where things like 'emit legacy hover event' or 'emit uuid as ints' vs 'emit uuid as string' are toggleable options. We'd need to produce presets for compatibility with different game versions, plus probably a 'latest' and 'most compatible' levels.

Should this be game versions? Then how do we handle snapshots?

Alternately, we could mark these revisions by:

- [datapack versions](https://minecraft.wiki/w/Data_pack#Pack_format)
- protocol versions
- [data versions](https://minecraft.wiki/w/Data_version)

Thoughts on what makes most sense?

text-serializer-gson/src/main/java/net/kyori/adventure/text/serializer/gson/UUIDSerializer.java
```diff
@@ -0,0 +1,60 @@
...
+  public UUID read(final JsonReader in) throws IOException {
+    // int-array format was added in 23w40a, a pre for 1.20.3
+    if (in.peek() == JsonToken.BEGIN_ARRAY) {
+      in.beginArray();
+      final int v0 = in.nextInt();
+      final int v1 = in.nextInt();
+      final int v2 = in.nextInt();
+      final int v3 = in.nextInt();
+      in.endArray();
+      return new UUID((long) v0 << 32 | ((long) v1 & 0xffffffffl), (long) v2 << 32 | ((long) v3 & 0xffffffffl));
```
kashike:
```suggestion
      final int msb0 = in.nextInt();
      final int msb1 = in.nextInt();
      final int lsb0 = in.nextInt();
      final int lsb1 = in.nextInt();
      in.endArray();
      return new UUID((long) msb0 << 32 | ((long) msb1 & 0xffffffffl), (long) lsb0 << 32 | ((long) lsb1 & 0xffffffffl));
```

kashike (dismissed):

KingOfSquares:
> So, my current thought for handling version differences is, we should have a 'feature flag' system, where things like 'emit legacy hover event' or 'emit uuid as ints' vs 'emit uuid as string' are toggleable options. We'd need to produce presets for compatibility with different game versions, plus probably a 'latest' and 'most compatible' levels.
> 
> Should this be game versions? Then how do we handle snapshots?
> 
> Alternately, we could mark these revisions by:
> 
> * [datapack versions](https://minecraft.wiki/w/Data_pack#Pack_format)
> * protocol versions
> * [data versions](https://minecraft.wiki/w/Data_version)
> 
> Thoughts on what makes most sense?

I think using the protocol version is a good idea, seems like the most precise number and easiest to map to any point in releases/snapshots etc..
To my knowledge Bedrock uses the same system right(for protocol versions)? In case the bedrock serialiser or other bedrock specific stuff ever makes it in here

zml2008:
We still need to implement:

- [ ] handling for the `type` field
- [ ] NBT component serialization (or DFU integration)
- [ ] matching the strictness of Vanilla serialization

but I think it's worth merging this as-is just to have published snapshots that others can depend on.

PR #996, opened 2023-11-23
feature: Cancellable FlattenerListener

FxMorin:
Adds the ability to stop a flattener while its running. 
This allows developers to make implementations with early exit conditions.


zml2008 (approved):
Looks great, thanks for this! Just needs style fixes and it should be ready to go.

PR #1004, opened 2023-12-09
feat(api): Support representing translation args losslessly

zml2008:
wip, still needs:

- [x] collapse hierarchy of args to just one class
- [x] propogate down to serializer impls


zml2008:
Still debating, we rename the `args()` getter to `arguments()` to have a different return type, do we want to rename the setters as well? I feel like that would introduce API churn for no functional change, but it would improve consistency and clarity going forward. We could also opt to delay that change to a later date?

PR #1005, opened 2023-12-10
feat(api): Expand the supported resource pack operations for 1.20.3

zml2008:
This PR makes it easy to handle multiple resource packs, and adds a callback system to track pack application.

(note, this does change some API used in snapshots so it'd be a binary break for any current testers)

### DIFF
--- a/api/src/main/java/net/kyori/adventure/audience/Audience.java
+++ b/api/src/main/java/net/kyori/adventure/audience/Audience.java
@@ -41,6 +41,7 @@ import net.kyori.adventure.pointer.Pointered;
 import net.kyori.adventure.resource.ResourcePackInfo;
 import net.kyori.adventure.resource.ResourcePackInfoLike;
 import net.kyori.adventure.resource.ResourcePackRequest;
+import net.kyori.adventure.resource.ResourcePackRequestLike;
 import net.kyori.adventure.sound.Sound;
 import net.kyori.adventure.sound.SoundStop;
 import net.kyori.adventure.text.Component;
@@ -734,7 +735,33 @@ public interface Audience extends Pointered {
    * @see ResourcePackInfo
    * @since 4.15.0
    */
+  @ForwardingAudienceOverrideNotRequired
+  default void sendResourcePacks(final @NotNull ResourcePackRequestLike request) {
+    this.sendResourcePacks(request.asResourcePackRequest());
+  }
+
+  /**
+   * Sends a request to apply resource packs to this audience.
+   *
+   * <p>Multiple resource packs are only supported since 1.20.3. On older versions, all requests behave as if {@link ResourcePackRequest#replace()} is set to {@code true}.</p>
+   *
+   * @param request the resource pack request
+   * @see ResourcePackInfo
+   * @since 4.15.0
+   */
   default void sendResourcePacks(final @NotNull ResourcePackRequest request) {
+  }
+
+  /**
+   * Clear resource packs with the IDs used in the provided requests if they are present.
+   *
+   * @param request the request used to originally apply the packs
+   * @since 4.15.0
+   * @sinceMinecraft 1.20.3
+   */
+  @ForwardingAudienceOverrideNotRequired
+  default void removeResourcePacks(final @NotNull ResourcePackRequestLike request) {
+    this.removeResourcePacks(request.asResourcePackRequest());
   }
 
   /**
--- a/api/src/main/java/net/kyori/adventure/audience/Audience.java
+++ b/api/src/main/java/net/kyori/adventure/audience/Audience.java
@@ -721,12 +721,12 @@ public interface Audience extends Pointered {
    */
   @SuppressWarnings("checkstyle:MethodName")
   @ForwardingAudienceOverrideNotRequired
-  default void setResourcePack(final @NotNull ResourcePackRequestLike request, final @NotNull ResourcePackRequestLike@NotNull... others) {
+  default void setResourcePacks(final @NotNull ResourcePackRequestLike request, final @NotNull ResourcePackRequestLike@NotNull... others) {
     final ResourcePackRequest[] otherReqs = new ResourcePackRequest[others.length];
     for (int i = 0; i < others.length; i++) {
       otherReqs[i] = others[i].asResourcePackRequest();
     }
-    this.setResourcePack(request.asResourcePackRequest(), otherReqs);
+    this.setResourcePacks(request.asResourcePackRequest(), otherReqs);
   }
 
   /**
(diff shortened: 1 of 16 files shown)


api/src/main/java/net/kyori/adventure/resource/ResourcePackRequest.java
```diff
@@ -191,6 +196,16 @@ interface Builder extends AbstractBuilder<ResourcePackRequest>, ResourcePackRequ
...
 
+    /**
+     * Builds, computing a hash based on the provided URL.
+     *
+     * <p>The hash computation will perform a network request asynchronously, exposing the built request via the returned future.</p>
+     *
+     * @return a future providing the new resource pack request
+     * @since 4.15.0
+     */
+    @NotNull CompletableFuture<ResourcePackRequest> computeHashAndBuild();
```
Machine-Maker: Any use for providing your own executor optionally?

api/src/main/java/net/kyori/adventure/audience/Audience.java
```diff
@@ -708,24 +710,192 @@ default void openBook(final @NotNull Book book) {
...
+   * <p>Multiple resource packs are only supported since 1.20.3. On older versions, only the first pack will be sent.</p>
+   *
+   * @param request the first resource pack request
+   * @param others other requests
+   * @see ResourcePackRequest
+   * @since 4.15.0
+   */
+  @SuppressWarnings("checkstyle:MethodName")
+  @ForwardingAudienceOverrideNotRequired
+  default void setResourcePack(final @NotNull ResourcePackRequestLike request, final @NotNull ResourcePackRequestLike@NotNull... others) {
```
Machine-Maker: It does not make sense to accept more than one pack request for a "setResourcePack" method imo. If you want to have multiple, you use the send methods. set should just be a single one. Maybe making the name setResourcePacks would be better.

api/src/main/java/net/kyori/adventure/audience/Audience.java
```diff
@@ -708,24 +710,192 @@ default void openBook(final @NotNull Book book) {
...
+   * Sends resource pack requests to this audience, adding to any existing resource packs that might be present.
+   *
+   * <p>Multiple resource packs are only supported since 1.20.3. On older versions, this behaves identically to {@link #setResourcePack(ResourcePackRequest, ResourcePackRequest...)}.</p>
+   *
+   * @param request the resource pack request
+   * @param others other requests
+   * @see ResourcePackRequest
+   * @since 4.15.0
+   */
+  default void sendResourcePack(final @NotNull ResourcePackRequest request, final @NotNull ResourcePackRequest@NotNull... others) {
```
Machine-Maker: Can this method just delegate to the "send" methods with an empty callback? That way you don't have to implement multiple methods. You can make the empty callback some static constant that is available to implementations so they know they don't have to track it.

Machine-Maker (changes requested):
Should there be Iterable methods for sending/setting multiple as well?

api/src/main/java/net/kyori/adventure/resource/ResourcePackRequest.java
zml2008: Sure, that'd make sense I suppose.

api/src/main/java/net/kyori/adventure/audience/Audience.java
zml2008: The difference between set and send is, set is (clear + send), so why wouldn't it accept multiple. Making the method name plural would make sense I suppose.

api/src/main/java/net/kyori/adventure/audience/Audience.java
zml2008: eh, depending on the object identity of a no-op callback seems a bit iffy to me (unless we exposed it as API on `ResourcePackCallback` itself), but tbh as I think about it your suggestion makes more and more sense.

zml2008:
hm, looking at adding Iterable overloads, type erasure means we can't support all the individual types for each method. Right now, i feel like the best signature for implementations would be `thing(List<RPRequest>)`, where for API end users, the `thing(Iterable<? extends RPRequestLike>)` signature would make most sense -- but then we have to be careful about unboxing *Like instances pretty eagerly maybe?

zml2008:
Looking at Vanilla again, it appears that while the client will accept multiple packs, it only uses the last `required` and `prompt` values sent, not one per pack. Maybe we want to push those values up to `Request` from `Info`?

api/src/main/java/net/kyori/adventure/resource/ResourcePackInfo.java
```diff
@@ -0,0 +1,229 @@
...
+import net.kyori.adventure.audience.Audience;
+import net.kyori.adventure.builder.AbstractBuilder;
+import net.kyori.adventure.text.Component;
+import net.kyori.examination.Examinable;
+import org.jetbrains.annotations.Contract;
+import org.jetbrains.annotations.NotNull;
+import org.jetbrains.annotations.Nullable;
+
+/**
+ * Represents a resource pack request that can be sent to players.
```
Machine-Maker: Don't think we should mention "request" in this type, now that info is separated from the request.

api/src/main/java/net/kyori/adventure/resource/ResourcePackInfo.java
Machine-Maker, line 1: All, the mentions of "request" probably need to be handled.

Machine-Maker (dismissed):

api/src/main/java/net/kyori/adventure/audience/Audience.java
```diff
@@ -708,24 +712,114 @@ default void openBook(final @NotNull Book book) {
   // ------------------------
 
   /**
-   * Sends a resource pack request to this audience.
+   * Sends a request to apply resource packs to this audience.
+   *
+   * <p>Multiple resource packs are only supported since 1.20.3. On older versions, all requests behave as if {@link ResourcePackRequest#replace()} is set to {@code true}.</p>
```
kezz, line 717: I wonder if we want to do something different here. Nobody is realistically going to want to set a bunch of packs with replace to true. I think we should just emulate this behaviour by only applying the final/first pack instead (effectively the same, but without the pointless network load/etc.

electronicboy: I'd imagine that the only place in which replace would be true is for the first pack in the stack, "reset and apply these", assuming that the client supports that behavior (I would hope so)
as for legacy versions, I think that adventure should just recommend for a behavior here, i.e. impls should only care about grabbing and sending the last pack in the request?

assuming I'm not barking up the wrong tree for the discussion here

zml2008: `replace` applies to the overall request, not to individual packs within the request here -- that's the distinction between `ResourcePackInfo` and `ResourcePackRequest` really.

But yeah, the comments about legacy versions really only apply to whoever is masochistic enough to implement the resource pack API in -platform, modern impls can just ignore that bit.

PR #1012, opened 2023-12-19
minimessage: Fix exception with single quote as tag part

### LINKED ISSUE #1011: `MiniMessage::deserialize` throws `StringIndexOutOfBoundsException` when double quote is unclosed in tag argument (in lenient mode)
The [documentation](https://docs.advntr.dev/minimessage/api.html#error-handling) states that `MiniMessage::deserialize` should never throw by default, but when a double quote is left unclosed in the argument list of a tag (e.g. `<hover:show_text_:">`), a `StringIndexOutOfBoundsException` is thrown.

Here is the relevant stacktrace:
```
	at java.base/java.lang.String.substring(String.java:2709)
	at net.kyori.adventure.text.minimessage.internal.parser.TokenParser.unescape(TokenParser.java:620)
	at net.kyori.adventure.text.minimessage.internal.parser.node.TagPart.unquoteAndEscape(TagPart.java:109)
	at net.kyori.adventure.text.minimessage.internal.parser.node.TagPart.<init>(TagPart.java:53)
	at net.kyori.adventure.text.minimessage.internal.parser.match.StringResolvingMatchedTokenConsumer.accept(StringResolvingMatchedTokenConsumer.java:90)
	at net.kyori.adventure.text.minimessage.internal.parser.TokenParser.parseString(TokenParser.java:247)
	at net.kyori.adventure.text.minimessage.internal.parser.TokenParser.resolvePreProcessTags(TokenParser.java:112)
	at net.kyori.adventure.text.minimessage.MiniMessageParser.parseToTree(MiniMessageParser.java:195)
```

rymiel:
The parser tries to un-quote tag parts surrounded by quotes, it does so by checking if the first character of the tag part is a quote and the last character is as well. However, things break if that first and last character are actually the same character! A StringIndexOutOfBoundsException is thrown later on.

This patch makes it so tag parts consisting of a single quote aren't tried to be un-quoted, they stay as is

Fixes https://github.com/KyoriPowered/adventure/issues/1011

### DIFF
--- a/text-minimessage/src/main/java/net/kyori/adventure/text/minimessage/internal/parser/node/TagPart.java
+++ b/text-minimessage/src/main/java/net/kyori/adventure/text/minimessage/internal/parser/node/TagPart.java
@@ -106,6 +106,11 @@ public final class TagPart implements Tag.Argument {
       endIndex--;
     }
 
+    if (startIndex > endIndex) {
+      // We were given only a single quote that doesn't terminate, we can't unescape it
+      return text.substring(start, end);
+    }
+
     return TokenParser.unescape(text, startIndex, endIndex, i -> i == firstChar || i == TokenParser.ESCAPE);
   }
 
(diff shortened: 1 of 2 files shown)


zml2008 (approved):

zml2008:
thank you <3

Issue #236, opened 2020-12-13
Adding a function to replace strings in the textreplacer


Foorcee:
Hello,
I would find it very handy if you had the possibility to replace strings in the text replacer as well, e.g. the action in click events. With hover events it already works. Currently I have to iterate over all components again and replace the action if necessary.

I would be very happy if you could find a solution for this.

Phoenix616:
What's the general opinion on how this should be implemented and exposed in the API? Seeing as click events don't use components the hover event logic in the TextReplacementRenderer can't be used in a similar way and a separate method specific to the click events would be necessary.

Also should this still be part of the TextReplacementRenderer at all or completely separate? I personally would kinda expect it to still be in that API as well as the Component#replaceText utility method (that's how I noticed this shortcoming initially) but maybe there are better ideas/differing opinions on that.

zml2008:
In about 5 minutes, I've done up a quick implementation of the plain-string replacement using existing API:

```java
          Component.textOfChildren(component).toBuilder().mapChildrenDeep(x -> {
            if (x.clickEvent() != null) {
              return (BuildableComponent<?, ?>) x.clickEvent(ClickEvent.clickEvent(x.clickEvent().action(), /* replaced */ x.clickEvent().value()));
            } else {
              return x;
            }
          });
```

There are some nasty casts that hurt ergonomics here (which can hopefully be resolved for 5.0.0), but imo since there really isn't the same complexity that exists with replacing within styled content, it's not worth trying to shove into the existing replacement systems.

If someone refines an implementation of click event replacement, we might accept some elements as a PR, but please talk to us before spending much time on that.

PR #1027, opened 2024-01-12
feat(api): replace text in ClickEvent value too

f4n74z14:
Hi! 👋

While developing a plugin, it became necessary for me to replace the text in the ClickEvent value. The test provides a case of using this feature.

I also refactored the code a bit because the function turned out to be very complicated.

### DIFF
--- a/api/src/main/java/net/kyori/adventure/text/TextReplacementRenderer.java
+++ b/api/src/main/java/net/kyori/adventure/text/TextReplacementRenderer.java
@@ -29,6 +29,8 @@ import java.util.function.BiFunction;
 import java.util.regex.MatchResult;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
+
+import net.kyori.adventure.text.event.ClickEvent;
 import net.kyori.adventure.text.event.HoverEvent;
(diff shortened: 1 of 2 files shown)


kezz (changes requested):
I'm gonna go ahead and say no to this. Click event values aren't components, and this entire mechanism is pretty specifically designed for components.

Perhaps this would be okay if it was an opt-in option on the replacement config, with a setting to define how the component is turned into a raw string? Would be interested to hear what others think.

However, even then, I'd be tempted to just say no as there are already existing tools like MiniMessage that are purpose built for this kind of replacement.

zml2008:
I agree -- I can see a place for some sort of partially-integrated mechanism so any pure-text replacements can be applied to both click event values and components in the tree, but kezz covered my issues pretty well. See also #236.

(also, please remember when PR'ing to create the PR from a feature branch rather than `main/4` -- it makes it easier for us to work with locally). I'm going to close this PR due to the branch issues, but if you want to look at adding separate options to the replacement config, feel free to open another PR from the proper branch.

f4n74z14:
Fair enough! I'll have to think about switching to MiniMessage placeholders.

PR #1035, opened 2024-02-10
add component decoder

### LINKED ISSUE #1034: Add ComponentDecoder
https://jd.advntr.dev/api/4.15.0/net/kyori/adventure/text/serializer/ComponentSerializer.html currently inherits from ComponentEncoder, which is nice, however the deserialising methods are not inherited from a type like e.g. ComponentDecoder.

This is unfortunate as such a type would allow platforms that only require deserialisation from an instance to skip ComponentSerialiser and properly type such instance as ComponentDecoder.

yannicklamprecht:
resolves https://github.com/KyoriPowered/adventure/issues/1034

### DIFF
--- a/api/src/main/java/net/kyori/adventure/text/serializer/ComponentDecoder.java
+++ b/api/src/main/java/net/kyori/adventure/text/serializer/ComponentDecoder.java
@@ -1,0 +1,80 @@
+/*
+ * This file is part of adventure, licensed under the MIT License.
+ *
+ * Copyright (c) 2017-2023 KyoriPowered
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in all
+ * copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+ * SOFTWARE.
+ */
+package net.kyori.adventure.text.serializer;
+
+import net.kyori.adventure.text.Component;
+import org.jetbrains.annotations.Contract;
+import org.jetbrains.annotations.NotNull;
+import org.jetbrains.annotations.Nullable;
+
+/**
+ * A {@link Component} decoder, which provides deserialization, but without serialization.
+ *
+ * <p>For both serialization and deserialization, use {@link ComponentSerializer}</p>
+ *
+ * @param <S> the serialized type
+ * @param <O> the output component type
+ * @since 4.16.0
+ */
+public interface ComponentDecoder<S, O extends Component> {
+  /**
+   * Deserialize a component from input of type {@code S}.
+   *
+   * @param input the input
+   * @return the component
+   * @since 4.16.0
+   */
+  @NotNull O deserialize(final @NotNull S input);
+
+  /**
+   * Deserialize a component from input of type {@code S}.
+   *
+   * <p>If {@code input} is {@code null}, then {@code null} will be returned.</p>
+   *
+   * @param input the input
+   * @return the component if {@code input} is non-null, otherwise {@code null}
+   * @since 4.16.0
+   */
+  @Contract(value = "!null -> !null; null -> null", pure = true)
+  default @Nullable O deserializeOrNull(final @Nullable S input) {
+    return this.deserializeOr(input, null);
+  }
+
+  /**
+   * Deserialize a component from input of type {@code S}.
+   *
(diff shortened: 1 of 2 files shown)


api/src/main/java/net/kyori/adventure/text/serializer/ComponentSerializer.java
```diff
@@ -89,6 +91,7 @@ public interface ComponentSerializer<I extends Component, O extends Component, R
    * @since 4.7.0
    */
   @Contract(value = "!null, _ -> !null; null, _ -> param2", pure = true)
+  @Override
```
kashike, line 94: `return ComponentDecoder.super.deserializeOr(input, fallback);`

api/src/main/java/net/kyori/adventure/text/serializer/ComponentSerializer.java
```diff
@@ -74,6 +75,7 @@ public interface ComponentSerializer<I extends Component, O extends Component, R
    * @since 4.8.0
    */
   @Contract(value = "!null -> !null; null -> null", pure = true)
+  @Override
   default @Nullable O deserializeOrNull(final @Nullable R input) {
     return this.deserializeOr(input, null);
```
kashike:
```suggestion
    return ComponentDecoder.super.deserializeOr(input, null);
```

kashike (changes requested):

kashike (approved):

api/src/main/java/net/kyori/adventure/text/serializer/ComponentDecoder.java
```diff
@@ -0,0 +1,80 @@
...
+/**
+ * A {@link Component} decoder, which provides deserialization, but without serialization.
+ *
+ * <p>For both serialization and deserialization, use {@link ComponentSerializer}</p>
+ *
+ * @param <S> the serialized type
+ * @param <O> the output component type
+ * @since 4.16.0
+ */
+public interface ComponentDecoder<S, O> {
```
emilyy-dev:
```suggestion
public interface ComponentDecoder<S, O extends Component> {
```

api/src/main/java/net/kyori/adventure/text/serializer/ComponentDecoder.java
```diff
@@ -0,0 +1,80 @@
+/*
+ * This file is part of adventure, licensed under the MIT License.
+ *
+ * Copyright (c) 2017-2023 KyoriPowered
```
emilyy-dev, line 4:
```suggestion
 * Copyright (c) 2017-2024 KyoriPowered
```

api/src/main/java/net/kyori/adventure/text/serializer/ComponentDecoder.java
```diff
@@ -0,0 +1,80 @@
...
+   * Deserialize a component from input of type {@code R}.
+   *
+   * @param input the input
+   * @return the component
+   * @since 4.16.0
+   */
+  @NotNull O deserialize(final @NotNull S input);
+
+  /**
+   * Deserialize a component from input of type {@code R}.
```
emilyy-dev:
```suggestion
   * Deserialize a component from input of type {@code S}.
```

api/src/main/java/net/kyori/adventure/text/serializer/ComponentDecoder.java
```diff
@@ -0,0 +1,80 @@
...
+ *
+ * <p>For both serialization and deserialization, use {@link ComponentSerializer}</p>
+ *
+ * @param <S> the serialized type
+ * @param <O> the output component type
+ * @since 4.16.0
+ */
+public interface ComponentDecoder<S, O extends Component> {
+  /**
+   * Deserialize a component from input of type {@code R}.
```
emilyy-dev:
```suggestion
   * Deserialize a component from input of type {@code S}.
```

api/src/main/java/net/kyori/adventure/text/serializer/ComponentDecoder.java
```diff
@@ -0,0 +1,80 @@
...
+   * @return the component if {@code input} is non-null, otherwise {@code null}
+   * @since 4.16.0
+   */
+  @Contract(value = "!null -> !null; null -> null", pure = true)
+  default @Nullable O deserializeOrNull(final @Nullable S input) {
+    return this.deserializeOr(input, null);
+  }
+
+  /**
+   * Deserialize a component from input of type {@code R}.
```
emilyy-dev:
```suggestion
   * Deserialize a component from input of type {@code S}.
```

api/src/main/java/net/kyori/adventure/text/serializer/ComponentSerializer.java
emilyy-dev: I disagree with this suggestion, if an implementation overrides `deserializeOr` (and/or the other methods where `ComponentDecoder.super.blah` is called) the call will invoke the default method in the superinterface, not the overridden one as was the case before. IMO the default methods in ComponentSerializer inherited from ComponentDecoder can be removed entirely, it is not a breaking change, or alternatively left as-is

emilyy-dev (changes requested):

api/src/main/java/net/kyori/adventure/text/serializer/ComponentSerializer.java
yannicklamprecht: Would personally prefer to remove them. Same should be done with the one from Encoder.
Except the one with the spelling mistake that gets removed in 5.0.0
What do you think of that?


Only reason I can think of, where the removal can break, is the use of reflection and the getDeclaredMethod call.

zml2008: what about:

```suggestion
    return ComponentDecoder.super.deserializeOrNull(input);
```

(having not looked closely at the context)

I'd like to keep the overrides, just so we can maintain the differing @since JD on them for people trying to work with older versions.

api/src/main/java/net/kyori/adventure/text/serializer/ComponentDecoder.java
yannicklamprecht: Will revert that change, this should be done separately using license.txt and license-header.txt

api/src/main/java/net/kyori/adventure/text/serializer/ComponentSerializer.java
yannicklamprecht: If you all agreed to one solution I'm happy to change it that way.

emilyy-dev: That should be fine, yeah, each one `super`ing to itself

zml2008 (approved):

Issue #1038, opened 2024-02-23
Add brightness/lightness and a wrap around tag to mini message <rainbow> tag


Antritus:
Add more tags to the minimessage rainbow tag ``<rainbow>``

This might make the tags too confusing so it's only a suggestion.

Brightness tag would be a double/float to control between blackness and whiteness
Brightness 0 = black
Brightness 1 = white
brightness between 0 and 1 would show different brightness of hex colors
```
<rainbow:0.5f>This is a rainbow string with 0.5f brightness
<rainbow:0.9f>This is a rainbow string with 0.9f brightness. This would be pretty whitish
<rainbow:0.1>This would be rainbow string with really dark almost black
```

Wrap around tag would be like the current number tag in rainbow
Wrap around would just set the maximum string length between red and purple.
This ignores spaces so it doesn't wrap around them
This could be possible way to display it..
```
<rainbow:wrap:10>This tag would split after "u" in the word "would" and redo the rainbow after it 
<rainbow:wrap:5>This tag would split after "t" in "tag"
```

What combine strings could look like
```
<rainbow;!5:0.5f:wrap:4> would make it display the rainbow reversed starting from the 5th color, with 0.5f brightness and wrapping every 4 characters
<rainbow;0.5f:wrap:10> would make it display the rainbow with 0.5f brightness and wrapping every 10 characters
<rainbow;!:wrap:7> would make it display the rainbow reversed and wrapping every 4 characters
```

Most of these ideas are from this "Skript" thread which had them. https://forums.skunity.com/threads/colorutils-1-16-skript-2-6.12134/

![upload_2020-12-22_17-16-23](https://github.com/KyoriPowered/adventure/assets/102650288/e91bd543-c3be-411e-8fd3-ce82c5f57af3)

zml2008:
I think some more advanced formatting options would be great, but the syntax starts getting pretty complex. I think we'd need to look at adding some sort of named arguments language feature to MM in order to make these sorts of features less convoluted.

PR #1041, opened 2024-02-25
fix(text-minimessage): Handle larger raninbow phases correctly

### LINKED ISSUE #1040: MM rainbow phase problems
I saw that a big phase creates problems: https://webui.advntr.dev/?x=bulbzQN4uX
Also it seems like range values are [1-10] 

https://github.com/KyoriPowered/adventure/assets/56083016/ef91830e-0840-4d22-9881-13c026e01e67

I'd like to achieve something like this: (minedown)

https://github.com/KyoriPowered/adventure/assets/56083016/c52803b5-eef2-4cf2-8326-a0786f067c7d

I think the phase range should be bigger, now you can only have 10 different phases.

zml2008:
Fixes GH-1040


kezz (approved):
LGTM. I do think it would be nice to be able to configure the division part but honestly we should probably wait for some sort of resolution to #1038 before we start adding more complexity to the rainbow tag.

PR #1062, opened 2024-04-03
Changed gson version

Keylessboi:
Changed gson version to prevent a deserialization bug
[CVE](https://devhub.checkmarx.com/cve-details/CVE-2022-25647/)


zml2008:
We want to maintain compatibility with environments using older MC versions which ship old gson. Depending on modern MC will pull in the newer gson anyways, and standalone projects can bump it based on their own needs.

PR #1063, opened 2024-04-05
Gradient API (v2)

thiccaxe:
Addresses some of the main concerns of my previous rendition, mainly separating the concept of a gradient and the interpolation of colors. 

By not requiring colors to be a TextColor the API interface is greatly simplified (far less conversions from rgb and back). However, this requires an additional step on the user's end to convert from whatever color space they are using (say, HSV, or perhaps, (ok)LAB)  

I feel that this interface is more versatile and easier for users to extend on their end. 

- [x] gradient API  
- [ ] update minimessage to use this API
  - [ ] gradients
    - [ ] stops? 
  - [ ] rainbows
- [X] color spaces
  - [X] RGB
  - [X] HSV (need to verify implementation)
  - [ ] OkLab
  - [ ] OkLCH
- [ ] tests
-   [ ] gradient with stops
  - [X] RGB
  - [ ] HSV
  - [ ] OKLAB
  - [ ] OKLCH
  - [ ] minimessage

### DIFF
--- a/api/src/main/java/net/kyori/adventure/text/format/gradient/ColorSpaceInterpolator.java
+++ b/api/src/main/java/net/kyori/adventure/text/format/gradient/ColorSpaceInterpolator.java
@@ -1,0 +1,176 @@
+/*
+ * This file is part of adventure, licensed under the MIT License.
+ *
+ * Copyright (c) 2017-2024 KyoriPowered
+ *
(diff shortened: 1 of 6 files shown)


mudkipdev:
Is this still being considered?

thiccaxe:
I don't plan on continuing this, but I suppose someone else could

PR #1076, opened 2024-05-19
feat(translation): Add recursive check on translatable

f4bii:


Machine-Maker:
If this is a valid path to go down, there are other places within the Component structure that would also have to be translated. See TranslatableComponentRenderer/AbstractComponentRenderer to see how that is handled during a "render".

kezz:
I'm going to go ahead and close this PR due to both a lack of activity but also due to the fact that this is the incorrect place to perform such recursive rendering. This instead should be done solely in the TranslatableComponentRenderer as the translate methods on Translator should only be used to directly translate the key/component being provided. By translating the children directly, we're also encountering DRY problems as rendering of children is already done by the TranslatableComponentRenderer.

If this PR is reopened in the future, use cases should also be provided so we can judge how much of a requirement this recursive functionality is.

PR #1079, opened 2024-05-31
feat(minimessage): Pride tag

kezz:
It's surprising how often I have to hand type pride gradients. Anyway, happy pride gamers 🏳️‍🌈

image

### DIFF
--- a/text-minimessage/src/main/java/net/kyori/adventure/text/minimessage/tag/standard/GradientTag.java
+++ b/text-minimessage/src/main/java/net/kyori/adventure/text/minimessage/tag/standard/GradientTag.java
@@ -45,7 +45,7 @@ import org.jetbrains.annotations.Range;
  *
  * @since 4.10.0
  */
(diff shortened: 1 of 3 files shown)


zml2008 (approved):

mudkipdev:
Do you think it's also possible to add a pride tag which uses the raw colors instead of smooth gradients

zml2008:
I think that'd be a great idea for pride next year :3 (as in, that sounds possible but also would take more time than simply repurposing the gradient tag)

kezz:
Wouldn't be a hard tag to make, would be nice to make it generic then repurpose it for pride like this tag. Any name ideas for the tag?

5RoD (commented):
what the fuck is this

carlosleondev:
where is the flag of heterosexuals?

iAtog:
> where is the flag of heterosexuals?

There is no such thing as heterosexuals

PR #1084, opened 2024-06-08
NBT Component Serializer

### LINKED ISSUE #995: NBT Component serializer
This should support the existing 1.20.3+ structure of components, writing to/from a BinaryTag of adventure-nbt.

Codestech1:
Adds a support for serializing components as NBT Binary Tags.

Resolves #995


Codestech1:
The serialization works now, but i need to clean it up, add backwards compatibility (old nbt), add javadocs and deserialization. But the hardest part with understanding how everything realted to this (in both adventure and Minecraft) is done.

Codestech1:
Component deserialization works properly (tested with a large component with many children and styling), component serialization however doesn't serialize lists of components (children, etc.) as anything else than compounds. I'll check if there's a way to make it check if all types of components can be serialized as a string without looping through the entire list twice (one time for checks, second time for creating the list).

What's left (let me know if i forgot anything):
- JUnit tests
- Services (for custom serializer imlementation, like the gson serializer)
- Cleanup (maybe split the serialization process to more classes?)
- More serializer options
- Make sure that legacy tags in hover event are serialized properly (I can't do it now, because i cannot even make them work through a `/tellraw` command in Minecraft)
- Javadocs

Codestech1:
Another TODO:
The hover event serializer should serialize legacy `show achievement` hover events with `value` key instead of `contents` key

Codestech1:
I forgot to add tests for hover event and click event, i'm adding them right now

Codestech1:
Ok, it's ready now, i think

text-serializer-nbt/src/main/java/net/kyori/adventure/text/serializer/nbt/NBTSerializerOptions.java
```diff
@@ -0,0 +1,134 @@
...
+import net.kyori.option.Option;
+
+/**
+ * Options that can apply to {@linkplain NBTComponentSerializer NBT serializers}.
+ *
+ * <p>See serializer documentation for specific details on which flags are supported.</p>
+ *
+ * @since 4.18.0
+ */
+public final class NBTSerializerOptions {
```
zml2008, line 39: These options are redundant -- they were added to the json serializer because the json ormat has changed over the years. the NBT version of the format is new as of 1.20.5 so doesn't need the extra complexity of backwards compat

text-serializer-nbt/src/main/java/net/kyori/adventure/text/serializer/nbt/NBTComponentSerializer.java
```diff
@@ -0,0 +1,190 @@
...
+    @NotNull Builder editOptions(final @NotNull Consumer<OptionState.Builder> optionEditor);
+
+    /**
+     * Sets whether the serializer should downsample hex colors to named colors.
+     *
+     * @param emit true if the serializer should downsample hex colors to named colors, false otherwise
+     * @return this builder
+     * @since 4.18.0
+     */
+    default @NotNull Builder emitRgb(final boolean emit) {
```
zml2008: don't duplicate methods for all these random options

text-serializer-nbt/src/main/java/net/kyori/adventure/text/serializer/nbt/ClickEventSerializer.java
```diff
@@ -0,0 +1,113 @@
...
+  private static final String CLICK_EVENT_OPEN_FILE = "open_file";
+  private static final String CLICK_EVENT_RUN_COMMAND = "run_command";
+  private static final String CLICK_EVENT_SUGGEST_COMMAND = "suggest_command";
+  private static final String CLICK_EVENT_CHANGE_PAGE = "change_page";
+  private static final String CLICK_EVENT_COPY_TO_CLIPBOARD = "copy_to_clipboard";
+
+  private ClickEventSerializer() {
+  }
+
+  static @Nullable ClickEvent deserialize(@NotNull CompoundBinaryTag tag) {
```
zml2008: isn't there an `Index` for these? how do the other serializers work?

text-serializer-nbt/src/main/java/net/kyori/adventure/text/serializer/nbt/HoverEventSerializer.java
```diff
@@ -0,0 +1,207 @@
...
+
+import java.util.HashMap;
+import java.util.Map;
+import java.util.UUID;
+
+final class HoverEventSerializer {
+
+  private static final String HOVER_EVENT_ACTION = "action";
+  private static final String HOVER_EVENT_CONTENTS = "contents";
+  @Deprecated
```
zml2008: none of these exist in the nbt pormat

zml2008 (changes requested):
please review what the game actually does, you've dragged over so much legacy baggage that'd be completely unnecessary if you actually thought things through

Codestech1:
I've added a possibility to (de)-serialize a style without a component, because it's useful for new chat type registry in Minecraft. For example Minestom uses adventure styles to (de)-serialize the chat type styles.

text-serializer-nbt/src/main/java/net/kyori/adventure/text/serializer/nbt/StyleSerializer.java
```diff
@@ -98,11 +98,7 @@ static void serialize(@NotNull Style style, CompoundBinaryTag.@NotNull Builder b
     TextColor color = style.color();
 
     if (color != null) {
-      if (color instanceof NamedTextColor) {
-        builder.putString(COLOR, color.toString());
-      } else if (serializer.flags().value(NBTSerializerOptions.EMIT_RGB)) {
-        builder.putString(COLOR, color.asHexString());
-      }
+      builder.putString(COLOR, color instanceof NamedTextColor ? color.toString() : color.asHexString());
```
RealBauHD: TextColor#toString does the same

Codestech1:
After ~ 3 months i can say that it's stable. I've not noticed any issues with the serializer. I was using it in my server software - Jet - for that time. I tested all functionality, including item data components etc.

kezz (approved):
Generally speaking I'm happy with this. My only _slight_ concern is that there's a lot of unchecked casts. I do wonder if we might be better off handling all of these and throwing nicer error messages (e.g. `Expected <tag> to be of type Byte, was String`)?

I don't think this is a blocker on a merge though, happy for this to go through and leave an issue for us to come back to that later.

RealBauHD (approved):

Codestech1:
> Generally speaking I'm happy with this. My only _slight_ concern is that there's a lot of unchecked casts. I do wonder if we might be better off handling all of these and throwing nicer error messages (e.g. `Expected <tag> to be of type Byte, was String`)?
> 
> I don't think this is a blocker on a merge though, happy for this to go through and leave an issue for us to come back to that later.

Yeah, but in my opinion that should be done in adventure nbt instead.

kezz:
How could that be done in adventure-nbt?

Codestech1:
> How could that be done in adventure-nbt?

Making method such as `asStringOrThrow` would be useful.

kezz:
I don't think that makes sense for an API.

MelonHell (approved):

MelonHell:
looks good, hopefully this will be added in the near future

AlexProgrammerDE:
Any update on this PR? There haven't been any commits in over eight months. Is this still happening?

Codestech1:
> Any update on this PR? There haven't been any commits in over eight months. Is this still happening?

The PR was complete, however nobody has merged it. Now it's not, the only thing missing is the shadow colors feature from new Minecraft versions, i'm going to add it soon.

Codestech1:
What's left:
- [x] Implement deserializing components from lists
- [x] Ensure that serialization codecs are the same as in Minecraft
- [x] Rewrite and make more "restrictive" JUnit tests
- [x] Ensure that the code is consistent
- [x] Do some tests comparing outputs of vanilla dfu codec (with NbtOps) with this serializer
- [x] Add versioned option state
- [x] Add support for deserialization of some legacy formats (as it's still valid for deserialization of modern vanilla serializer)
- [x] Replace Key serialization with a new serializer class

Codestech1:
I added support for deserializing legacy hover events (those where contents are serialized as text components) as they were still valid for vanilla component serializer while deserializing in versions with NBT component serializing and camel case hover events, however serialization is not supported. I did it like that since I want to make this serializer work the same as the vanilla one, so developers can easily switch to this serializer.

Codestech1:
I'd say this serializer is done. It's definitely better than the previous one, it also follows new approach used in the GSON serializer - vanilla-like output. There is one thing left (output comparisons with the dfu serializer) in a list posted by me few comments above, however I'm going to do that in a different repo and it's more like a "proof" that the serializer is stable and world properly and it shouldn't block the PR from being marked as "ready".

Codestech1:
Oh, I just realized that there are checkstyle errors, I'm going to fix them soon

Codestech1:
All of the checkstyle errors have been fixed.

Codestech1:
Btw, you probably want to squash-merge this PR since it has over 80 commits ( :open_mouth:  ).

Codestech1:
The latest commit fixes the order in which `TagStringWriter` writes compound tag entries. It's in this PR as I noticed it during comparison between this serializer and dfu serializer. I hope that it's not a blocker on a merge.

Codestech1:
> I'd say this serializer is done. It's definitely better than the previous one, it also follows new approach used in the GSON serializer - vanilla-like output. There is one thing left (output comparisons with the dfu serializer) in a list posted by me few comments above, however I'm going to do that in a different repo and it's more like a "proof" that the serializer is stable and world properly and it shouldn't block the PR from being marked as "ready".

Tests comparing 23w40a (first Minecraft version serializing components in NBT) serializer and this serializer's output are running successfully (https://github.com/Codestech1/nbt-dfu-comparison). Now I only need to copy-paste these tests, replace the testing version to 1.21.7, remove legacy features from testing components and add tests for new features.

Codestech1:
Ok, now the dfu comparison tests are complete (they're available [here](https://github.com/Codestech1/nbt-dfu-comparison)). All issues have been fixed, the output is the same as in vanilla, so there is really no need to test it in-game. I'd say it's stable enough, hopefully this gets merged soon, the PR is now **fully** completed.

All you need to run these tests is to publish adventure (on the branch associated with this PR) to maven local.

MelonHell:
Hopefully, this branch will be merged by the end of the year :D

kezz:
Superceded by https://github.com/PaperMC/adventure/pull/1404

PR #1123, opened 2024-10-30
Optionally emit type field for JSON serializers

### LINKED ISSUE #1121: Optionally emit `type` field in JSON serializers

kezz:
Closes #1121


zml2008:
I get the goal of emitting the type field to have some initial support, but it feels kinda icky to have the types be just part of the gson serializer's giant if tree, and not even enforced when parsing. Is this level of support super essential for you or would you be fine pushing this to a later release to develop further? Cuz like the whole structural/keyed typing hybrid already lol

PR #1147, opened 2024-12-23
shadow gradient and rainbow tags

zml2008:
Done as a quick tech demo, would need a bit of actual design work and cleanup if we want to actually incorporate this into MM

`<gradow:red:blue>hello world`

image

`<shainbow>what kind of day is <aqua>this?`

image


Andre601:
Personal opinion, but I feel like `shadient` would be a better name than `gradow`. Tho, that's more a personal preference (maybe allow it as an alias?).

oliviathevampire:
What would this look like with a normal gradient and a shadow gradient?

PR #1152, opened 2025-01-04
chore(deps): Bump kotlin language + plugin version

zml2008:
The newer Kotlin Gradle plugin no longer supports maintaining our existing language version target. To keep our build toolchain up-to-date, we've had to bump the target language version to the minimum supported by the newest Kotlin plugin. We expect these sorts of bumps to continue for the forseeable future because of overly tight coupling between the Kotlin gradle plugin and the Kotlin compiler.

This may cause source compatibility breaks for some users of `adventure-extra-kotlin`. Those users should consider using a programming language that cares about stability if this is an issue.


lucko:
> Those users should consider using a programming language that cares about stability if this is an issue.

😆

PR #1166, opened 2025-02-24
feat(api): Make ComponentLike extends ComponentBuilderApplicable

indyteo:
This PR make the `ComponentLike` interface extends `ComponentBuilderApplicable`.

The main advantage of doing so is to allow using `ComponentLike`s in the `LinearComponents#linear` method. Imagine something in the lines of:
```java
LinearComponent.linear(NamedTextColor.YELLOW, player, NamedTextColor.AQUA, text(" just obtained "), reward, text("!"));
```
With `player` and `reward` being instances of a class that declares itself `ComponentLike`. It would be very clear syntax! Actually, we need to add `.asComponent()` so it takes away unnecessarily a bit of readability, I think.

Overall, I feel this serves well the purpose of `LinearComponents#linear` being to reduce the verbosity of declaring a succession of components.

Side note about the [`FunctionalInterfaceMethodChanged` warning](https://errorprone.info/bugpattern/FunctionalInterfaceMethodChanged), I suppressed it after discussing about it on Discord [here](https://discord.com/channels/289587909051416579/1342377788266512415/1343625412600791112).

Thanks!

### DIFF
--- a/api/src/main/java/net/kyori/adventure/text/ComponentLike.java
+++ b/api/src/main/java/net/kyori/adventure/text/ComponentLike.java
@@ -39,7 +39,7 @@ import static java.util.Objects.requireNonNull;
  * @since 4.0.0
  */
 @FunctionalInterface
-public interface ComponentLike {
+public interface ComponentLike extends ComponentBuilderApplicable {
   /**
    * Converts a list of {@link ComponentLike}s to a list of {@link Component}s.
    *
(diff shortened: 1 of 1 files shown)


kezz (approved):
Thanks for the PR!

kashike:
I'm unsure if it's best to do this, or adapt `LinearComponents#linear` to accept more types.

kezz:
The only way we'd be able to do this (allow usage of ComponentLike in LinearComponents alongside style elements) is allowing an `Object...` accepting method which is ugly.

indyteo:
> The only way we'd be able to do this (allow usage of ComponentLike in LinearComponents alongside style elements) is allowing an `Object...` accepting method which is ugly.

Technically, we could add another interface above to both `ComponentLike` and `ComponentBuilderApplicable`, and make the `LinearComponents#linear` function accepts this new super-interface instead. But I feel this is already kind of the purpose of `ComponentBuilderApplicable`, hence why I made it this way

Strokkur424 (changes requested):
Needs a rebase for `main/5`

PR #1168, opened 2025-02-28
1.21.5 component changes

### LINKED ISSUE #1195: Support flattened HoverEvent structure
for 1.21.5

Gerrygames:
things that need to be looked at:
- [x] show_entity field renames
- [x] change_page is now an int (or number?)
- [x] order of checks when deserializing hover events

### DIFF
--- a/text-serializer-gson/src/main/java/net/kyori/adventure/text/serializer/gson/ShowEntitySerializer.java
+++ b/text-serializer-gson/src/main/java/net/kyori/adventure/text/serializer/gson/ShowEntitySerializer.java
@@ -33,21 +33,26 @@ import java.util.UUID;
 import net.kyori.adventure.key.Key;
 import net.kyori.adventure.text.Component;
 import net.kyori.adventure.text.event.HoverEvent;
+import net.kyori.adventure.text.serializer.json.JSONOptions;
+import net.kyori.option.OptionState;
 import org.jetbrains.annotations.Nullable;
 
 import static net.kyori.adventure.text.serializer.commons.ComponentTreeConstants.SHOW_ENTITY_ID;
 import static net.kyori.adventure.text.serializer.commons.ComponentTreeConstants.SHOW_ENTITY_NAME;
 import static net.kyori.adventure.text.serializer.commons.ComponentTreeConstants.SHOW_ENTITY_TYPE;
+import static net.kyori.adventure.text.serializer.commons.ComponentTreeConstants.SHOW_ENTITY_UUID;
 
 final class ShowEntitySerializer extends TypeAdapter<HoverEvent.ShowEntity> {
-  static TypeAdapter<HoverEvent.ShowEntity> create(final Gson gson) {
-    return new ShowEntitySerializer(gson).nullSafe();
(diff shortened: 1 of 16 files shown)


text-serializer-gson/src/main/java/net/kyori/adventure/text/serializer/gson/StyleSerializer.java
```diff
@@ -99,6 +110,9 @@ static TypeAdapter<Style> create(final net.kyori.adventure.text.serializer.json.
   private final net.kyori.adventure.text.serializer.json.LegacyHoverEventSerializer legacyHover;
   private final boolean emitLegacyHover;
   private final boolean emitModernHover;
+  private final boolean emitSuperModernHover;
```
turikhay: I'm not sure if `superModern` is a right fit here. What if Mojang introduced further changes, would it be `superDuperModern`?

kezz: Agreed. Let's deprecate the old options and be explicit in either what they do, or what version the changes were introduced in.

MiniDigger: ideally end users should just use something like 
```java
GsonComponentSerializer.builder()
              .options(JSONOptions.byDataVersion().at(Bukkit.getUnsafe().getDataVersion()))
              .build();
```
anyways

MiniDigger: pushed something, how does that feel?

turikhay: It's way better

MiniDigger:
(rebased to fix conflicts)

text-serializer-gson/src/main/java/net/kyori/adventure/text/serializer/gson/ShowEntitySerializer.java
```diff
@@ -60,9 +65,10 @@ public HoverEvent.ShowEntity read(final JsonReader in) throws IOException {
 
     while (in.hasNext()) {
       final String fieldName = in.nextName();
-      if (fieldName.equals(SHOW_ENTITY_TYPE)) {
+      // TODO make this more flexible
```
MiniDigger: I cant think of a way to make this support old and new rn, need input

turikhay: In any case, I don't think this can be done in backwards-compatible way.

We should probably create a wrapper for `HoverEvent.ShowItem` and `HoverEvent.ShowEntity`, and serialize that instead. Wrapper would include all necessary context for Serializers to work with.

MiniDigger:
rebased on adventure 4.20 and fixed the failing tests

kezz (approved):

lucko (approved):

kashike (approved):

PR #1179, opened 2025-03-05
fix(minimessage): Carry over target into context deserialize calls

### LINKED ISSUE #1175: MiniMessage: target of Context gets "lost" when deserializing
When using `Context#deserialize` on a Context that has a target, that target doesn't get passed down to the deserialization.
For example, assume a `<placeholder>` tag that uses the target for something. Now when deserializing a string like `<tr:example.key:'<placeholder:lolxd/>'/>` with a target, that target doesn't get passed to the deserialization of the argument.

kezz:
Closes #1175.

### DIFF
--- a/text-minimessage/src/main/java/net/kyori/adventure/text/minimessage/ContextImpl.java
+++ b/text-minimessage/src/main/java/net/kyori/adventure/text/minimessage/ContextImpl.java
@@ -138,19 +138,34 @@ class ContextImpl implements Context {
 
   @Override
   public @NotNull Component deserialize(final @NotNull String message) {
-    return this.miniMessage.deserialize(requireNonNull(message, "message"), this.tagResolver);
+    requireNonNull(message, "message");
+    if (this.target != null) {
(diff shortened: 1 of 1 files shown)


text-minimessage/src/main/java/net/kyori/adventure/text/minimessage/ContextImpl.java
```diff
@@ -138,19 +138,34 @@ public UnaryOperator<String> preProcessor() {
 
   @Override
   public @NotNull Component deserialize(final @NotNull String message) {
-    return this.miniMessage.deserialize(requireNonNull(message, "message"), this.tagResolver);
+    requireNonNull(message, "message");
+    if (this.target != null) {
+      return this.miniMessage.deserialize(message, this.target, this.tagResolver);
+    } else {
+      return this.miniMessage.deserialize(message, this.tagResolver);
+    }
```
zml2008: silly thing but this block is repeated in all the overloads, can you break it out into its own method?

zml2008 (changes requested):

PR #1180, opened 2025-03-06
feature(minimessage): Check for colors before parsing phases

### LINKED ISSUE #1178: Look into an alternative for standard JDK number parsing
We have a lot of speculative calls to parseInt/parseDouble that, on failure, create exceptions that are ignored. Exceptions are "expensive", so we could consider looking into if it's quicker to a) write a "check" to see if a number could be parsed as a double/int or b) provide our own non-throwing versions of the above methods.

kezz:
This is another incredibly minor performance improvement as parseDouble incurs a slight performance cost when failing. If we instead check for colors first, we can guarantee that we are either parsing a valid double or throwing an exception (in which case we don't really care about performance).

Closes #1178.


MiniDigger (approved):
stonks 🚀

emilyy-dev (approved):
sexy

kashike (approved):

zml2008 (approved):

PR #1188, opened 2025-03-11
Add NumberBinaryTag#numberValue

GliczDev:
simplifies getting the actual number value from a number tag instead of doing a switch statement

### DIFF
--- a/nbt/src/main/java/net/kyori/adventure/nbt/ByteBinaryTagImpl.java
+++ b/nbt/src/main/java/net/kyori/adventure/nbt/ByteBinaryTagImpl.java
@@ -73,6 +73,11 @@ final class ByteBinaryTagImpl extends AbstractBinaryTag implements ByteBinaryTag
   }
 
   @Override
+  public Number numberValue() {
+    return this.value;
+  }
+
+  @Override
   public boolean equals(final @Nullable Object other) {
     if (this == other) return true;
     if (other == null || this.getClass() != other.getClass()) return false;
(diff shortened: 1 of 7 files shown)


kezz:
See the build failures

GliczDev:
should be fixed

nbt/src/main/java/net/kyori/adventure/nbt/ByteBinaryTagImpl.java
```diff
@@ -72,6 +72,11 @@ public short shortValue() {
     return this.value;
   }
 
+  @Override
+  public Number numberValue() {
```
kezz: Should be using nullability annotations for all values here

nbt/src/main/java/net/kyori/adventure/nbt/NumberBinaryTag.java
```diff
@@ -81,4 +81,12 @@ public interface NumberBinaryTag extends BinaryTag {
    * @since 4.0.0
    */
   short shortValue();
+
+  /**
+   * Gets the value as a {@code Number}.
```
kezz:
```suggestion
   * Gets the value as a {@link Number}.
```

kezz (changes requested):
Minor comments

nbt/src/main/java/net/kyori/adventure/nbt/ByteBinaryTagImpl.java
GliczDev: do you mean `@NotNull`?

kezz (approved):
LGTM!

GliczDev:
I updated the version in `@since` tag according to the milestone :)

zml2008 (approved):

PR #1191, opened 2025-03-14
feat(dfu, nbt-dfu): init

zml2008:
These new modules are intended to allow better integration with the DFU library that drives a growing amount of serialization in modern Minecraft.

## to do

- [ ] Finish implementing `BinaryTagOps`
- [ ] Write codecs for base adventure types
    - `Key`
    - `TextColor`
    - `TextDecoration`
    - `UUID`
    - `TranslationArgument`
    -  `Index` types
- [ ] Write codec for `Component`
- [ ] Figure out how we want to support the whole version range (different copies of the module for different DFU versions?? do we use `option`s to configure the same way we do on Gson?)

### DIFF
--- a/bom/build.gradle.kts
+++ b/bom/build.gradle.kts
@@ -14,9 +14,11 @@ dependencies {
     sequenceOf(
       "api",
       "annotation-processors",
+      "dfu",
       "extra-kotlin",
       "key",
       "nbt",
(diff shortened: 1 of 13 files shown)


kezz:
Progress now being tracked in a separate repo: https://github.com/PaperMC/adventurex

PR #1199, opened 2025-03-28
feat(minimessage): add option to stop the serializer from escaping minimessage tags in component content

ProdPreva1l:
While trying to support both minimessage and legacy ampersand codes to format text in my project, I came across an annoyance where the serializer would escape all minimessage tags in the `content` of a component. 

Meaning from an input of `<aqua>&l a` into the legacy component deserializer I was getting `\<aqua><bold> a` when re-serializing with the minimessage serializer, which obviously wouldn't format the aqua tag.
So this adds a simple option to the minimessage builder to tell the serializer, more specifically the collector to not escape tags in this situation.
And so now by setting `MiniMessage.Builder#escapeContent(boolean)` to false, an input of `<aqua>&l a` will result in a re-serialized output of `<aqua><bold> a` in term formatting everything as intended.

### DIFF
--- a/text-minimessage/build.gradle.kts
+++ b/text-minimessage/build.gradle.kts
@@ -9,6 +9,7 @@ dependencies {
   api(projects.adventureApi)
   testImplementation(projects.adventureTextSerializerPlain)
   testImplementation(projects.adventureTextSerializerAnsi)
+  testImplementation(projects.adventureTextSerializerLegacy)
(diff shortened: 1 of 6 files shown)


zml2008:
the point of MM's serializer is that deserialize(serialize(x)) should be visually equivalent to a component x. This breaks that essential behaviour.

PR #1201, opened 2025-03-31
fix(key): Use String#intern in KeyImpl

kermandev:
I changed the KeyImpl to use a String#intern to save on memory usage with the same key objects created, a better option than using a key pool, and should be faster to compute #equals by reference checks for long-lived keys.

## Motovation
Minestom recently migrated to using keys from its namespace type; From this change, we saw an increased memory usage of about ~4MB. This was due to our old implementation using an intern for the namespace and value. This increase isn't great. This is testable on the demo server available in Minestom's repo, and we hold about 39k instances of Key at runtime from its registries. After this change, we went from 5.42MB to 1.1MB retained from KeyImpl.

This issue could be primarily mitigated in Minestom internals by dropping the default namespace. Still, any user could do this with any namespace, and since we bundle adventure to our users, this is not an applicable option.

### Another solution
The adventure could change the [parser](https://github.com/KyoriPowered/adventure/blob/main/4/key/src/main/java/net/kyori/adventure/key/Key.java#L107-L112) to use intern instead after substrings, but users could still inadvertently create many long-lived equal strings.

Let me know your feedback on this change and if it breaks any style guides that Adventure may have.


kezz (changes requested):
Hi! Thanks for the PR! We've had a discussion internally and would like to instead implement this via a system property, that sets whether NONE, NAMESPACE or ALL parts of the key are interned. You can see our AdventureProperties internal helper class for creating system properties (probably would want it to be `key.internKeyParts`) for a pointer to this.

kashike:
`adventure.internKeyParts`

kermandev:
Wouldn't creating a shared properties package between the two at that point be a better solution than two separate properties API's (as the key doesn't require adventures api package)?

I would think pulling out the properties package from the internals would be nicer than implementing the same system twice, and the property would be under `net.kyori.adventure.internKeyParts` using that system

kermandev:
I've provided an example of the cryptic wording I used above; please let me know your feedback and the correct default value that has been decided. (I'm assuming NONE is good enough to provide backward compatibility).

Just to let you know, there is still a TODO on rewriting tests to follow the strategy used; I am unsure if I should expose the internal intern strategy. (Maybe move this property entirely under Key/KeyImpl vs AdventureProperties) and expose the enum there, but not all implementations of Key will ever have to have this strategy. It's only for the current implementation.

kermandev:
This should be good to review; I don't know what way you would like the test structured as all of the properties in Adventure are also stored as final in their implementations.

properties/src/main/java/net/kyori/adventure/internal/properties/AdventureProperties.java
```diff
@@ -60,6 +60,13 @@ public final class AdventureProperties {
    */
   public static final Property<Boolean> TEXT_WARN_WHEN_LEGACY_FORMATTING_DETECTED = property("text.warnWhenLegacyFormattingDetected", Boolean::parseBoolean, Boolean.FALSE);
 
+  /**
+   * Property for specifying what strategy to use for interning strings in keys.
+   *
+   * @since 4.21.0
+   */
+  public static final Property<KeyInternStrategy> KEY_INTERN_STRATEGY = property("key.internKeyParts", KeyInternStrategy::valueOf, KeyInternStrategy.NAMESPACE);
```
kashike:
```suggestion
  public static final Property<KeyInternStrategy> KEY_INTERN_STRATEGY = property("key.internStrategy", KeyInternStrategy::valueOf, KeyInternStrategy.NAMESPACE);
```

kashike (changes requested):

kashike (approved):

api/build.gradle.kts
```diff
@@ -10,6 +10,7 @@ configurations {
 }
 
 dependencies {
+  api(projects.adventureProperties)
```
kezz: We should make this implementation so it's not exposed to users

key/build.gradle.kts
```diff
@@ -9,6 +9,7 @@ configurations {
 }
 
 dependencies {
+  api(projects.adventureProperties)
```
kezz: Same here

key/src/test/java/net/kyori/adventure/key/KeyTest.java
```diff
@@ -124,4 +127,31 @@ void testNulChar() {
   void testParseWithEmptyNamespace() {
     assertEquals(Key.key(Key.MINECRAFT_NAMESPACE, "test"), Key.key(":test"));
   }
+
+  @SuppressWarnings("ConstantConditions") // We know backed as not null
+  @Test
+  void testInternDefault() {
```
kezz: This test doesn't actually do anything as we don't rerun the test with different intern modes. I'm not sure if junit has the ability to do this, or if we should be mocking here. I'm happy to just remove this either way as it's simple code.

properties/src/main/java/net/kyori/adventure/internal/properties/AdventureProperties.java
```diff
@@ -94,4 +101,16 @@ public interface Property<T> {
      */
     @Nullable T value();
   }
+
+  /**
+   * The strategy to use for interning strings in keys.
+   *
+   * @since 4.21.0
```
kezz: Since tags need bumping up 4.22

kezz (changes requested):
Just a few minor changes!

key/src/test/java/net/kyori/adventure/key/KeyTest.java
kermandev: The test was removed, as I did not find a good way of doing it.

kezz (approved):
Thanks!

settings.gradle.kts
```diff
@@ -37,6 +37,7 @@ sequenceOf(
   "extra-kotlin",
   "key",
   "nbt",
+  "properties",
```
zml2008, line 40: Let's add this module to the bom just for completeness sake -- feels like the kind of thing that could cause subtle funky issues down the road if versions are mismatched.

kermandev:
I saw the milestone moved to 4.23.0; Should I go ahead and change 4.22.0 to 4.23.0?

kezz:
Don't worry about that, we can do it on merge.

Strokkur424 (changes requested):
This needs a rebase to the `main/5` branch.

PR #1210, opened 2025-04-10
feature(minimessage): Rainbow tag saturation argument

### LINKED ISSUE #1038: Add brightness/lightness and a wrap around tag to mini message <rainbow> tag

Privatech38:
Added a saturation argument (optional) to `rainbow` tag. It simply sets the saturation value of each `HSVLike` to the argument value or 1 if omited.

Why? Max saturation can be unpleasant for eyes, especially on longer texts.

The syntax would follow `<rainbow:[!][phase]:[saturation]>`. 

Since the current algortihm has default values if argument is of lenght 0 this allows for such syntax:
- `<rainbow:0:0.65>`
- `<rainbow::0.65>`

It also uses `Float#parseFloat` method meaning any legal Java float representation is allowed: `.65`, `0.65f`, `1` etc.

This PR also partially closes #1038

### DIFF
--- a/text-minimessage/src/main/java/net/kyori/adventure/text/minimessage/tag/standard/RainbowTag.java
+++ b/text-minimessage/src/main/java/net/kyori/adventure/text/minimessage/tag/standard/RainbowTag.java
@@ -86,7 +86,7 @@ final class RainbowTag extends AbstractColorChangingTag {
           }
         }
       }
-     }
+    }
     return new RainbowTag(reversed, phase, saturation, ctx);
   }
 
--- a/text-minimessage/src/main/java/net/kyori/adventure/text/minimessage/tag/standard/RainbowTag.java
+++ b/text-minimessage/src/main/java/net/kyori/adventure/text/minimessage/tag/standard/RainbowTag.java
@@ -51,12 +51,14 @@ final class RainbowTag extends AbstractColorChangingTag {
 
   private final boolean reversed;
   private final double dividedPhase;
+  private final float saturation;
 
   private int colorIndex = 0;
 
   static Tag create(final ArgumentQueue args, final Context ctx) {
     boolean reversed = false;
     int phase = 0;
+    float saturation = 1f;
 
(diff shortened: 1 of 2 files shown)


kezz (approved):
LGTM - needs an associated docs PR before merge. Thanks!

Privatech38:
https://github.com/KyoriPowered/adventure-docs/pull/210 documentation PR

Strokkur424 (approved):
This PR needs to be rebased to `main/5`

PR #1211, opened 2025-04-10
fix(minimessage): rename `numeric` to `string` 

tjalp:
biggest pr ever

the method name is wrong. That's all there is😁

### DIFF
--- a/text-minimessage/src/main/java/net/kyori/adventure/text/minimessage/translation/Argument.java
+++ b/text-minimessage/src/main/java/net/kyori/adventure/text/minimessage/translation/Argument.java
@@ -92,7 +92,7 @@ public final class Argument {
    */
   @Deprecated
   public static @NotNull ComponentLike numeric(final @TagPattern @NotNull String name, final @NotNull String value) {
-    return argument(name, TranslationArgument.component(Component.text(value)));
+    return string(name, value);
   }
 
   /**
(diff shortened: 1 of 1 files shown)


kezz (changes requested):
damn, now i can't bully zml anymore..

good spot! unfortunately as this is technically a breaking change we need to leave the old method deprecates for removal and add a new one in its place

tjalp:
> damn, now i can't bully zml anymore..
> 
> 
> 
> good spot! unfortunately as this is technically a breaking change we need to leave the old method deprecates for removal and add a new one in its place 

right, I thought that too. Will update tomorrow

zml2008 (approved):

text-minimessage/src/main/java/net/kyori/adventure/text/minimessage/translation/Argument.java
```diff
@@ -88,11 +88,25 @@ private Argument() {
    * @param value the value
    * @return the named argument
    * @since 4.20.0
+   * @deprecated for removal since 4.21.0, use {@link #string(String, String)} instead.
    */
+  @Deprecated
   public static @NotNull ComponentLike numeric(final @TagPattern @NotNull String name, final @NotNull String value) {
     return argument(name, TranslationArgument.component(Component.text(value)));
```
kashike:
```suggestion
    return string(name, value);
```

kashike (dismissed):

kezz (approved):

PR #1213, opened 2025-04-11
add fromString(String name) to NamedTextColor

Endkind:


RealBauHD:
`NamedTextColor.NAMES.value("blue");`

kezz:
Thanks for the PR but I'm going to close this as we already have this functionality in the Index as exampled above.

PR #1217, opened 2025-04-16
Fix potential unwanted behavior TextColor.java

vadim-soude:
If you pass a long in the function instead of an int it will result to some undifined behavior even if the value of the long is inside the range of 0 - 255.

Code exemple :
```
long time = player.getWorld().getFullTime() % 64;
TextColor color1 = TextColor.color((int)time, 0, 0);
TextColor color2 = TextColor.color(time, 0, 0);

player.sendMessage("Game time value : " + time);
player.sendMessage("Color1 red value : " + color1.red());
player.sendMessage("Color2 red value : " + color2.red());
```

In game i will have :
```
Game time value : 61
Color1 red value : 61
Color2 red value : 195
```

I just added a cast to the methode so any long used will be treated as an int, the range of the methode allready prevent number bigger that the interger limit to be put inside.


kezz:
The issue here is that by using using a long as a parameter you are instead calling `TextColor#color(float, float, float)` instead of the method that accepts ints. This means the resulting color values will differ as the float method requires values in the range of [0, 1]. This fix does nothing to change that and instead you should be careful about what method you are calling.

PR #1219, opened 2025-04-22
feature(api): Skip style builder creation on merge

kezz:
`mergeStyle` is called a lot, especially in TranslatableComponentRenderer, where there is no merging to be done. By skipping the creation of the builder if the merge is a no-op, we can save some allocations.

### DIFF
--- a/api/src/main/java/net/kyori/adventure/text/AbstractComponentBuilder.java
+++ b/api/src/main/java/net/kyori/adventure/text/AbstractComponentBuilder.java
@@ -289,7 +289,13 @@ abstract class AbstractComponentBuilder<C extends BuildableComponent<C, B>, B ex
   @Override
   @SuppressWarnings("unchecked")
   public @NotNull B mergeStyle(final @NotNull Component that, final @NotNull Set<Style.Merge> merges) {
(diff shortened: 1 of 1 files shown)


api/src/main/java/net/kyori/adventure/text/AbstractComponentBuilder.java
```diff
@@ -289,7 +289,13 @@ private void prepareChildren() {
   @Override
   @SuppressWarnings("unchecked")
   public @NotNull B mergeStyle(final @NotNull Component that, final @NotNull Set<Style.Merge> merges) {
-    this.styleBuilder().merge(requireNonNull(that, "component").style(), merges);
+    final Style thatStyle = requireNonNull(that, "that").style();
+    if (thatStyle.isEmpty() && merges.isEmpty()) return (B) this;
+    if (!this.hasStyle()) {
+      this.style = thatStyle;
```
lucko: does this case also need to take into account `merges`?

kezz: Yes you're right, good spot!

Issue #1157, opened 2025-01-22
Translating a component inside a TranslatableComponent as an argument


MerryDev:
Using [this ](https://hastebin.com/share/benucusedu.java) registry implementation and this properties file
```properties
team.red = <red>Team Red
mapsetup.stage.9.name.success = <grey>The spawnpoint of <arg:0> <grey>has <green>successfully<grey>been set.
```

Returns `<grey>The spawnpoint of team.red <grey>has <green>successfully<grey>been set.`

The code that sends the component to the player is
```java
TranslatableComponent translatableComponent = Component.translatable("team.red");
Component component = Component.translatable("mapsetup.stage.9.name.success", translatableComponent);
player.sendMessage(component);
```

What I expect to get as an output is `<grey>The spawnpoint of <red>Team Red <grey>has <green>successfully<grey>been set.`

After debugging my code I realized that the argument is not treated as a TranslatableComponent although internally a implementation of it is used.

kezz:
Fixed with #1226

PR #1226, opened 2025-05-04
Render translatable arguments and children

TonytheMacaroni:
This is an attempt at fixing #1157.

~~Currently, it renders translation arguments & children components twice when using `MessageFormat` translations - not exactly sure on the best approach for avoiding this.~~

### DIFF
--- a/api/src/main/java/net/kyori/adventure/text/renderer/TranslatableComponentRenderer.java
+++ b/api/src/main/java/net/kyori/adventure/text/renderer/TranslatableComponentRenderer.java
@@ -89,34 +89,12 @@ public abstract class TranslatableComponentRenderer<C> extends AbstractComponent
       }
 
       @Override
-      protected @NotNull Component renderTranslatable(final @NotNull TranslatableComponent component, final @NotNull Locale context) {
+      protected @NotNull Component renderTranslatableInner(final @NotNull TranslatableComponent component, final @NotNull Locale context) {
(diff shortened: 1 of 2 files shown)


TonytheMacaroni:
Extracted out the parent component rendering to avoid rendering translation arguments & child components twice. Noticed, however, that this PR still does not account for rendering named arguments - not sure how it should be done w.r.t. the virtual components. For example, this test would fail:
```java
@Test
public void testRecursiveNamedArguments() {
    assertEquals(
        Component.text("Kezz is cool!"),
        this.translate(
            Component.translatable(
                "<name> is <thing>!",
                Argument.component("name", Component.translatable("<arg:0>", Component.text("Kezz"))),
                Argument.component("thing", Component.translatable("<arg:0>", Component.text("cool")))
            )
        )
    );
    assertEquals(
        Component.text("Kezz is cool!"),
        this.translate(
            Component.translatable(
                "<arg:0> is <arg:1>!",
                Argument.component("name", Component.translatable("<arg:0>", Component.text("Kezz"))),
                Argument.component("thing", Component.translatable("<arg:0>", Component.text("cool")))
            )
        )
    );
}
```

TonytheMacaroni:
Fixed another issue, where translatable tags in MiniMessage translations aren't passed through the global translator.

kezz (approved):

zml2008 (approved):

PR #1227, opened 2025-05-05
feat(key): Add KeyFormat

### LINKED ISSUE #1222: Improve `Key` creation customization
Having `minecraft` as the default namespace in `Key.key(String)` is sometimes annoying to work with, especially when the `Key` is used outside of Minecraft concepts. While there are workarounds (e.g., parsing the string manually), it would be appreciated if the API provided a possibility to handle this more cleanly.

A `Key` factory/format utility would also be useful in some cases, particularly since [`Key.key(String, char)`](https://jd.advntr.dev/key/4.20.0/net/kyori/adventure/key/Key.html#key(java.lang.String,char)) allows for some more custom formatting. Something like that:
```java
KeyFormat MC_KEYS = keyFormat().fallbackNamespace("minecraft").separator(':').build(); // Practically providing the default static `Key` methods in instance form

KeyFormat MY_KEYS = keyFormat("fallback", '>');

boolean isFineKey = MY_KEYS.parseable("minecraft:cat"); // false
Key mcCat = MY_KEYS.key("minecraft>cat"); // Key.key("cat") 
Key fbDog = MY_KEYS.key("dog"); // Key.key("fallback:dog")

Key castleKey = Key.key("fallback:castle"); // MY_KEY.key("castle")
String castleMyStr = MY_KEYS.asString(castleKey); // "fallback>castle"
String castleMyMinStr = MY_KEYS.asMinimalString(castleKey); // "castle"
```
`Key#parseable` is also missing an alternative with `char` as separator.

Emilxyz:
This pull request adds a `KeyFormat` class to allow for easier creation of custom Keys using the API.

`KeyFormat` has following functionality:
- Configuring a fallback  `namespace` to be used when no other namespace is specified in the key parsing
- Configuring a default `separator` to be used when parsing keys
- `KeyFormat#parse(String)` which parses a input string using the configured separator and fallback namespace
- `KeyFormat#key(String)` which creates a key using the configured namespace and the input string as value
- `KeyFormat#parseable(String)` to check if a input string can be parsed using the configuration of the key format
- `KeyFormat#asString(Key)` to turn a key into a String using the configuration
- `KeyFormat#asMinimalString(Key)` to turn a Key into its minimal String using the configuration

Closes #1222

### DIFF
--- a/key/src/main/java/net/kyori/adventure/key/KeyFormat.java
+++ b/key/src/main/java/net/kyori/adventure/key/KeyFormat.java
@@ -1,0 +1,267 @@
+/*
+ * This file is part of adventure, licensed under the MIT License.
+ *
+ * Copyright (c) 2017-2025 KyoriPowered
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in all
+ * copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+ * SOFTWARE.
+ */
+package net.kyori.adventure.key;
+
+import org.jetbrains.annotations.Contract;
+import org.jspecify.annotations.Nullable;
+
+import static java.util.Objects.requireNonNull;
+
+/**
+ * A factory to create {@link Key}s using configured defaults.
+ *
+ * <p>The configuration options include:</p>
+ * <dl>
+ *   <dt>namespace</dt>
+ *   <dd>The namespace to be used as fallback, when no other namespace is specified</dd>
+ *   <dt>separator</dt>
+ *   <dd>The character separating namespace and value</dd>
+ * </dl>
+ *
+ * <p>Example usage of this class:</p>
+ * <pre>{@code
+ *   KeyFormat format = keyFormat("adventure", ':');
+ *   Key key = format.key("translations");
+ *   format.asString(key); // -> "adventure:translations"
+ * }</pre>
+ *
+ * @see Key#key(String)
+ * @since 5.2.0
+ */
+public sealed interface KeyFormat extends Namespaced permits KeyFormatImpl {
+  /**
+   * Gets the default minecraft key format instance.
+   *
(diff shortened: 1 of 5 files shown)


key/src/main/java/net/kyori/adventure/key/KeyFormat.java
```diff
@@ -0,0 +1,249 @@
...
+   * Creates a key format with a fallback {@code namespace} and the {@link Key#DEFAULT_SEPARATOR}.
+   *
+   * @param namespace the namespace to be used as fallback in {@link #key(String)}
+   * @return the key format
+   * @throws IllegalArgumentException if the namespace contains an invalid character
+   * @since 4.22.0
+   */
+  static KeyFormat namespace(@KeyPattern.Namespace final @NotNull String namespace) {
+    return keyFormat(namespace, Key.DEFAULT_SEPARATOR);
+  }
```
Machine-Maker, line 92: Should probably also have a method to create from a `Namespaced` instance

key/src/main/java/net/kyori/adventure/key/KeyFormat.java
```diff
@@ -0,0 +1,249 @@
...
+  /**
+   * Creates a key format with a {@code separator} and the default {@link Key#MINECRAFT_NAMESPACE}.
+   *
+   * @param separator the separator to be used when parsing keys using {@link #key(String)}
+   * @return the key format
+   * @since 4.22.0
+   */
+  static KeyFormat separator(final char separator) {
+    return keyFormat(Key.MINECRAFT_NAMESPACE, separator);
+  }
```
Machine-Maker: This feels weird to have a method. I think just the methods with the namespace arg, and a method with the namespace arg and char separator, instead of implicitly using minecraft here.

key/src/main/java/net/kyori/adventure/key/KeyFormat.java
```diff
@@ -0,0 +1,249 @@
...
+  /**
+   * Turns a {@link Key} into its string representation in minimal form using the configuration of this format.
+   *
+   * <p>If the {@link Key#namespace()} of the key is {@link #namespace()}, only the {@link Key#value()} will be returned.</p>
+   *
+   * @param key the key
+   * @return the minimal string representation of the key
+   * @since 4.22.0
+   */
+  default @NotNull String asMinimalString(final @NotNull Key key) {
```
Machine-Maker, line 214: is `compact` better than `minimal`? Just a thought I had, idk if its better.

Machine-Maker (changes requested):

key/src/main/java/net/kyori/adventure/key/KeyFormat.java
Emilxyz: I don't entirely understand what you mean with this.
If it's about the method name, that was chosen to be consistent with `Key#asMinimalString`.

Machine-Maker: oh I didn't realize that method existed. Then its fine.

Emilxyz:
As for #1249, im unsure on what best to name that method, as we can't have another `#key(String)`. I see 2 options and need some input on them:

A:
The method to parse an input string is renamed to `#parse(String)`.
`#key(String)` takes in the value only and always uses the configured namespace.

B:
`#key(String)` remains the method for parsing.
And the method taking only a value is named something like `#value(String)`.

I like A more, just cause it makes more sense to me and sounds better (at least in my head). But, we would be breaking consistency with `Key#key(String)`, so I would also agree in going with B. When going with B we would need to come up with a better name for that method as I really don't like `#value`, but it's the best I could come up with rn.

I understand that the methods in `Key` were named to be used as static imports, but none of these methods are static and always require an instance of a key format. So, in my opinion, it might be better to disregard the consistency and go with what makes more sense/sounds nicer.

Emilxyz:
Another thing I thought about recently, is that it might be better to move the actual logic of creating/checking parsability of keys into the key format class. All the overloads I added in Key could be removed then and the existing ones just forward into `KeyFormat.minecraft()`.

Emilxyz:
Went ahead and implemented option A now, can always just rename the methods again if required.

imDaniX:
For consistency, we can use the name `parseKey` instead, create `Key#parseKey`, deprecate `Key#key(String)` and `Key#key(String, char)`, and move their responsibility to newly created `Key#parseKey`. It'll also allow for creating `Key#parseKey(String keyStr, String fallbackNamespace)`.

But ofc, I'm not a maintainer, just my thoughts on the topic.

Strokkur424 (approved):

PR #1232, opened 2025-05-22
fix(api): Don't throw an exception if a PointersSupplier has no parent

kezz:


api/src/main/java/net/kyori/adventure/pointer/PointersSupplierImpl.java
```diff
@@ -87,7 +87,10 @@ static final class ForwardingPointers<U> implements Pointers {
 
       // Fallback to the parent.
       if (resolver == null) {
-        resolver = this.supplier.parent.resolver(pointer);
+        final PointersSupplier<? super U> parent = this.supplier.parent;
+        if (parent != null) {
+          resolver = this.supplier.parent.resolver(pointer);
```
kashike:
```suggestion
        if (parent != null) {
          resolver = parent.resolver(pointer);
```

kashike (changes requested):

kashike (approved):

PR #1233, opened 2025-05-22
Typo fix in PrideTag.java

Stxellxa:
One of the colours in the Lesbian flag is typod to be D4662A6, and instead should be D462A6


kezz:
Thanks for the PR! Finally we can fix lesbianism

kezz (approved):

PR #1234, opened 2025-05-24
feat(dfu): impl codec

huanmeng-qwq:
for this PR, I've completed the content according to the TODO comments. However, since I directly implemented the codec for Component using Gson as the intermediary, we might not need `TRANSLATION_ARGUMENT`? I'm not sure—maybe this PR is just one possible implementation approach?


zml2008:
using gson as the intermediary completely defeats the purpose of having dfu codecs, don't do that

huanmeng-qwq:
I just submitted the refactored codec implementation for the new Component.

zml2008:
These codecs don't appear to match vanilla output, what is your source for the data structures used here? It also feels like more idiomatic DFU usage would be to use RecordCodecBuilder where possible.

dfu/build.gradle.kts
```diff
@@ -4,6 +4,9 @@ plugins {
 
 dependencies {
   api(libs.dfu8)
+  api(projects.adventureNbtDfu)
```
zml2008: why?

dfu/src/main/java/net/kyori/adventure/dfu/AdventureCodecs.java
```diff
@@ -1,13 +1,721 @@
...
+
+  public static final Codec<Component> COMPONENT = new ComponentCodec();
+  public static final Codec<BlockNBTComponent.WorldPos.Coordinate> COORDINATE = new CoordinateCodec();
+  public static final Codec<BlockNBTComponent.Pos> POS = new PosCodec();
+  public static final Codec<ClickEvent> CLICK_EVENT = new ClickEventCodec();
+  public static final Codec<HoverEvent<?>> HOVER_EVENT = new HoverEventCodec();
+  public static final Codec<Style> STYLE = new StyleCodec();
+  public static final Codec<DataComponentValue> DATA_COMPONENT_VALUE = new DataComponentValueCodec();
+
+  private static <A, S> Codec<S> xmap(Codec<A> codec, Function<? super A, ? extends S> to, Function<? super S, ? extends A> from, String name) {
```
zml2008: why not use the existing xmap method?

zml2008:
See build failures too, have you tested this at all? The DFU codecs should have test coverage comparable to other serializer implementations.

dfu/src/main/java/net/kyori/adventure/dfu/AdventureCodecs.java
huanmeng-qwq: Because the existing xmap method cannot specify a name, lol.

huanmeng-qwq:
Since the value in `TranslationArgument` is of type Object and could represent multiple numeric types, I can't refactor it to use `RecordCodecBuilder` for implementation.

huanmeng-qwq:
The build failure is caused by `nbt-dfu`.

kezz:
Thanks for the PR! However, we are going to continue work on a dfu module in-house over at https://github.com/PaperMC/adventurex.

PR #1239, opened 2025-06-08
feat(nbt): TagStringIO Enhancements

kermandev:
## Motivation
Since adventure release 4.21.0, users can now write any BinaryTag to a String; that left a gap in reading them.
There is also another case where it may be useful to capture the rest of the string, for example, in a command argument. (Otherwise, users would have to group count)
### So this commit:
* Adds support to read any BinaryTag (disallows trailing like the Compound version)
* Use a writer for any BinaryTag
* Adds support for reading embedded any BinaryTag, and CompoundBinaryTag (Allows trailing data, through passed Appendable)
* Adds support to CharBuffer to take the rest of the buffer.
### Whats with the special cases for CompoundBinaryTag a child of BinaryTag?
The special cases exist for compound methods, so you can immediately throw on the first failed parsing token, even though BinaryTags can represent a Compound; I did think about doing this through a functional interface because functions don't support checked throws, but this is more complex than it needs to be. (This wouldn't even be exposeable)

### DIFF
--- a/nbt/src/main/java/net/kyori/adventure/nbt/TagStringIO.java
+++ b/nbt/src/main/java/net/kyori/adventure/nbt/TagStringIO.java
@@ -93,6 +93,77 @@ public final class TagStringIO {
   }
 
   /**
+   * Read the string into a tag.
+   *
+   * <p>When working with untrusted input (such as from the network), users should be careful
+   * to validate that the {@code input} string is of a reasonable size.</p>
+   *
+   * @param input Input data
+   * @return the parsed tag
+   * @throws IOException on any syntax errors
+   * @since 4.22.0
+   */
+  public BinaryTag asTag(final String input) throws IOException {
+    try {
+      final CharBuffer buffer = new CharBuffer(input);
+      final TagStringReader parser = new TagStringReader(buffer);
+      parser.legacy(this.acceptLegacy);
+      final BinaryTag tag = parser.tag();
+      if (buffer.skipWhitespace().hasMore()) {
+        throw new IOException("Document had trailing content after first Tag");
+      }
+      return tag;
+    } catch (final StringTagParseException ex) {
+      throw new IOException(ex);
+    }
+  }
+
+  /**
+   * Read the string into an embedded tag, returning the remainder of the input.
+   *
+   * @param input the input string
+   * @param remainder the appendable to write the remainder to
+   * @return the parsed tag with the remainder
+   * @throws IOException on any syntax errors
+   * @since 4.22.0
+   */
+  public CompoundBinaryTag asCompound(final String input, final Appendable remainder) throws IOException {
+    try {
+      final CharBuffer buffer = new CharBuffer(input);
+      final TagStringReader parser = new TagStringReader(buffer);
+      parser.legacy(this.acceptLegacy);
+      final CompoundBinaryTag tag = parser.compound();
+      remainder.append(buffer.takeRest());
+      return tag;
+    } catch (final StringTagParseException ex) {
+      throw new IOException(ex);
+    }
+  }
+
+
+  /**
+   * Read the string into an embedded tag, returning the remainder of the input.
+   *
+   * @param input the input string
+   * @param remainder the appendable to write the remainder to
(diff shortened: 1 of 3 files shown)


kezz (approved):

nbt/src/main/java/net/kyori/adventure/nbt/TagStringIO.java
```diff
@@ -92,6 +92,76 @@ public CompoundBinaryTag asCompound(final String input) throws IOException {
...
+   *
+   * <p>When working with untrusted input (such as from the network), users should be careful
+   * to validate that the {@code input} string is of a reasonable size.</p>
+   *
+   * @param input Input data
+   * @return the parsed tag
+   * @throws IOException on any syntax errors
+   * @since 4.22.0
+   */
+  public BinaryTag asTag(final String input) throws IOException {
```
zml2008:
```suggestion
  public @NotNull BinaryTag asTag(final @NotNull String input) throws IOException {
```

nbt/src/main/java/net/kyori/adventure/nbt/TagStringIO.java
```diff
@@ -92,6 +92,76 @@ public CompoundBinaryTag asCompound(final String input) throws IOException {
...
+  /**
+   * Read the string into an embedded compound tag, returning the remainder of the input.
+   *
+   * @param input the input string
+   * @param remainder the appendable to write the remainder to
+   * @return the parsed tag with the remainder
+   * @throws IOException on any syntax errors
+   * @since 4.22.0
+   */
+  public CompoundBinaryTag asCompound(final String input, final Appendable remainder) throws IOException {
```
zml2008:
```suggestion
  public @NotNull CompoundBinaryTag asCompound(final @NotNull String input, final @NotNull Appendable remainder) throws IOException {
```

nbt/src/main/java/net/kyori/adventure/nbt/TagStringIO.java
```diff
@@ -92,6 +92,76 @@ public CompoundBinaryTag asCompound(final String input) throws IOException {
...
+  /**
+   * Read the string into an embedded tag, returning the remainder of the input.
+   *
+   * @param input the input string
+   * @param remainder the appendable to write the remainder to
+   * @return the parsed tag with the remainder
+   * @throws IOException on any syntax errors
+   * @since 4.22.0
+   */
+  public BinaryTag asTag(final String input, final Appendable remainder) throws IOException {
```
zml2008:
```suggestion
  public @NotNull BinaryTag asTag(final @NotNull String input, final @NotNull Appendable remainder) throws IOException {
```

nbt/src/main/java/net/kyori/adventure/nbt/TagStringIO.java
```diff
@@ -132,6 +202,20 @@ public String asString(final BinaryTag input) throws IOException {
    * @since 4.0.0
    */
   public void toWriter(final CompoundBinaryTag input, final Writer dest) throws IOException {
```
zml2008:
```suggestion
  public void toWriter(final @NotNull CompoundBinaryTag input, final @NotNull Writer dest) throws IOException {
```

nbt/src/main/java/net/kyori/adventure/nbt/TagStringIO.java
```diff
@@ -132,6 +202,20 @@ public String asString(final BinaryTag input) throws IOException {
...
+   * Writes a tag to in string format.
+   *
+   * <p>The provided {@link Writer} will remain open after reading a tag.</p>
+   *
+   * @param input Tag to write
+   * @param dest Writer to write to
+   * @throws IOException if any IO or syntax errors occur while parsing
+   * @since 4.22.0
+   */
+  public void toWriter(final BinaryTag input, final Writer dest) throws IOException {
```
zml2008:
```suggestion
  public void toWriter(final @NotNull BinaryTag input, final @NotNull Writer dest) throws IOException {
```

zml2008 (dismissed):

PR #1241, opened 2025-06-10
Avoid interpreting the "byte" suffix as a binary radix

### LINKED ISSUE #1240: NBT: Parsing byte 0 causes a StringIndexOutOfBoundsException
https://github.com/KyoriPowered/adventure/blob/92a5a6eab440cbca53f593660e963bff2dc22dda/nbt/src/main/java/net/kyori/adventure/nbt/TagStringReader.java#L334
The radix extraction logic fails to account that a byte value (with the suffix `b`) can be 0. This results in the entire string being extracted and a StringIndexOutOfBoundsException to throw when parsing the next characters
https://github.com/KyoriPowered/adventure/blob/92a5a6eab440cbca53f593660e963bff2dc22dda/nbt/src/main/java/net/kyori/adventure/nbt/TagStringReader.java#L270

Test SNBT: `{"test": 0b}`

IllusionTheDev:
Fixes #1240

### DIFF
--- a/nbt/src/main/java/net/kyori/adventure/nbt/TagStringReader.java
+++ b/nbt/src/main/java/net/kyori/adventure/nbt/TagStringReader.java
@@ -266,6 +266,10 @@ final class TagStringReader {
     // Determine the radix and strip its prefix if present
     final int radix = this.extractRadix(builder, original);
 
+    if (builder.length() == 0) {
+      throw this.buffer.makeError("Input is a radix, not a number");
+    }
+
     // Check for the sign before removing the type token because of hex number always needing a sign thanks to byte types
     final char last = builder.charAt(builder.length() - 1);
     boolean hasSignToken = false;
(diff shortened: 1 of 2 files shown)


nbt/src/main/java/net/kyori/adventure/nbt/TagStringReader.java
```diff
@@ -332,15 +336,18 @@ private int extractRadix(final StringBuilder builder, final String original) {
     if (first == '+' || first == '-') {
       radixPrefixOffset = 1;
     }
-    if (original.startsWith("0b", radixPrefixOffset) || original.startsWith("0B", radixPrefixOffset)) {
+
+    final int radixEndIndex = 2 + radixPrefixOffset;
+
+    if (original.length() > radixEndIndex && (original.startsWith("0b", radixPrefixOffset) || original.startsWith("0B", radixPrefixOffset))) {
```
kennytv, line 343: I'd add a comment here like `// There should be more after '0b', else it would be a regular byte tag`

kennytv (approved):

kezz:
Gonna merge this manually too, thanks!

PR #1247, opened 2025-06-20
Add `equals` to `VirtualComponent`

Seggan:
Fixes a problem I had with using virtual components as named translation arguments (like MiniMessage does), and them not updating because they were being compared equal despite not being equal.


TheMolkaPL (changes requested):
You forgot to implement `hashCode` my brother..

Seggan:
Oops 😅 done

kezz (approved):

PR #1249, opened 2025-06-21
Add Key#minecraft static method

Machine-Maker:


key/src/main/java/net/kyori/adventure/key/Key.java
```diff
@@ -126,6 +126,18 @@ public interface Key extends Comparable<Key>, Examinable, Namespaced, Keyed {
     return key(Objects.requireNonNull(namespaced, "namespaced").namespace(), value);
   }
 
+  /**
+   * Creates a key in the {@code minecraft} namespace.
+   *
+   * @param value the value
+   * @return the key
+   * @throws InvalidKeyException if value contains an invalid character
+   * @since 4.23.0
```
mbax: How can it be since a version that already came out?

Machine-Maker: I SWEAR I typed 4.24.0. Like I thought about it, I knew next was 4.24. idk what happened.

kezz:
I'd prefer a merge of #1227 than this, we could then expose a standard minecraft keyformat or something

kezz:
I'm going to go ahead and close this in favour of the other PR linked in my previous comment.

PR #1252, opened 2025-06-26
Implement #canTranslate in GlobalTranslatorImpl and call #canTranslate of its sources

ivi-kiwi:
Similar to the other methods in the GlobalTranslator, it simply iterates through the sources and checks whether a translation is possible there.

The reason for the proposed change is that if you add a source to the GlobalTranslator which overwrites #canTranslate, the logic in there is not taken into account.

In my individual case, this is necessary because I have my own implementations of TranslatableComponent in my project, to which my Translator only reacts if the instance passed to #translate is of the type of my own implementation and therefore my Translator is no longer addressed at all.

### DIFF
--- a/api/src/main/java/net/kyori/adventure/translation/GlobalTranslatorImpl.java
+++ b/api/src/main/java/net/kyori/adventure/translation/GlobalTranslatorImpl.java
@@ -81,6 +81,17 @@ final class GlobalTranslatorImpl implements GlobalTranslator {
   }
 
   @Override
+  public boolean canTranslate(final @NotNull String key, final @NotNull Locale locale) {
+    requireNonNull(key, "key");
+    requireNonNull(locale, "locale");
+    for (final Translator source : this.sources) {
+      boolean canTranslate = source.canTranslate(key, locale);
+      if (canTranslate) return true;
+    }
+    return false;
+  }
+
+  @Override
   public @Nullable MessageFormat translate(final @NotNull String key, final @NotNull Locale locale) {
     requireNonNull(key, "key");
     requireNonNull(locale, "locale");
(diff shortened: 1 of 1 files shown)


api/src/main/java/net/kyori/adventure/translation/GlobalTranslatorImpl.java
```diff
@@ -80,6 +80,17 @@ public boolean removeSource(final @NotNull Translator source) {
     return TriState.FALSE;
   }
 
+  @Override
+  public boolean canTranslate(final @NotNull String key, final @NotNull Locale locale) {
+    requireNonNull(key, "key");
+    requireNonNull(locale, "locale");
+    for (final Translator source : this.sources) {
+      boolean canTranslate = source.canTranslate(key, locale);
+      if (canTranslate) return true;
```
kezz:
```suggestion
      if (source.canTranslate(key, locale)) return true;
```

kezz (changes requested):
See build failures and suggestion.

For reference, although this PR is a good change, your use case of having your own implementation of `TranslatableComponent` is entirely unsupported and will break in the upcoming Adventure 5.0 release. I would suggest removing your use case for that or migrating to virtual components instead.

ivi-kiwi:
> For reference, although this PR is a good change, your use case of having your own implementation of `TranslatableComponent` is entirely unsupported and will break in the upcoming Adventure 5.0 release. I would suggest removing your use case for that or migrating to virtual components instead.

Oh, then I'll look into finding a better solution for my case. But it was a handy solution, and it worked very well until the change. 😅

kezz (approved):
Thanks for the PR!

PR #1255, opened 2025-06-28
fix: component flattener not popping styles in correct order

### LINKED ISSUE #1254: Component Flattener does not pop styles of nested components in correct order
Take the following component:

```java
final Component input = Component.text()
      .content("Hi there my")
      .append(Component.text(" clickable ")
        .clickEvent(ClickEvent.copyToClipboard("some text")
        .append(
          Component.text("and blue ", NamedTextColor.BLUE)))
      .append(Component.text("friend"))
      .build();
```

Until Adventure 4.22.0, this would pop in the following order:
- blue
- click event
- empty
- emtpy

Which is expected and correct. As of 4.22.0 (specifically, PR #1237), this is no longer the case, and it is popped in this order:
- blue
- empty
- click event
- empty

This happens because the new implementation does not have any code to pop the styles of components that have children, so they are popped at the end only, which is incorrect.

cc PR author @kezz 

I've made a fix and will submit a PR shortly.

diogotcorreia:
Closes #1254

### DIFF
--- a/api/src/main/java/net/kyori/adventure/text/flattener/ComponentFlattenerImpl.java
+++ b/api/src/main/java/net/kyori/adventure/text/flattener/ComponentFlattenerImpl.java
@@ -79,10 +79,12 @@ final class ComponentFlattenerImpl implements ComponentFlattener {
   private static final class StackEntry {
     final Component component;
     final int depth;
+    final int stylesToPop;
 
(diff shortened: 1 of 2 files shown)


kezz (approved):
Nice spot, thanks for the PR!

jpenilla (approved):

PR #1263, opened 2025-07-07
fix(nbt): Correct `defaultValue` nullability

tal5:
Currently all the `getX(String Key, X defaultValue)` methods have the `defaultValue` as `@NotNull`, which means you can't do something like `compound.getString("key", null)` to get the string under the key, or `null` otherwise.

This PR changes the relevant methods on `CompoundBinaryTag` and `ListBinaryTag` to have `@Nullable` return types, `@Nullable` `defaultValue` params, and a contract to specify that a non-null default value means a non-null return type.

> [!NOTE]
> Wasn't sure if `@Contract` should be in the same line as the method or above it and checkstyle was passing either way, so I put it above the method - let me know if I should in-line it.

image


kezz (approved):

kezz:
Thanks for the PR!

PR #1264, opened 2025-07-07
feature(nbt): Builders with an initial capacity

tal5:
Adds methods to create `ListBinaryTag` and `CompoundBinaryTag` builders with an initial capacity.

Added an `int initialCapacity` field to both builder impls that defaults to -1.
When creating the backing list/map, the builders now use `initialCapacity` if it isn't -1.
For the builder impl constructors, I just overloaded them and added `initialCapcity`, as I didn't want to specify the `-1` outside the impl class/change the existing ones - lmk if that's actually fine and should be changed.

image


kezz (approved):
Thanks for the PR! Just made a few minor style adjustments, but all good to go.

PR #1265, opened 2025-07-09
feature(nbt): `CompoundBinaryTag` contains methods

tal5:
Adds `CompoundBinaryTag#containsKey(Key)` and `#contains(Key, Type)` utility methods.

Pretty straightforward, only potential point of concern is that `#contains` currently uses the same numeric-types-match logic, (I.e. checking an int for being a long type would return true) - probably fine because that's how the other methods behave? but if not it can be changed into exact type matching.
image

### DIFF
--- a/nbt/src/main/java/net/kyori/adventure/nbt/CompoundBinaryTag.java
+++ b/nbt/src/main/java/net/kyori/adventure/nbt/CompoundBinaryTag.java
@@ -164,11 +164,11 @@ public interface CompoundBinaryTag extends BinaryTag, CompoundTagSetter<Compound
   boolean contains(final @NotNull String key);
 
   /**
-   * Returns whether the compound contains a tag of a specific type under a key.
+   * Returns whether the compound contains a tag of a specific type with a key.
    *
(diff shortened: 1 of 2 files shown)


nbt/src/main/java/net/kyori/adventure/nbt/CompoundBinaryTag.java
```diff
@@ -154,6 +154,25 @@ public interface CompoundBinaryTag extends BinaryTag, CompoundTagSetter<Compound
...
   }
 
+  /**
+   * Returns whether the compound contains a specific key.
+   *
+   * @param key the key to check for
+   * @return whether the compound contains the key
+   * @since 4.25.0
+   */
+  boolean containsKey(final @NotNull String key);
```
kashike:
```suggestion
  boolean contains(final @NotNull String key);
```

kashike (changes requested):

kezz (approved):
Thanks for the PR! Just a few more minor style changes but otherwise LGTM.

PR #1266, opened 2025-07-09
feature(nbt): `CompoundBinaryTag#builder(existing)`

tal5:
Adds `CompoundBinaryTag#createBuilder`, which creates a builder from the compound tag's current contents.
This is useful if you're editing a couple of values at once, like `compound.createBuilder().putInt("num", 6).putString("name", "Timothy").remove("old_value").build()`.

Might have merge conflicts with #1264, but I'll fix up whichever one is merged last.

### DIFF
--- a/nbt/src/main/java/net/kyori/adventure/nbt/CompoundBinaryTag.java
+++ b/nbt/src/main/java/net/kyori/adventure/nbt/CompoundBinaryTag.java
@@ -31,6 +31,7 @@ import java.util.stream.Collector;
 import java.util.stream.Stream;
 import org.jetbrains.annotations.NotNull;
 import org.jetbrains.annotations.Nullable;
+import org.jetbrains.annotations.Unmodifiable;
 
 import static java.util.Objects.requireNonNull;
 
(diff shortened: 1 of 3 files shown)


kashike:
Prefer `CompoundBinaryTag.builder(existingTag)`

tal5:
> Prefer `CompoundBinaryTag.builder(existingTag)`

Switched to that style, with a new `CompoundBinaryTag#asMap` method (returns an unmodifiable view of the tag's contents as a map) for a cleaner implementation without casting to the impl type or other approaches - that method is nice to have either way imo, for more convenient usage with any general map utility methods.

nbt/src/main/java/net/kyori/adventure/nbt/CompoundBinaryTag.java
```diff
@@ -170,31 +167,20 @@ public interface CompoundBinaryTag extends BinaryTag, CompoundTagSetter<Compound
...
-  boolean contains(final @NotNull String key, final @NotNull BinaryTagType<?> type);
-
-  /**
-   * Gets a set of all keys.
-   *
-   * @return the keys
-   * @since 4.0.0
-   */
-  @NotNull Set<String> keySet();
+  @Unmodifiable @NotNull Map<String, BinaryTag> asMap();
```
Strokkur424, line 183: Based on the implementation, this actually returns a view. I would mark it as such.

```suggestion
  @UnmodifiableView @NotNull Map<String, BinaryTag> asMap();
```

Strokkur424 (changes requested):
This PR needs rebasing for the `main/5` branch, which includes removing/replacing the Jetbrains nullability annotations with JSpecify ones.

PR #1267, opened 2025-07-09
feature(nbt): getting `BinaryTagType`s by their id

tal5:
Exposes `BinaryTagType#binaryTagType(byte id)`, to get a tag type by it's byte id.
This is needed if you're writing any sort of custom serialization, as currently you can get the id to serialize, but there's no way to de-serialize.

I also changed `BinaryTagType.List TYPES` to a `BinaryTagType[]` - technically hard-coding the size is not great, but seeing as:
- There haven't been any new types in over 8 years.
- It's a very easy change to make if any ever get added.
- It'll error on startup when calling `register` with any potential new tag type, so it's not an issue that's going to be missed.

it should be fine imo; still, lmk if you want me to revert it to the old list iteration & compare id approach.

Also, maybe it should return `null` instead of throwing? I just followed what the method already did, but maybe it would be better to change that, idk.

image


zml2008:
I think it was pretty intentional that we didn't expose raw type id's -- those are considered an implementation detail, sorry.

PR #1269, opened 2025-07-13
fix(nbt): zero byte tag radix parsing

kennytv:
e.g. `{test:0b}` would error before this change. The second change is just cosmetic


kezz (approved):

emilyy-dev:
duplicate of https://github.com/KyoriPowered/adventure/pull/1241 ?

kennytv:
Uh apparently, and theirs handles a case of what actually isn't accepted without quotes, `0x` 🐥

IllusionTheDev:
I'm happy with either solution as long as the problem gets fixed.

kennytv:
Yours is better

PR #1272, opened 2025-07-25
Add a less verbose way of creating titles

LaserSlime:
I have complained on the discord before about how creating titles is unnecessarily verbose and multiple people agreed with me. This is my idea to make it a little simpler. I think just specifying the duration in ticks is quite natural. Pretty much all APIs including Minecraft's own commands do it that way, so it shouldn't be confusing to anyone.

Another idea I had was adding a showTitle method that doesn't require a Title object and just takes the params directly but I'm not sure if that goes against adventure's design philosophy or not. I'm of course still open to any other ideas to improve title creation!

### DIFF
--- a/api/src/main/java/net/kyori/adventure/title/Title.java
+++ b/api/src/main/java/net/kyori/adventure/title/Title.java
@@ -73,6 +73,21 @@ public interface Title extends Examinable {
   }
 
   /**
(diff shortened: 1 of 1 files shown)


kezz (approved):
Thanks for the PR!

PR #1274, opened 2025-07-26
feature(minimessage): SmallCaps tag

### LINKED ISSUE #1271: SmallCaps format tag in MiniMessage
### Feature
Add a tag that will format text with SmallCaps style.

### Reason
Many servers use the SmallCaps characters for their messages, scoreboards etc.

Image

### TinyTextGenerator
Here's the generator that people usually use:
https://lingojam.com/TinyTextGenerator 

### So now the question is:
Is it possible to add it into kyori adventure? It would be an awesome feature.

kokiriglade:
Closes #1271

Still some stuff to do if this is wanted, but I had the code lying around for this in another project and figured I'd open up a PR just so there's something to work off of. 

I'm not entirely sure how to get the `assertSerializedEquals` working, and also not sure if my character mapping is the best default. 

cheers :-)

### DIFF
--- a/text-minimessage/src/main/java/net/kyori/adventure/text/minimessage/tag/standard/SmallCapsTag.java
+++ b/text-minimessage/src/main/java/net/kyori/adventure/text/minimessage/tag/standard/SmallCapsTag.java
@@ -1,0 +1,166 @@
+/*
+ * This file is part of adventure, licensed under the MIT License.
+ *
+ * Copyright (c) 2017-2025 KyoriPowered
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in all
+ * copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+ * SOFTWARE.
+ */
+package net.kyori.adventure.text.minimessage.tag.standard;
+
+import java.util.Collections;
+import java.util.HashMap;
+import java.util.Locale;
+import java.util.Map;
+import net.kyori.adventure.text.Component;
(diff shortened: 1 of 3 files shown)


Navio1430:
Thanks!

text-minimessage/src/main/java/net/kyori/adventure/text/minimessage/tag/standard/SmallCapsTag.java
```diff
@@ -0,0 +1,166 @@
...
+    smallFontCharMap.put('j', 'ᴊ');
+    smallFontCharMap.put('k', 'ᴋ');
+    smallFontCharMap.put('l', 'ʟ');
+    smallFontCharMap.put('m', 'ᴍ');
+    smallFontCharMap.put('n', 'ɴ');
+    smallFontCharMap.put('o', 'ᴏ');
+    smallFontCharMap.put('p', 'ᴘ');
+    smallFontCharMap.put('q', 'ǫ');
+    smallFontCharMap.put('r', 'ʀ');
+    smallFontCharMap.put('s', 's');
```
Navio1430, line 75: no need to replace s because they are the same

text-minimessage/src/main/java/net/kyori/adventure/text/minimessage/tag/standard/SmallCapsTag.java
```diff
@@ -0,0 +1,166 @@
...
+    smallFontCharMap.put('t', 'ᴛ');
+    smallFontCharMap.put('u', 'ᴜ');
+    smallFontCharMap.put('v', 'ᴠ');
+    smallFontCharMap.put('w', 'ᴡ');
+    smallFontCharMap.put('x', 'x');
```
Navio1430, line 80: no need to replace x

MiniDigger:
how does the narrator read small caps? I am having accessibility concerns

text-minimessage/src/main/java/net/kyori/adventure/text/minimessage/tag/standard/SmallCapsTag.java
```diff
@@ -0,0 +1,166 @@
...
+import static net.kyori.adventure.text.Component.text;
+
+/**
+ * Replaces text inside the tag with the "small caps" character set.
+ *
+ * @since 4.24.0
+ */
+final class SmallCapsTag implements Modifying {
+
+  private static final Map<Character, Character> smallFontCharMap = new HashMap<>();
```
MiniDigger, line 54: why not Map.of?

kokiriglade:
yeah accessibility is definitely something to think about here. the narrator seems to just ignore small caps. 

the ideal solution would probably be for servers to add a font that looks the same as small caps, but that doesn't seem to be a very popular choice.

text-minimessage/src/main/java/net/kyori/adventure/text/minimessage/tag/standard/SmallCapsTag.java
kokiriglade: intellij is saying thats java 9 api

RedVortexDev:
since this is modifying characters instead of changing text style I think this is less of a fit for adventure, additionally accessibility is also a concern (and is this an attempt for a precedent to adding other tags that change characters, such as the characters that have a circle around them?)

kokiriglade:
> since this is modifying characters instead of changing text style I think this is less of a fit for adventure

I'm inclined to agree, the more I think about it the more this feels like the perfect use case for a font with a resource pack.


I'm gonna close this PR. Happy for others to pick up where I left off if thats what people want, but just my thoughts :-)

MiniDigger:
thanks for the effort tho! 
sorry for not catching the accessibility concern on the original issue before you spend time on this

text-minimessage/src/main/java/net/kyori/adventure/text/minimessage/tag/standard/SmallCapsTag.java
novitpw, line 75: Use 'ꜱ' instead of 's'

PR #1280, opened 2025-07-30
fix invalid url extraction in legacy component serializer

derklaro:
The legacy component serializer currently extracts some invalid urls from an input string, leading to a client disconnect due to the validation performed in `net.minecraft.Util.parseAndValidateUntrustedUri(String)` (specifically calling `new URI()` on the input).

An example input where this is happening is: `Discord: <https://example.com/>`, as this results in an open url action with the url `https://example.com/>` which is invalid.

To circument this I've made the url regex more strict on the path part (only allowing chars as specified in rfc 3986 section 3.3), and added a validity check by calling `new URI()` before adding the click event. I'm not sure if the second check is actually required, however, I feel it is better to be safer here than accidentially disconnecting clients by extracting invalid urls.

Please let me know if additional test cases are required, I only added a few to validate the behaviour, maybe a few more would be better.


kezz (approved):

kezz:
Thanks for the PR!

PR #1286, opened 2025-08-12
Add `allowedInKey(char)` method to check character validity for keys

NonSwag:
Also fixed javadoc parameter names


kezz:
Can you explain the use-case for this? I'm not sure if it makes sense as a method given whether a character is allowed depends on where in the key it is used.

NonSwag:
My use case is/was in a `StringReader` for a custom argument type, but I guess you are right, maybe parsing the key would be smarter
image

PR #1293, opened 2025-08-22
feat: ObjectComponent serialization

Emilxyz:
Adds serialization support for the new ObjectComponent in:
- api - ComponentFlattener
- serializer-configurate4
- text-serializer-gson


Emilxyz:
Updated for 25w35a. 
I'll keep this as draft for now, probably until there's a pre release or release candidate even. I don't think it makes much sense to review it before that when mojang could be introducing changes again with the next snapshot

If anyone wants to leave feedback anyway, feel free!

kezz (approved):
haven't reviewed the configurate changes but json and flattener changes lgtm!

PR #1298, opened 2025-09-05
`TranslatableComponent$Builder#addArgument`

tal5:
## Additions

- `TranslatableComponent$Builder#addArgument` - adds a single translation argument, as opposed to the current methods which only allow setting the entire list.
- `TranslatableComponentImpl#asArgument` - converts a single `ComponentLike` into a `TranslationArgument`, with an optional `int` index for extra error info (ignored when `-1`, from e.g. the new `addArgument` ^).

## Changes

- `TranslatableComponentImpl$BuilderImpl.args` - now defaults to null, with:
- `TranslatableComponentImpl$BuilderImpl#args` - acting as a getter that defaults it to a new `ArrayList` if unset and returns its value (similarly to other builders around the API).
- `TranslatableComponentImpl$BuilderImpl#arguments` - now initializes `args` to a new `ArrayList` of the correct size and adds each argument manually using the new `TranslatableComponentImpl#asArgument`, so that the list is mutable (unlike the one returned by `asArguments`) .
- `TranslatableComponentImpl$BuilderImpl#build` - now defaults `args` to `Collections.emptyList()` if `null`.

> [!NOTE]
> `TranslatableComponentImpl#asArgument`'s logic to include/exclude the `int index` param from errors may be a little messy? I don't think it's a very big problem as it's just an internal util method, but let me know if you'd prefer a different approach.
> Or maybe using the string supplier variants for `requireNonNull` instead of computing the error strings every time.

image

### DIFF
--- a/api/src/main/java/net/kyori/adventure/text/TranslatableComponentImpl.java
+++ b/api/src/main/java/net/kyori/adventure/text/TranslatableComponentImpl.java
@@ -145,7 +145,7 @@ final class TranslatableComponentImpl extends AbstractComponent implements Trans
   static final class BuilderImpl extends AbstractComponentBuilder<TranslatableComponent, Builder> implements TranslatableComponent.Builder {
     private @Nullable String key;
     private @Nullable String fallback;
-    private List<TranslationArgument> args = Collections.emptyList();
+    private List<TranslationArgument> args = null;
 
     BuilderImpl() {
     }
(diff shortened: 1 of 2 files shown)


api/src/main/java/net/kyori/adventure/text/TranslatableComponentImpl.java
```diff
@@ -172,7 +185,11 @@ static final class BuilderImpl extends AbstractComponentBuilder<TranslatableComp
 
     @Override
     public @NotNull Builder arguments(final @NotNull List<? extends ComponentLike> args) {
-      this.args = asArguments(requireNonNull(args, "args"));
+      requireNonNull(args, "args");
+      this.args = new ArrayList<>(args.size());
+      for (int i = 0; i < args.size(); i++) {
+        this.args.add(asArgument(args.get(i), i));
+      }
```
kezz, line 192: why the new defensive copy here?

kezz (changes requested):

api/src/main/java/net/kyori/adventure/text/TranslatableComponentImpl.java
tal5: You mean the manual `ArrayList`? Because it needed to be a mutable list for the builder, so it couldn't use the existing `asArguments` (and I didn't want to change that method around because it's used in other places)

PR #1300, opened 2025-09-11
feat(minimessage): Added support for translatable tags to provide the target as an argument if present

4drian3d:
Currently, if you use the `<lang>` and `<lang_or>` tags with a custom MiniMessageTranslator with a TagResolver that obtains information from the parsing context, you cannot access the target initially provided when parsing the `<lang>` tag. This pull request provides support for this.
_In addition, a test has been added that provides an example of a real use case for this feature._


kezz (commented):
Would this change not break existing calls with translatables by inserting the "fake" target argument at the start?

4drian3d:
> Would this change not break existing calls with translatables by inserting the "fake" target argument at the start?

I don't see how it could affect anything. This pull request only modifies the `<lang>` and `<lang_or>` tags, where the only arguments it can receive are text. This pull request only inserts a target argument if it is present, and inserts it at the end of the arguments, so it should not change the order or behavior of any previously inserted arguments

PR #1304, opened 2025-09-18
feat(minimessage): named arguments

### LINKED ISSUE #1296: Named MiniMessage arguments
It would be nice to be able to optionally use a more powerful argument processor that accepted arguments in the form `<tag_name arg_name='arg value' boolean_presence_arg>`.

Strokkur424:
closes #1296.


Strokkur424:
I've rebased the PR to the `unstable/4` branch so that I can do some work for the upcoming head tag with the changes introduced here already.

text-minimessage/src/main/java/net/kyori/adventure/text/minimessage/tag/resolver/TagResolver.java
```diff
@@ -223,6 +273,18 @@ public boolean has(final @NotNull String name) {
...
+   * Gets a tag with named arguments from this resolver based on the current state.
+   *
+   * @param name the tag name
+   * @param arguments the arguments passed to the tag
+   * @param ctx the parse context
+   * @return a possible tag
+   * @throws ParsingException if the provided arguments are invalid
+   * @since 4.25.0
+   */
+  @Nullable Tag resolveNamed(@TagPattern final @NotNull String name, final @NotNull NamedArgumentMap arguments, final @NotNull Context ctx) throws ParsingException;
```
kezz: Needs to have a default null impl otherwise this is a breaking change

kezz (changes requested):

kezz (approved):

text-minimessage/src/main/java/net/kyori/adventure/text/minimessage/tag/resolver/TagResolver.java
```diff
@@ -142,6 +142,56 @@ public boolean has(final @NotNull String name) {
     };
   }
 
+  /**
+   * Create a tag resolver that only responds to certain tag names, and whose value does not depend on that name.
+   *
+   * <p>
+   * This method creates a special resolver which listens to tags with named arguments instead of sequences ones.
```
jpenilla:
```suggestion
   * This method creates a special resolver which listens to tags with named arguments instead of sequential ones.
```
there looks to be a couple other spots with the same issue

Strokkur424: I think I fixed them all

Strokkur424:
Rebased to main/4 after the unstable/4 merge

zml2008:
So a thought on this, why are we creating two different colours of MM tags? I feel like the design would be cleaner if we only had one type of tag, which accepted both named and sequential arguments, same as say with a CLI. For consistency's sake I think it's fair to require all the named args to be at the beginning or end of the tag, but either way there's a bunch of api noise and opportunity for breakages here due to adding named args in as a separate parallel concept.

I'm seeing something like:
```xml
<gradient phase=0.4 speed=2 red:green:blue>hi strokkur</gradient>
```
or:
```xml
<gradient:red:green:blue phase=0.4 speed=2>hi strokkur</gradient>
```

What do you think? Users would just pop named args from the `ArgumentQueue` same as other items

Strokkur424:
Alright, I've done everything I've wanted to do for this PR now. Arguments have been implemented, as per @zml2008's suggestion, in a combined way. They new syntax looks like this: 

```xml
<tagname
    someflag
    somename='a quoted value with spaces'
    anotherflag :queued arguments behind:here>Hey abby</tagname>
```

This means named arguments are simply placed between the tag name and the first colon.

PR #1253, opened 2025-06-27
Adventure 5.0

kezz:
### Introduction
This PR exists to track the progress on Adventure 5.0. It only exists as a PR for review purposes, once Adventure 5.0 is complete, the main branch of the repo will be changed to [`main/5`](https://github.com/KyoriPowered/adventure/tree/main/5).

Discussion in this PR will be limited to reviews of the PR itself. For discussion about broader changes in Adventure 5.0, please see https://github.com/KyoriPowered/adventure/issues/1202.

### To-do list:
The list below tracks remaining tasks. For reference, "refactor" means removal of all deprecations and refactoring of code to use modern Java language features.

- [x] Bump Java version to 21.
- [x] Refactor `annotation-processors` module.
- [x] Refactor `api` module.
- [x] Remove `extra-kotlin` module.
- [x] Refactor `key` module.
- [x] Refactor `nbt` module.
- [x] Refactor `serializer-configurate-4` module.
- [x] Refactor `text-logger-slf4j` module.
- [x] Refactor `text-minimessage` module.
- [x] Refactor `text-serializer-ansi` module.
- [x] Refactor `text-serializer-commons` module.
- [x] Refactor `text-serializer-gson` module.
- [x] Refactor `text-serializer-gson-legacy-impl` module.
- [x] Refactor `text-serializer-json` module.
- [x] Refactor `text-serializer-json-legacy` module.
- [x] Refactor `text-serializer-legacy` module.
- [x] Refactor `text-serializer-plain` module.
- [x] Fix GitHub actions builds.
- [x] Check all Javadocs, IJ's reformatting changes JD references to impl classes and fucks up spacing for no reason.
- [x] Migrate to jspecify.
- [x] Double check jspecify everywhere, some weirdness remaining.
- [x] Make `TextFormat` sealed.
- [x] Modularize!
- [x] Find some sort of abi/api break checker just to double check we've not removed anything important.
- [X] Check if there's any libraries we can/should update
- [x] Add a note on Component, Payload, Action that new types may be added at any time

### 4.0 changes required
Deprecate:
- [x] BuildableComponent
- [X] The "enumness" of ClickEvent$Action
- [X] ClickEvent$Action#payloadType

Add `ApiStatus.NonExtendible` to:
- [x] Title.Times
- [x] Sound.Builder
- [x] ResourcePackRequest (and Builder)
- [x] ResourcePackInfo (and Builder)
- [x] PointersSupplier (and Builder)
- [x] Pointers (and Builder)
[...]

PR #1307, opened 2025-09-21
fix(workflows): use Java 21 to build

tgbhy:
This just set the runtime_version of [`shared-ci.yml`](https://github.com/KyoriPowered/.github/blob/trunk/.github/workflows/shared-ci.yaml) to Java 21.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -9,4 +9,6 @@ on:
 jobs:
   call-build:
     uses: "KyoriPowered/.github/.github/workflows/shared-ci.yaml@trunk"
+    with:
+      runtime_version: 21
     secrets: "inherit"


zml2008:
Why?

tgbhy:
On https://github.com/KyoriPowered/adventure/pull/1253 the Adventure/5.0.0 use Java 21.
So, when building, the workflows currently use Java 17 to build which cause an error for incorrect JVM setup, so this PR aims to fix that so you can see the real errors to fix to have a functionnal build.
And when everything will be fixed, the CI will be able to work correctly.

I did tests on my GitHub before sending this PR, I just send unnecessary commits so I re-created this fork when everything was good for me.

kezz:
Thanks for the PR but I'm going to close this as Adventure 5.x is still WIP at this point - it doesn't compile anyway at this time.

PR #1312, opened 2025-09-27
fix(text-serializer-legacy): colors with no text are skipped

### LINKED ISSUE #1085: MiniMessage strips "redundant" tags which does not allow me to set prefixes as I can with legacy color codes
It's probably ideal to read up on this issue which I just opened and closed again as I was forwarded to here: https://github.com/NEZNAMY/TAB/issues/1312

My use case is basically I want to use a color code in a prefix, so I'm doing something for LuckPerms like '/lp group default meta setprefix "\<green\>"' so the prefix is nothing but <green>.
The problem is that with MiniMessage if the message ends in just <green> it will be stripped which is not the case with legacy color code like &a.

Now I have to either create a pretty horrible workaround by creating a meta for the groups that's legacy code while using MiniMessage in others and that thought is already driving me nuts.

Can we not have MiniMessage act like legacy code and carry over a color code if it ends in a color tag?

I know in most cases you'll have a string like "\<green\>This is really funny\<blue\>" in which you want to strip the <blue> but in cases like mine, there might be more components to the text which aren't included in just one string, so you would want it to actually use the <blue> from the previous string.

I hope I didn't ramble too much and this does make some sense. It's a little hard to explain all this but I tried the best I can.
If you need further information, let me know.

RaynLegends:
Fixes KyoriPowered/adventure#1085 (Not 100% sure the issue is related, I can open a separate one if needed ^^)

LegacyComponentSerializer does not serialize colors added to components with an empty text, which causes issues when using components as prefixes.
While this is not really an intended use case, it's unintuitive, causing for example the second of these tests to fail:

```
  @Test
  void testJustColor() { // Passes (Existing test case)
    assertEquals(Component.text("", TextColor.color(0xabcdef)), LegacyComponentSerializer.legacyAmpersand().deserialize("&#abcdef"));
  }

  @Test
  void testJustColorInverse() { // Fails (Basically the inverse of the first test, does not currently exist)
    assertEquals("&#abcdef", LegacyComponentSerializer.legacyAmpersand().serialize(Component.text("", TextColor.color(0xabcdef))));
    // Actual result: "" (an empty string)
  }
 ```
 
 The change in the PR does not change any other behaviour.
 Most importantly, it does not add "unused" formatting codes (for example colors that are immediately replaced with no text in between or erroneous resets like the one reported in KyoriPowered/adventure#287).

 I added a specific test (testToLegacyJustColor) that fails on the current version of adventure but works with the patched version.

### DIFF
--- a/text-serializer-legacy/src/main/java/net/kyori/adventure/text/serializer/legacy/LegacyComponentSerializerImpl.java
+++ b/text-serializer-legacy/src/main/java/net/kyori/adventure/text/serializer/legacy/LegacyComponentSerializerImpl.java
@@ -297,6 +297,8 @@ final class LegacyComponentSerializerImpl implements LegacyComponentSerializer {
     private @Nullable TextFormat lastWritten;
     private StyleState[] styles = new StyleState[8];
     private int head = -1;
+    private boolean applyTrailingStyle = false;
+    private int tail = -1;
 
     @Override
     public void pushStyle(final @NotNull Style pushed) {
@@ -308,6 +310,7 @@ final class LegacyComponentSerializerImpl implements LegacyComponentSerializer {
 
       if (state == null) {
         this.styles[idx] = state = new StyleState();
+        this.tail = Math.max(this.tail, idx);
       }
 
       if (idx > 0) {
@@ -323,12 +326,17 @@ final class LegacyComponentSerializerImpl implements LegacyComponentSerializer {
 
     @Override
     public void component(final @NotNull String text) {
-      if (!text.isEmpty()) {
-        if (this.head < 0) throw new IllegalStateException("No style has been pushed!");
-
-        this.styles[this.head].applyFormat();
-        this.sb.append(text);
+      if (text.isEmpty()) {
+        if (this.head == 0) {
+          this.applyTrailingStyle = true;
+        }
+        return;
       }
+
+      if (this.head < 0) throw new IllegalStateException("No style has been pushed!");
+
+      this.styles[this.head].applyFormat();
+      this.sb.append(text);
     }
 
     @Override
@@ -351,6 +359,10 @@ final class LegacyComponentSerializerImpl implements LegacyComponentSerializer {
 
     @Override
     public String toString() {
+      if (this.applyTrailingStyle) {
+        this.styles[this.tail].applyFormat();
+      }
+
       return this.sb.toString();
     }
 
--- a/text-serializer-legacy/src/test/java/net/kyori/adventure/text/serializer/legacy/LegacyComponentSerializerTest.java
+++ b/text-serializer-legacy/src/test/java/net/kyori/adventure/text/serializer/legacy/LegacyComponentSerializerTest.java
@@ -161,6 +161,12 @@ class LegacyComponentSerializerTest {
   }
 
   @Test
+  void testToLegacyJustColor() {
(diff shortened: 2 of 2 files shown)

### CODE REVIEW

**Comment by zml2008** [comment]
this is not a goal of the legacy serializer

**Comment by RaynLegends** [comment]
I understand that this particular use case may not be a goal, but the legacy serializer is really the only one that has this behavior.

I think this really is a bug when you consider that what the legacy serializer is doing is not "preventing" using it to represent prefixes, because you can absolutely do that as long as there is something after the color, but simply refusing to serialize color codes followed by an empty string (which is okay-ish when those colors would not change anything, like for "&a&eSomething" being changed to "&eSomething" -- even though this is also kind of unexpected and something that I would personally not see a serializer doing, this should probably be a job for a compacter/optimizer).

See the following tests:

## Minimessage Serializer - Both tests pass
```java
  @Test
  void testColorNonEmpty() {
    final String expected = "<green>Test";
    final Component component = text("Test").color(NamedTextColor.GREEN);
    this.assertSerializedEquals(expected, component);
    this.assertParsedEquals(component, expected);
  }

  @Test
  void testColorEmpty() {
    final String expected = "<green>";
    final Component component = text("").color(NamedTextColor.GREEN);
    this.assertSerializedEquals(expected, component);
    this.assertParsedEquals(component, expected);
  }
```

## Gson Serializer - Both tests pass
```java
  @Test
  void testColorNonEmpty() {
    final String expected = "{\"color\":\"green\",\"text\":\"Test\"}";
    final Component component = text("Test").color(NamedTextColor.GREEN);
    this.assertSerializedEquals(expected, component);
    this.assertParsedEquals(component, expected);
  }

  @Test
  void testColorEmpty() {
    final String expected = "{\"color\":\"green\",\"text\":\"\"}";
    final Component component = text("").color(NamedTextColor.GREEN);
    this.assertSerializedEquals(expected, component);
    this.assertParsedEquals(component, expected);
  }

  private void assertSerializedEquals(String serialized, Component component) {
    assertEquals(serialized, GsonComponentSerializer.gson().serialize(component));
  }

  private void assertParsedEquals(Component component, String serialized) {
    assertEquals(component, GsonComponentSerializer.gson().deserialize(serialized));
  }
```

## Legacy Serializer - The second test fails only on the first assert
```java
  @Test
  void testColorNonEmpty() {
    final String expected = "&aTest";
    final Component component = text("Test").color(NamedTextColor.GREEN);
    this.assertSerializedEquals(expected, component);
    this.assertParsedEquals(component, expected);
  }

  @Test
  void testColorEmpty() {
    final String expected = "&a";
    final Component component = text("").color(NamedTextColor.GREEN);
    this.assertSerializedEquals(expected, component); // This fails
    this.assertParsedEquals(component, expected); // This passes!
  }

  private void assertSerializedEquals(String serialized, Component component) {
    assertEquals(component, LegacyComponentSerializer.legacyAmpersand().deserialize(serialized));
  }

  private void assertParsedEquals(Component component, String serialized) {
    assertEquals(serialized, LegacyComponentSerializer.legacyAmpersand().serialize(component));
  }
```

Could this PR maybe be reconsidered? Thanks!

ytnoos:
I need this

RaynLegends:
Rebased onto https://github.com/KyoriPowered/adventure/commit/fe49db381684f32a539b187438b8fa8d771af1db

PR #1305, opened 2025-09-18
feat(minimessage): add head tag

Strokkur424:
This PR depends on #1304.

PR #1320, opened 2025-10-03
feat: add sequential (simple) head tag

Strokkur424:
This PR separates out the simplified/sequential form of the head tag from #1305.


kezz (approved):
Thanks for the PR!

PR #1321, opened 2025-10-03
fix(minimessage): First pass state not returning to NORMAL from STRING if quote is escaped

### LINKED ISSUE #1315: Escaped quotations in arguments cause a tag to be returned as a TEXT token
If you provide an argument like `\"foo\"`, it will never find a closing > causing the tag to not be parsed.

Example of this would be
```
<hover:show_text:'test\'>Hello!
```
or
```
<hover:show_text:t\"est\">Hello!
```

Privatech38:
Closes #1315 

This PR addresses the `TokenParser#parseString`'s problem when faced with an input like `<hover:show_text:"Foo\">Bar`. The parsed result would be literal text as the parser never returns to `NORMAL` first pass due to it being mislead by the escaped quote.

While this solution is not the prettiest and adds more complexity it doesn't touch escape logic. Alternatively (which I think would be better) is to drop escape logic in first pass for strings entirely, because IIRC it doesn't really do much.

### Second pass problem
The second pass and tree build produce a tag value that unquotes the value even though the second quote (or terminating) is escaped.

### DIFF
--- a/text-minimessage/src/main/java/net/kyori/adventure/text/minimessage/internal/parser/TokenParser.java
+++ b/text-minimessage/src/main/java/net/kyori/adventure/text/minimessage/internal/parser/TokenParser.java
@@ -254,8 +254,14 @@ public final class TokenParser {
             case '\'':
             case '"':
               currentStringChar = (char) codePoint;
-              // Look ahead if the quote being opened is ever closed
-              if (message.indexOf(codePoint, i + 1) != -1) {
+              // Look ahead if the quote being opened is ever closed and not escaped
+              int fromIndex = i + 1;
+              int closingQuoteIndex = message.indexOf(codePoint, fromIndex);
+              while (closingQuoteIndex != -1 && isEscaped(message, closingQuoteIndex, fromIndex)) {
+                fromIndex = closingQuoteIndex + 1;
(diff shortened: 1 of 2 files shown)


Strokkur424:
I don't think I particularly like this solution. Imagine you have a very long MiniMessage string with a lot of escaped and not escaped quotes. From what I see here, this would result in a lot of double-iterating, which is very bad for time complexity.

For the problem which you've shown here `<hover:show_text:"Foo\">Bar`, I believe a better solution would be to just continue parsing to the end, and if it ends up in string state never being closed, it should throw a parsing exception, seeing as whoever wrote that String probably forgot a `"` at the end, which we should make them aware of.

What your original issue was trying to show with `<tag:\"foo\">` is that the first string is not escaped, which is the bigger issue here. I think you should modify your solution so that it properly escapes the first `"`, as that should fix it just fine.

Privatech38:
>For the problem which you've shown here <hover:show_text:"Foo\">Bar, I believe a better solution would be to just continue parsing to the end, and if it ends up in string state never being closed, it should throw a parsing exception, seeing as whoever wrote that String probably forgot a " at the end, which we should make them aware of.

I don't believe this is a good solution as `<hover:show_text:"Foo>Bar` is a valid mm string (since it never enters string state).

PR #1322, opened 2025-10-03
Allow creation of a ClickEvent from an Action and Payload

Phoenix616:
This change adds the method `ClickEvent#clickEvent(Action, Payload)` which allows creation of new ClickEvent objects from arbitrary Actions and Payloads. This is useful for custom string -> Component parsers like [Minimessage](https://github.com/KyoriPowered/adventure/blob/fe49db381684f32a539b187438b8fa8d771af1db/text-minimessage/src/main/java/net/kyori/adventure/text/minimessage/tag/standard/ClickTag.java#L67) or [Minedown](https://github.com/Phoenix616/MineDown/blob/e2125596d8475b3d5b29e2c45872890220c2e638/src/main/java/de/themoep/minedown/adventure/MineDownParser.java#L585) which in the past used the now deprecated `ClickEvent#clickEvent(Action, String)` method to easily create new click events from user-supplied action and value strings.

### DIFF
--- a/api/src/main/java/net/kyori/adventure/text/event/ClickEvent.java
+++ b/api/src/main/java/net/kyori/adventure/text/event/ClickEvent.java
@@ -241,7 +241,8 @@ public final class ClickEvent implements Examinable, StyleBuilderApplicable {
    * @return a click event
    * @throws IllegalArgumentException if the action does not support a string payload
    * @since 4.0.0
-   * @deprecated For removal since 4.22.0, not all actions support string payloads
+   * @deprecated For removal since 4.22.0, not all actions support string payloads.
+   *     Use {@link #clickEvent(Action, Payload)} or the event-specific methods.
    */
   @Deprecated
   public static @NotNull ClickEvent clickEvent(final @NotNull Action action, final @NotNull String value) {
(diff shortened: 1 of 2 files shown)


kezz:
What use case do you have for this that isn't solved by switching on the action?

Phoenix616:
Switching on the action means that you have to implement each action type individually causing (imo unnecessary) duplicate code and additionally the need to manually add new action types on updates, while having the dynamic "constructor" exposed would allow "switching" on the payload types so remove a lot of code otherwise required to get the now deprecated functionality working again.

kezz:
Where would the duplicate code be? You can chain switch statements. Also I'm not sure how encouraging error prone code is better than a simple switch.

Phoenix616:
Switching would require code like this which imo. indicates non-clean API design and also fails to automatically support newly added actions with existing payload types:

```java
    public static ClickEvent clickEvent(ClickEvent.Action action, ClickEvent.Payload payload) {
        return switch (action) {
            case CUSTOM -> ClickEvent.custom(((ClickEvent.Payload.Custom) payload).key(), ((ClickEvent.Payload.Custom) payload).nbt());
            case OPEN_URL -> ClickEvent.openUrl(((ClickEvent.Payload.Text) payload).value());
            case OPEN_FILE -> ClickEvent.openFile(((ClickEvent.Payload.Text) payload).value());
            case RUN_COMMAND -> ClickEvent.runCommand(((ClickEvent.Payload.Text) payload).value());
            case SUGGEST_COMMAND -> ClickEvent.suggestCommand(((ClickEvent.Payload.Text) payload).value());
            case COPY_TO_CLIPBOARD -> ClickEvent.copyToClipboard(((ClickEvent.Payload.Text) payload).value());
            case CHANGE_PAGE -> ClickEvent.changePage(((ClickEvent.Payload.Int) payload).integer());
            case SHOW_DIALOG -> ClickEvent.showDialog(((ClickEvent.Payload.Dialog) payload).dialog());
            case null -> null;
            default -> throw new UnsupportedOperationException("Unsupported action " + action);
        };
    }
```

Also I don't see how offering a way to directly use the Action and Payload is encouraging error prone code. You can either check `Action#payloadType` yourself before or have the `ClickEvent` constructor validate this and handle the `IllegalArgumentException` that is thrown when the payload type doesn't match the action.

Phoenix616:
I guess one way to improve this approach could be to use generics similarly to how the HoverEvent does it which could provide at least some additional compile time validation.

EDIT: Hm, I don't think this will work seeing as the `Action` is an enum. That would need to be converted to a class to be able to use generics :S

kezz:
Yeah I see what you mean and actually looking through even the adventure codebase there's actually a fair few places this pattern would repeat. You're right that generics would be best, but unfortunately we're stuck with an enum for now. :sweat_smile:

kezz (approved):
Thanks for the PR!

PR #1324, opened 2025-10-06
chore(5.0): modernize and cleanup a bunch of modules

Strokkur424:
Modernized and removed deprecated stuff from the following modules:

- configurate-4
- nbt
- minimessage
- ansi
- serializer-gson
- serializer-gson-legacy-impl (completely removed)
- serializer-json(-legacy)
- serializer-legacy
- serializer-plain

*(( directly related to #1253 ))*

PR #984, opened 2023-10-10
chore: migrate jetbrains annotations to jspecify

powercasgamer:
closes #982 

- [] api
- [] nbt
- [] key
- [] configurate3
- [] configurate4
- [] text-logger-slf4j
- [] text-minimessage
- [] text-serializer-ansi
- [] text-serializer-gson
- [] text-serializer-gson-legacy-impl
- [] text-serializer-json
- [] text-serializer-json-legacy-impl
- [] text-serializer-legacy
- [] text-serializer-plain

PR #1325, opened 2025-10-07
chore(5.0): nullability and module-info.java cleanup

### LINKED ISSUE #982: Migrate to JSpecify

### LINKED ISSUE #1287: Use module-info instead of manifest
Currently adventure does not declare a `module-info` and relies on the `Automatic-Module-Name`, this is bad for people wanting to add adventure as a transitive dependency in their modules as javac will warn about this behavior and provides other problems. As without module info's you are required to requires every path used, instead of its known as a transitive dependency as an API through the base adventure package.

Currently required: (not exhaustive)
```java
module example {
    requires net.kyori.adventure;
    requires net.kyori.adventure.key;
    requires net.kyori.examination.api;
    requires net.kyori.examination.string;
}
```

Expected:
```java
module example {
    requires net.kyori.adventure; // Or with transitive for platform implementations.
}
```

Strokkur424:
*(( directly related to #1253 ))*

This PR adds a `module-info.java` file to each module and replaces all nullability annotations with jspecify's annotations.

depends on #1324
supersedes #984
closes #982
closes #1287

### DIFF
--- a/api/build.gradle.kts
+++ b/api/build.gradle.kts
@@ -14,6 +14,7 @@ dependencies {
   api(libs.examination.api)
   api(libs.examination.string)
   compileOnlyApi(libs.jetbrainsAnnotations)
+  compileOnlyApi(libs.jspecify)
   testImplementation(libs.guava)
   annotationProcessor(projects.adventureAnnotationProcessors)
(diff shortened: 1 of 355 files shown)


kezz:
Merging manually.

PR #1327, opened 2025-10-10
feat(minimessage): add tag completions

### LINKED ISSUE #654: Add a 'completion' mode to the exposed API
in order to allow for improved usability in commands, it would be helpful to provide an easy way to perform tab-completion of partial messages.

This would work at just the tokenizer level, and given a partial tag, provide completions based on known tags/transformations/placeholders.

Rossterd:
Would close #654


MiniDigger:
while I dont have time to look at this in detail rn, its a ton of code, we should have unit tests for it to verify that it works as expected.
this looks like great work tho!

zml2008:
This is pretty hard to review when you've smashing in commits from another PR, please base it off one of the mainline branches (main/4 or main/5) instead.

Feels like there's quite a bit of duplication of parser logic in here as well, try to avoid that.

Is there a way to distinguish between "format" suggestions (so hints, that shouldn't be added to the cmdline literally such as "#RRGGBB") and value suggestions (such as, `<c:r` suggesting `red`), where you might want to actually tab-complete those into the command line?

Rossterd:
Hey thanks for the response. yeah its probably not worth spending much time looking at this then. I will wait for main/5 to start successfully building and base it off that.

On the differentiation between hints and concrete suggestions: there is not. While bashing this together I mainly had Brigadier commands on my mind and I don't think those have any dynamic way of showing 'format' suggestions. I will add this (and the tests!) to the todo list for when I get to rebase this PR.

PR #1360, opened 2026-01-22
feat(nbt): expose compression methods

notnwse:
There is no reason why these methods should be package-private; making them public costs nothing (it does not break backward compatibility) and allows for the implementation of external compression types (e.g., Zstd).


zml2008:
We intentionally don't support other compression types.

PR #1366, opened 2026-02-09
feat(text-logger-slf4j): add structured log context injector hook

jpenilla:
- add internal ComponentLogContextInjector and ComponentLogRecord types for forwarding raw component log metadata
 - extend ComponentLoggerProvider.LoggerHelper with backward-compatible delegating(..., injector) overload
 - wire Handler.LoggerHelperImpl to construct WrappingComponentLoggerImpl with optional injector
 - wrap all enabled log paths in WrappingComponentLoggerImpl with injector scope begin/close, swallowing injector errors and preserving existing output behavior
 - add tests for raw record capture, throwable forwarding, scope close lifecycle, and injector failure tolerance

This will allow for [Endermux](https://github.com/jpenilla/better-fabric-console/pull/143) to forward rich log messages over the socket.

### DIFF
--- a/text-logger-slf4j/src/main/java/net/kyori/adventure/text/logger/slf4j/ComponentLogContextInjector.java
+++ b/text-logger-slf4j/src/main/java/net/kyori/adventure/text/logger/slf4j/ComponentLogContextInjector.java
@@ -1,0 +1,38 @@
+/*
+ * This file is part of adventure, licensed under the MIT License.
+ *
+ * Copyright (c) 2017-2025 KyoriPowered
+ *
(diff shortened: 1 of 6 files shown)


zml2008:
How worried are we about the unconditional extra allocations from capturing new `Runnable` instances for every log method? I'm used to most logging libs being very concerned about allocations, but i'm not sure how real that impact is in modern JVMs

jpenilla:
I think it's a valid concern, I've inlined the scope management.

jpenilla:
I was able to get Endermux working with this PR:
image

Issue #1349, opened 2025-11-18
Replace BossBarViewer with Audience or use it more.


TurboJax:
Currently there is an interface called "BossBarViewer", and it has two uses:
1. It is used in `net.kyori.adventure.bossbar.BossBar#viewers` which returns an `Iterable<? extends BossBarViewer>`.
2. It defines an `activeBossBars` function that returns an `Iterable<? extends BossBar>`.

Given how this interface is not used in the `BossBar#addViewer` or `BossBar#removeViewer`, I think it'd be a good idea to either remove the BossBarViewer interface or spread out its use.

To remove the interface, the `activeBossBars` function could be moved to the Audience interface, which would not be out of line due to Audience already having the `showBossBar` and `hideBossBar` functions.  Then, the `BossBar#viewers` function would simply return `Iterable<? extends Audience>`.

To expand it's use, I think it would be a good idea to move the `Audience#showBossBar` and `Audience#hideBossBar` functions to BossBarViewer.  The `BossBar#addViewer` and `BossBar#removeViewer` functions would also be modified to take a BossBarViewer object as the parameter.

As for the effects these changes will have on the implementations, adventure-platforms doesn't seem to support bossbars yet anyways, and I'd be willing to help with working on a solution for paper, but I doubt I'll be much help considering I don't know much about the codebase.

I'll work on both these solutions in time on a fork, but I figured I'd make this issue to pose the idea.

TurboJax:
Made the api changes, available here:
[More BossBarViewer](https://github.com/Turbojax07/adventure/tree/turbo/more_bbv)
[No BossBarViewer](https://github.com/Turbojax07/adventure/tree/turbo/no_bbv)

If I should make a pull request, lemme know.

I'll try making the corresponding changes to the paper api, but I'm running into weird build errors with paper-server and will not be able to test the code.

TurboJax:
Closing because I made a pull request for this and it's been inactive for 3 months

kezz:
3 months over a busy winter holiday period is not that long - no need to close issues.

TurboJax:
My bad, I thought it should be closed due to the prs being open.  I think I saw a way to automatically close it with the prs so I'm gonna try to do that

TurboJax:
Never mind, github docs say it only works with the main branch, and I may also need some permission for it.

PR #1370, opened 2026-02-14
feature(api): Remove BossBarViewer

TurboJax:
Simply removes the BossBarViewer class and moves its one function over to the Audience class.  All references to BossBarViewer have been made compatible with the Audience class.

Original Issue: #1349


kashike:
The existing methods need deprecation in 4.x.

TurboJax:
I don't think I can do that on this pr, so I'll make a second one in to add the deprecation warnings.  I'll link it here when it's been made.

TurboJax:
Just made a pr that deprecates functions on the main/4 branch
#1376

build-logic/src/main/kotlin/adventure.common-conventions.gradle.kts
```diff
@@ -84,7 +84,15 @@ indraCrossdoc {
...
     val options = options as? StandardJavadocDocletOptions ?: return@javadoc
+<<<<<<< turbo/no_bbv
+    options.tags(
+      "obsolete:a:Obsolete",
+      "sinceMinecraft:a:Since Minecraft:",
+      "obsoleteSinceMinecraft:a:Obsolete since Minecraft",
+    )
+=======
     options.applyCommonJavadocOptions()
+>>>>>>> main/5
```
Strokkur424: You accidentally committed an unresolved merge conflict.

api/src/main/java/net/kyori/adventure/audience/ForwardingAudience.java
```diff
@@ -162,6 +165,16 @@ default void hideBossBar(final BossBar bar) {
     for (final Audience audience : this.audiences()) audience.hideBossBar(bar);
   }
 
+  @Override
+  default @UnmodifiableView Iterable<? extends BossBar> activeBossBars() {
+    final Set<BossBar> bossBars = new HashSet<>();
+    for (final Audience audience : this.audiences()) {
+      for (final BossBar bb : audience.activeBossBars()) bossBars.add(bb);
```
Strokkur424: Pure style, but I wouldn't inline the for-loop body.
```suggestion
      for (final BossBar bb : audience.activeBossBars()) {
        bossBars.add(bb);
      }
```

api/src/main/java/net/kyori/adventure/audience/Audience.java
```diff
@@ -441,6 +440,17 @@ default void showBossBar(final BossBar bar) {
   default void hideBossBar(final BossBar bar) {
   }
 
+  /**
+   * Gets an unmodifiable view of all known currently active bossbars.
+   *
+   * @return an unmodifiable view of all known currently active bossbars
+   * @since 4.14.0
+   */
+  @UnmodifiableView
```
Strokkur424: I would generally just make it `@Unmodifiable` instead of a view, since you do not really want people to modify the backing `Iterable` and except it to change the currently active boss bars.
```suggestion
  @Unmodifiable
```

api/src/main/java/net/kyori/adventure/audience/ForwardingAudience.java
```diff
@@ -162,6 +165,16 @@ default void hideBossBar(final BossBar bar) {
     for (final Audience audience : this.audiences()) audience.hideBossBar(bar);
   }
 
+  @Override
+  default @UnmodifiableView Iterable<? extends BossBar> activeBossBars() {
```
Strokkur424: This is not a view. The method returns a **new** HashSet with the values copied over.

```suggestion
  default @Unmodifiable Iterable<? extends BossBar> activeBossBars() {
```

Strokkur424 (changes requested):

build-logic/src/main/kotlin/adventure.common-conventions.gradle.kts
TurboJax: My bad, this bit looks out of scope for the PR so I'm just going to delete the change

TurboJax:
Awesome, thanks for the reviews!  Sorry that I missed that merge conflict, I'll double check the implementations again to make sure I didn't miss anything else.  About #1376, I'll rebase that soon.

api/src/main/java/net/kyori/adventure/audience/Audience.java
TurboJax: Sorry, rechecking this, and figured it should be reopened.

Wouldn't `@UnmodifiableView` be the correct annotation for the Audience#activeBossBars function?  `@Unmodifiable` is for if the contents of the collection never change, but the active boss bars of a player can change, which is why `@UnmodifiableView` was used.

Sorry about the confusion, I haven't used the jetbrains annotations for very long.

Strokkur424: Sure, but the contents of the return value do not change.

The "view" is dependent on whether the object returned by the method will be able to change. Think of it like this:

1. I call `Iterable<? extends BossBar> bossbars = Audience.activeBossBars()`
2. I add a boss bar to the audience.
3. I now iterate through `bossbars`. Does it contain the newly added boss bar?

If `3` is true, then it is a view. If it does not contain the newly added boss bar, then it's not a view, but a copy.


Generally, the question on if you want to use a view or not depends on what you expect users to do. In immutable design, which the adventure library mostly follows, I personally generally expect the return values of methods to not be mutable/change.

In the `ForwardingAudience` implementation of the `activeBossBars` method, you create a new set and add all other audience's boss bars to that set. It is inherently impossible to provide a view, as `addAll` copies the collection entries, it doesn't keep a reference to the original collection. So that one is definitely not a view.

I think it would be fair to generally expected implementations to return a copy of the data, and not a view, aka. the internal instance.

TurboJax: Ah that makes sense.  Thanks for explaining it!

Strokkur424: No problem!

TurboJax:
Hey thinking about making the BossBar#viewers and Audience#activeBossBars functions return `Set<[class]>` instead of `Iterable<? extends [class]>`.  Is anyone opposed to this?

PR #1372, opened 2026-02-18
feat(api): Add pages-only constructor to BookImpl

### LINKED ISSUE #1348: Book author and title should be optional
They don't have to be nullable, but the builder should default to the empty component and we should have a pages-only constructor.

Privatech38:
Closes #1348

### DIFF
--- a/api/src/main/java/net/kyori/adventure/inventory/Book.java
+++ b/api/src/main/java/net/kyori/adventure/inventory/Book.java
@@ -23,7 +23,6 @@
  */
 package net.kyori.adventure.inventory;
 
-import java.util.ArrayList;
 import java.util.Collection;
 import java.util.List;
 import net.kyori.adventure.audience.Audience;
@@ -32,6 +31,8 @@ import net.kyori.adventure.text.Component;
 import net.kyori.adventure.util.Buildable;
 import org.jetbrains.annotations.Contract;
 import org.jetbrains.annotations.Unmodifiable;
+
+import static java.util.Objects.requireNonNull;
 
 /**
  * Represents the in-game interface of a book.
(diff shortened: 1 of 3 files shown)


kezz:
This is useless without an associated change to the book interface

Privatech38:
@kezz implemeted the creation methods

api/src/main/java/net/kyori/adventure/inventory/Book.java
```diff
@@ -70,6 +70,28 @@ static Book book(final Component title, final Component author, final Component.
...
 
+  /**
+   * Creates a book with title and author set to {@link Component#empty()}.
+   *
+   * @param pages the collection of pages
+   * @return a book
+   * @since 5.1.0
+   */
+  static Book book(final Collection<Component> pages) {
+    return new BookImpl(new ArrayList<>(pages));
```
kashike, line 81:
```suggestion
    return new BookImpl(List.copyOf(pages));
```

kashike (changes requested):

api/src/main/java/net/kyori/adventure/inventory/Book.java
Privatech38: But that would differ from the implementation that uses title and author?
https://github.com/PaperMC/adventure/blob/906e4f021c00a7d5020d5d4055b670a96da71313/api/src/main/java/net/kyori/adventure/inventory/Book.java#L56-L58

Privatech38: @kashike

kezz:
Thanks for the PR! Just made a few changes to simplify a bit of the logic here.

kezz (approved):

PR #1373, opened 2026-02-18
feat(api): Add toString implementation for StyleImpl

### LINKED ISSUE #1371: `StyleImpl` does not override `Object#toString` method
The `net.kyori.adventure.text.format.StyleImpl` class does not override `Object#toString` method resulting in the default `Object#toString` returning the class name and hash value.

Privatech38:
Closes #1371


kezz (approved):
Thanks for the PR!

PR #1374, opened 2026-02-19
chore: add back dropped toString implementations

Emilxyz:
This PR adds back some toString implementations that were dropped with the removal of the examination api on classes that were not converted to records.

Unless I missed something, everything that was previously implementing `Examinable` should now be either a record or have its own toString (except `StyleImpl`, which is covered by #1373).

(The links take you to the old examinable implementation on main/4)
- [`SoundImpl`](https://github.com/PaperMC/adventure/blob/main/4/api/src/main/java/net/kyori/adventure/sound/SoundImpl.java#L109)
- [`DecorationMap`](https://github.com/PaperMC/adventure/blob/main/4/api/src/main/java/net/kyori/adventure/text/format/DecorationMap.java#L137)
- [`AbstractTranslationStore`](https://github.com/PaperMC/adventure/blob/main/4/api/src/main/java/net/kyori/adventure/translation/AbstractTranslationStore.java#L160) (missing `name` and `defaultLocale`)
- [`AbstractTranslationStore.Translation`](https://github.com/PaperMC/adventure/blob/5e2284abb3259be2334911dc024bd4aaede65e6e/api/src/main/java/net/kyori/adventure/translation/AbstractTranslationStore.java#L213)
- [`GlobalTranslatorImpl`](https://github.com/PaperMC/adventure/blob/5e2284abb3259be2334911dc024bd4aaede65e6e/api/src/main/java/net/kyori/adventure/translation/GlobalTranslatorImpl.java#L116)
- [`LegacyFormat`](https://github.com/PaperMC/adventure/blob/main/4/text-serializer-legacy/src/main/java/net/kyori/adventure/text/serializer/legacy/LegacyFormat.java#L120)

### DIFF
--- a/api/src/main/java/net/kyori/adventure/sound/SoundImpl.java
+++ b/api/src/main/java/net/kyori/adventure/sound/SoundImpl.java
@@ -100,6 +100,17 @@ sealed abstract class SoundImpl implements Sound permits SoundImpl.Eager, SoundI
     return result;
   }
 
+  @Override
+  public String toString() {
+    return "SoundImpl{" +
+      "name=" + this.name() +
+      ", source=" + this.source +
+      ", volume=" + this.volume +
+      ", pitch=" + this.pitch +
+      ", seed=" + this.seed +
+      '}';
+  }
+
   static final class BuilderImpl implements Builder {
     private static final float DEFAULT_VOLUME = 1f;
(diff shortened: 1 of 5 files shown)


kezz (approved):
Thanks for the PR!

PR #1375, opened 2026-02-19
fix(text-minimessage): support all uuid versions in head tag

Emilxyz:
Change the `<head>` tag to simply try and parse a UUID instead of using a (v4 only) regex.

Fixes an issue raised by a user on [discord](https://discord.com/channels/289587909051416579/1342377752774443008/1473744214117318717)

### DIFF
--- a/text-minimessage/src/main/java/net/kyori/adventure/text/minimessage/tag/standard/SequentialHeadTag.java
+++ b/text-minimessage/src/main/java/net/kyori/adventure/text/minimessage/tag/standard/SequentialHeadTag.java
@@ -26,7 +26,6 @@ package net.kyori.adventure.text.minimessage.tag.standard;
 import java.util.Objects;
 import java.util.UUID;
 import java.util.function.Function;
-import java.util.regex.Pattern;
 import net.kyori.adventure.key.Key;
 import net.kyori.adventure.text.Component;
 import net.kyori.adventure.text.ObjectComponent;
@@ -52,8 +51,6 @@ import org.jspecify.annotations.Nullable;
  */
 final class SequentialHeadTag {
   static final String HEAD = "head";
-  // TODO: Move this to HeadTag
-  private static final Pattern UUIDv4_PATTERN = Pattern.compile("[0-9a-f]{8}-[0-9a-f]{4}-4[0-9a-f]{3}-[89ABCD][0-9a-f]{3}-[0-9a-f]{12}", Pattern.CASE_INSENSITIVE);
 
   static final TagResolver RESOLVER = SerializableResolver.claimingComponent(
     HEAD,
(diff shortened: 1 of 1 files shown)


text-minimessage/src/main/java/net/kyori/adventure/text/minimessage/tag/standard/SequentialHeadTag.java
```diff
@@ -92,13 +89,15 @@ static Tag create(final ArgumentQueue args, final Context ctx) throws ParsingExc
...
 
-    if (SequentialHeadTag.UUIDv4_PATTERN.matcher(argument).matches()) {
+    try {
       return Tag.selfClosingInserting(Component.object(
         ObjectContents.playerHead()
           .id(UUID.fromString(argument))
           .hat(outerLayer.toBooleanOrElse(PlayerHeadObjectContents.DEFAULT_HAT))
           .build()
       ));
+    } catch (final IllegalArgumentException ignored) {
```
kezz: Just to be safe, it'd be better if this was as close to the UUID.fromString call as possible

kezz (changes requested):

text-minimessage/src/main/java/net/kyori/adventure/text/minimessage/tag/standard/SequentialHeadTag.java
```diff
@@ -92,10 +89,16 @@ static Tag create(final ArgumentQueue args, final Context ctx) throws ParsingExc
       throw ctx.newException("Too many arguments present", args);
     }
 
-    if (SequentialHeadTag.UUIDv4_PATTERN.matcher(argument).matches()) {
+    UUID uuid = null;
+    try {
+      uuid = UUID.fromString(argument);
+    } catch (final IllegalArgumentException ignored) {
+      // If UUID parsing fails, try other options
+    }
```
Strokkur424, line 97: This is a minor thing, but you might as well remove the if-statement and move the return block into the try block directly. That also saves you from having a mutable uuid variable floating around.

```suggestion
    try {
      final UUID uuid = UUID.fromString(argument);
      return Tag.selfClosingInserting(Component.object(
        ObjectContents.playerHead()
          .id(uuid)
          .hat(outerLayer.toBooleanOrElse(PlayerHeadObjectContents.DEFAULT_HAT))
          .build()
      ));
    } catch (final IllegalArgumentException ignored) {
      // If UUID parsing fails, try other options
    }
```

emilyy-dev: This is no different than the code that was there before, Kezz pointing out the try-catch should surround the fromString call as closely as possible

Strokkur424: Ah, the try/catch was narrowed deliberately? Then the current code is fine.

kezz (approved):
LGTM, thanks!

PR #1388, opened 2026-03-24
feature(minimessage): Presets

kezz:
As evidenced by recent publicity around an """""exploit""""" in Paper servers, MiniMessage could do with an improvement to allow developers to access really simple "safe" presets. This PR introduces such a system via addition of a Preset enum that contains a standardised configuration for builders.

### DIFF
--- a/text-minimessage/src/main/java/net/kyori/adventure/text/minimessage/MiniMessage.java
+++ b/text-minimessage/src/main/java/net/kyori/adventure/text/minimessage/MiniMessage.java
@@ -23,14 +23,19 @@
  */
 package net.kyori.adventure.text.minimessage;
 
+import java.util.Objects;
 import java.util.function.Consumer;
 import java.util.function.UnaryOperator;
 import net.kyori.adventure.builder.AbstractBuilder;
 import net.kyori.adventure.pointer.Pointered;
 import net.kyori.adventure.text.Component;
+import net.kyori.adventure.text.TextComponent;
+import net.kyori.adventure.text.VirtualComponent;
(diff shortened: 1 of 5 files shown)


text-minimessage/src/main/java/net/kyori/adventure/text/minimessage/MiniMessage.java
```diff
@@ -54,6 +59,134 @@ static MiniMessage miniMessage() {
...
+    };
+
+    /**
+     * The name map.
+     *
+     * <p>This can be used to easily make configurable presets.</p>
+     *
+     * @since 5.0.0
+     */
+    public static final Index<String, Preset> NAMES = Index.create(Preset.class, Preset::name);
```
Strokkur424, line 187: Personally, I'd maybe move this to the top of the file just so it's easier accessible (all fields at the start).

Strokkur424 (approved):

text-minimessage/src/main/java/net/kyori/adventure/text/minimessage/MiniMessage.java
kennytv: nah the enum fields always go first

kennytv (approved):
I like it!

text-minimessage/src/main/java/net/kyori/adventure/text/minimessage/MiniMessage.java
Strokkur424: /shrug

PR #1390, opened 2026-03-25
26.1 update

kezz:
Includes the following changes:
- Addition of `plain` property to NBT components, mutually exclusive with `interpret`
- Addition of `fallback` property to object components

### DIFF
--- a/api/src/main/java/net/kyori/adventure/text/Component.java
+++ b/api/src/main/java/net/kyori/adventure/text/Component.java
@@ -570,7 +570,7 @@ public sealed interface Component extends ComponentBuilderApplicable, ComponentL
    */
   @Contract(value = "_ -> new", pure = true)
   static ObjectComponent object(final ObjectContents objectContents) {
-    return ObjectComponentImpl.create(List.of(), Style.empty(), objectContents);
+    return ObjectComponentImpl.create(List.of(), Style.empty(), objectContents, null);
   }
 
   /*
--- a/api/src/main/java/net/kyori/adventure/text/Component.java
+++ b/api/src/main/java/net/kyori/adventure/text/Component.java
@@ -326,7 +326,7 @@ public sealed interface Component extends ComponentBuilderApplicable, ComponentL
    */
   @Contract(value = "_, _, _, _ -> new", pure = true)
   static BlockNBTComponent blockNBT(final String nbtPath, final boolean interpret, final @Nullable ComponentLike separator, final BlockNBTComponent.Pos pos) {
-    return BlockNBTComponentImpl.create(List.of(), Style.empty(), nbtPath, interpret, separator, pos);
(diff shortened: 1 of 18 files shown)


api/src/main/java/net/kyori/adventure/text/ObjectComponent.java
```diff
@@ -47,7 +48,32 @@ public sealed interface ObjectComponent extends ScopedComponent<ObjectComponent>
...
+   * components are not supported, such as server MOTDs.</p>
+   *
+   * @return the fallback component, if any
+   * @since 5.0.0
+   * @sinceMinecraft 26.1.0
+   */
+  @Nullable Component fallback();
+
+  /**
+   * Creates a copy of this object componet with the given fallback.
```
aerulion: Minor typo (componet)

kezz: Nice spot. Thanks!

api/src/main/java/net/kyori/adventure/text/Component.java
```diff
@@ -570,7 +570,7 @@ static ObjectComponent object(final Consumer<? super ObjectComponent.Builder> co
    */
   @Contract(value = "_ -> new", pure = true)
   static ObjectComponent object(final ObjectContents objectContents) {
-    return ObjectComponentImpl.create(List.of(), Style.empty(), objectContents);
+    return ObjectComponentImpl.create(List.of(), Style.empty(), objectContents, null);
   }
```
Emilxyz, line 574: Maybe we can also add a `object(ObjectContents, ComponentLike)` overload for creating a ObjectComponent with a fallback?

api/src/main/java/net/kyori/adventure/text/ObjectComponent.java
```diff
@@ -47,7 +48,32 @@ public sealed interface ObjectComponent extends ScopedComponent<ObjectComponent>
...
+
+  /**
+   * Gets the fallback component.
+   *
+   * <p>The client displays this instead of this component in areas where object
+   * components are not supported, such as server MOTDs.</p>
+   *
+   * @return the fallback component, if any
+   * @since 5.0.0
+   * @sinceMinecraft 26.1.0
```
Emilxyz:
```suggestion
   * @sinceMinecraft 26.1
```
Small nitpick, would also apply to the other tags on the 2 methods below

Emilxyz (commented):
We might also want to make use of the `fallback` property of object components in ComponentFlattener

api/src/main/java/net/kyori/adventure/text/Component.java
kezz: I think it's fine as is, I don't really think anyone is ever going to use the fallback property anyway and there are other cases where we don't make static constructors for uncommonly used properties.

kezz:
> We might also want to make use of the fallback property of object components in ComponentFlattener

I'm also fine to leave the current behaviour as-is. If we get any real-world requests for this we can reconsider then, but for now I think it's fine to ignore the fallback in this example.

PR #1396, opened 2026-04-15
fix(all): Fix SPI when using modules.

### LINKED ISSUE #1384: Use JPMS Service API
Adventure 5.0 has support for JPMS and should be able to use the `uses` keyword in its `module-info` for it platform services API. See https://dev.java/learn/modules/services/. This does not replace the service files in most implementations as they are required if the user application runs in the unnamed module. (Using JPMS erasure like shadowJar or fatjars)

Also it seems you might be missing the `java.modularity.inferModulePath = true` option in Gradle for JPMS which could explain why module annotations didnt work.

Would love to see this before 5.0 is released.

kermandev:
Previously, these will fail because utils doesn't declare uses on every provider. The simple fix would just to add this to the uses, but this would just be a hack. Instead, we should pass through the service loader inside the module so it has the correct module information so we can declare uses in the correct modules that uses them.

All previous services methods have been deprecated in favor of this. ~~Also, the behavior is slightly changes as Services0 uses ServiceLoader(clazz, clazz.getClassLoader()) Which should no longer be needed as it uses the same class loader as the thread that initializes the class.~~

Should also resolve https://github.com/PaperMC/adventure/issues/1384, didn't realize it would be an issue before then.


Strokkur424 (approved):

kezz (approved):
Thanks for the PR!

PR #1403, opened 2026-04-16
refactor(minimessage): revert change to Hover tag SNBT resolution

### LINKED ISSUE #1397: Revert change to Hover Tag SNBT item display resolution
5.0.0 changed the hover tag resolution by replacing item display SNBT deserialization with an exception say its not supported. 

This change was introduced with commit 77945ebbe09c7701020d368571fd792e7ed99364. 

Reverting the change will resolve this issue.

Privatech38:
This PR reverts to the previous behavior where it allowed SNBT to be passed as an argument for item display. The commit changed that behaviour by throwing an exception instead, marking it as deprecated and removed.

Closes #1397


kezz (approved):
Thanks for the PR!

PR #1405, opened 2026-04-21
Add intersex inclusive pride flag to PrideTag

me4502:
Noticed this flag was missing from the list, and it's one of the more common forms of general pride flag. 

On a personal level, I also find this flag fairly beneficial for intersex awareness as it creates conversations around what it means. Whether that can happen in the very "low information density" form of MC text colours is another question :P

Note that this is colours of "Intersex flag" + "Progress pride flag" as I couldn't find consistent descriptions of the colours in this flag online.


purpurcof:
For what?

kezz (approved):
Thanks for the PR!

PR #1421, opened 2026-05-13
feature(api): Easier way to create object components

kezz:


api/src/main/java/net/kyori/adventure/text/object/ObjectContentsLike.java
```diff
@@ -0,0 +1,42 @@
...
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+ * SOFTWARE.
+ */
+package net.kyori.adventure.text.object;
+
+import org.jetbrains.annotations.Contract;
+
+/**
+ * Something that can be represented as an {@link ObjectContents}.
```
Privatech38, line 29: Maybe drop the an since the noun is in plural
```suggestion
 * Something that can be represented as {@link ObjectContents}.
```

api/src/main/java/net/kyori/adventure/text/Component.java
```diff
@@ -561,6 +563,22 @@ static ObjectComponent object(final Consumer<? super ObjectComponent.Builder> co
...
+   * @return an object component
+   * @since 5.2.0
+   */
+  @Contract(value = "_ -> new", pure = true)
+  static ObjectComponent object(final ObjectContentsLike objectContentsLike) {
+    if (requireNonNull(objectContentsLike, "objectContentsLike") instanceof PlayerHeadObjectContents.SkinSource skinSource) {
+      return object(ObjectContents.playerHead(skinSource));
+    } else {
+      return object(objectContentsLike.asObjectContents());
+    }
```
Privatech38, line 579: Why are you checking if the contents are an instance of SkinSource and not adding any extra logic, but rather duplicating code? Couldn't you just call `skinSource.asObjectContents()` or perhaps even shorter:
```suggestion
    requireNonNull(objectContentsLike, "objectContentsLike")
    return object(objectContentsLike.asObjectContents());
```

Privatech38 (changes requested):

api/src/main/java/net/kyori/adventure/text/Component.java
kezz: It skips a method call and is similar to how it's done in other areas of the codebase.

PR #1423, opened 2026-05-24
chore: change gpg key

MiniDigger:


hboyd2003:
Since you are swapping keys from Kyori to paper the prefix should also be changed
https://github.com/PaperMC/adventure/blob/2e719c4325fd6f79f5467bbdc122603c744b4369/build-logic/src/main/kotlin/adventure.base-conventions.gradle.kts#L23

kezz (approved):

Issue #1299, opened 2025-09-10
VirtualComponents don't render when inside translatable components


Pablete1234:
I'm unsure as to why this was done, but can be seen here:
https://github.com/KyoriPowered/adventure/blob/main/4/api/src/main/java/net/kyori/adventure/text/renderer/TranslatableComponentRenderer.java#L202

`if (arg.value() instanceof Component && !(arg.value() instanceof VirtualComponent)) {`

The second part of that `if`, means that virtual components get completely ignored when rendering translatable, making them unusable other than for their fallback.

These are some examples of what you'd expect code to result to:
Works and has correct behavior:
```java
// Renders as "Hello Username" just fine
virtual(Player.class, p -> text("Hello " + p.getName())); 
```

Does not work, has incorrect behavior:
```java
// Assume some.key=Greeting: {}

// Renders as "Greeting: " (no fallback defined), should be "Greeting: Hello Username"
translatable(some.key, virtual(Player.class, p -> text("Hello " + p.getName())));

// Renders as "Greeting: fallback" (a fallback is defined), should be "Greeting: Hello Username"
var rendererWithFallback = new VirtualComponentRenderer<Player> {
  @Override
  public ComponentLike apply(Player p) {
    return text("Hello " + p.getName());
  }
  @Override
  public String fallbackString() {
    return "fallback";
  }
}

translatable(some.key, virtual(Player.class, renderWithFallback);
```

TonytheMacaroni:
The `if` is there because there is currently no logic for rendering virtual components. Attempting to render the virtual component there would [collapse it to the fallback string](https://github.com/KyoriPowered/adventure/blob/main/4/api/src/main/java/net/kyori/adventure/text/renderer/AbstractComponentRenderer.java#L48-L52), but would cause issues with MiniMessage translators, since virtual components are used for named arguments and tag resolvers.

Pablete1234:
Rendering a virtual component runs it thru the renderer, if is responsability of the plugin to define a renderer that can handle them. It is inconsistent that just because you wrap it in a translatable it no longer has any way of functioning

TonytheMacaroni:
The virtual components I mentioned for [MiniMessage translators](https://docs.advntr.dev/minimessage/translator.html#minimessage-translator) are not provided by plugins. They are how you specify named arguments, tag resolvers and a target when using MiniMessage translations:

[...]

PR #1437, opened 2026-07-22
feat: introduce MiniMessageContext to distinguish MiniMessage when rendering

kashike:
We currently prevent rendering `VirtualComponent`s through `TranslatableComponentRenderer` when they are used as arguments to a `TranslatableComponent`. Supporting them requires extending the renderer and overriding `renderTranslatable` to restore that behavior, which is not ideal.

### DIFF
--- a/api/src/main/java/net/kyori/adventure/text/VirtualComponentRenderer.java
+++ b/api/src/main/java/net/kyori/adventure/text/VirtualComponentRenderer.java
@@ -52,4 +52,15 @@ public interface VirtualComponentRenderer<C> {
   default String fallbackString() {
     return "";
   }
+
+  /**
+   * Checks if a {@link VirtualComponent} should be skipped when being rendered through a {@link net.kyori.adventure.text.renderer.TranslatableComponentRenderer} when used as an argument.
+   *
+   * @param component the component
+   * @return {@code true} to skip rendering, or {@code false} to render
+   * @since 5.2.1
+   */
+  default boolean skipRenderingWhenUsedAsTranslationArgument(final VirtualComponent component) {
+    return false;
+  }
 }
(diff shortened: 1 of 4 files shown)


Pablete1234:
Resolves #1299

masmc05:
Maybe instead a VirtualComponentRenderer#translatable() returning a boolean would be better? Then other api usages could indicate if the component is translatable, with a default method always returning true, while minimessage always returning false

Pablete1234:
Thing is, more than if the virtual component is translatable or not, the thing in question is if it can be rendered at all, or if its just a piece of context that shall remain untouched when inside translatable components, so that a translator (minimessage's translator in this case) can take care of it.

In these cases, it does make sense that the context type is a dummy class because its explicitly being marked as cannot-and-shouldnot-render-this, the context type isnt needed.

I do think a more generic name for this context type could be used but i can't come up with anything better.

masmc05:
I still consider that forcing a specific context type for virtual components that are such data holders is wrong. Some may want this renderer to ignore the component, while a different non-adventure renderer to still render it. And also in the future if a new adventure renderer would need to handle them differently again this approach would force either both ignored or none.
A different solution to this if the method isn't good would be a marker interface. If the renderer implements it then it would be ignored (and all minimessage ones would implement it)

kashike:
The naming of such a marker interface is difficult - do you have any suggestions? Keep in mind that this is, currently, only applicable when the `VirtualComponent` is a `TranslatableComponent` argument.

masmc05:
`TranslationOpaque` sounds like a good fit

kashike:
What about just adding a method to `VirtualComponentRenderer`? Something like:
```java
  default boolean skipRenderingWhenUsedAsTranslationArgument(final VirtualComponent component) {
    return false;
  }
```

PR #1438, opened 2026-07-23
perf(minimessage): add system property to skip tag validation

ProdPreva1l:
This PR introduces a system property intended for use by advanced users who are confident that they are using valid minimessage tags.

This has a very small usecase but a pretty major impact.

https://spark.lucko.me/Aqo35xCxyT?hl=8160

As seen in this spark at the linked method call, its allocation rate is about 650kbps which is quite high for what the call does and adds up.


SirYwell:
That looks like a problem with GraalVM. Can you provide a spark profile using HotSpot/C2?

If there is a problem with excessive allocations, I think it would be better to implement the validation manually instead of relying on `Pattern`. Skipping the validation for performance is a somewhat flawed idea imo.

ProdPreva1l:
Here's with just C2 via Zulu.
https://spark.lucko.me/iAnMKIsXbP

A very slight improvement.

kezz:
> If there is a problem with excessive allocations, I think it would be better to implement the validation manually instead of relying on Pattern. Skipping the validation for performance is a somewhat flawed idea imo.

Agreed. I'm going to close this because of that.

PR #1441, opened 2026-07-31
feature(api): Add locale-specific translation unregistering

### LINKED ISSUE #1361: Ability to override or remove translations from a translation store
It is currently impossible to alter single translationss after they have been registered without removing the entire key with all translations
My use case for this is a command to register translations during runtime
example:
`/translation add <key> <locale> <translation>`
`/translation remove <key> <locale>`

```
/translation add hello-world en-US Hello World!
/translation add hello-world en-US Hallo Weltt!
/translation remove hello-world de-DE # because of typo
/translation add hello-world de-DE Hallo Welt! # fix spelling
```

`TranslationStore#unregister(String)` unregisteres all translations tied to `hello-world` in this instance even though I only want to remove/override the german translation

ZapolyarnyDev:
Adds `TranslationStore#unregister(String, Locale)` to allow unregistering a translation for a specific locale without affecting the other translations for that key

Includes tests for preserving other locales and removing empty keys

Closes #1361

### DIFF
--- a/api/src/main/java/net/kyori/adventure/translation/AbstractTranslationStore.java
+++ b/api/src/main/java/net/kyori/adventure/translation/AbstractTranslationStore.java
@@ -141,6 +141,14 @@ public abstract class AbstractTranslationStore<T> implements TranslationStore<T>
   }
 
   @Override
+  public final void unregister(final String key, final Locale locale) {
+    this.translations.computeIfPresent(requireNonNull(key, "key"), (ignored, translation) -> {
+      translation.unregister(requireNonNull(locale, "locale"));
+      return translation.isEmpty() ? null : translation;
+    });
+  }
+
+  @Override
   public final Key name() {
     return this.name;
   }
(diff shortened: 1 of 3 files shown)


api/src/main/java/net/kyori/adventure/translation/AbstractTranslationStore.java
```diff
@@ -140,6 +140,14 @@ public final void unregister(final String key) {
     this.translations.remove(key);
   }
 
+  @Override
+  public final void unregister(final String key, final Locale locale) {
+    this.translations.computeIfPresent(requireNonNull(key, "key"), (ignored, translation) -> {
```
kashike, line 145: Why `computeIfPresent`? I don't know why we need to create one if absent when we are unregistering.

ZapolyarnyDev: I used `computeIfPresent` specifically to avoid creating a translation entry when unregistering an unknown key. If the key exists, we remove the locale; if no locales remain, returning `null` removes the key. Otherwise, it’s a no-op.

